PR #340, opened 2025-12-09
Unify bedrock & openai client creation for integ tests

zastrowm:
In order to better share code between browser and node integ tests, use a single implementation for creating clients.  We accomplish this by taking advantage of vitest's [global-setup](https://vitest.dev/config/globalsetup) which can pass/provide information to the tests regardless of what environment they're running in - we thus provide the credentials in global setup to the tests, which can then access the credentials in both environments.

Per the last PR (#337), I also unified the skipping & logic of creating clients making it much more standard how we'll write tests going forward.

## Notes

 - I also enable all logs when running tests in CI/CD
 - I moved test setup into the `__fixtures__` folder for better code organization
 - Added test time out to 60s

## Tests

 - 77 passed tests: https://github.com/strands-agents/sdk-typescript/actions/runs/20489351356/job/58878293836
 - Matching the normal 77 tests: https://github.com/strands-agents/sdk-typescript/actions/runs/20475753586/job/58839920237

## Follow-up

A follow-up PR is planned to run more tests in both the browser & node; this gets us one step closer

### DIFF
--- a/test/integ/__fixtures__/_setup-global.ts
+++ b/test/integ/__fixtures__/_setup-global.ts
@@ -1,5 +1,7 @@
 /**
- * Global setup that runs once before all integration tests and possibly runs in the *parent* process
+ * Global setup that runs once before all integration tests and possibly runs in the *parent* process.
+ *
+ * _setup-test on the other hand runs in the *child* process.
  */
 
 import { SecretsManagerClient, GetSecretValueCommand } from '@aws-sdk/client-secrets-manager'
@@ -55,7 +57,7 @@ async function loadApiKeysFromSecretsManager(): Promise<void> {
 }
 
 /**
- * Perform shared setup for the integration tests unless it's already been.
+ * Perform shared setup for the integration tests.
  */
 export async function setup(project: TestProject): Promise<void> {
   console.log('Global setup: Loading API keys from Secrets Manager...')
--- a/test/integ/__fixtures__/_setup-test.ts
+++ b/test/integ/__fixtures__/_setup-test.ts
@@ -1,5 +1,7 @@
 /**
- * Global setup that runs once before all integration tests and possibly runs in the *parent* process
+ * Test setup that runs once before all integration tests, but in the *child* process.
+ *
+ * _setup-global on the other hand runs in the *parent* process.
  */
 
 import { beforeAll } from 'vitest'
--- a/test/integ/__fixtures__/model-providers.ts
+++ b/test/integ/__fixtures__/model-providers.ts
@@ -2,7 +2,6 @@
  * Contains helpers for creating various model providers that work both in node & the browser
  */
 
-import { isCI } from './test-helpers.js'
 import { inject } from 'vitest'
 import { BedrockModel, type BedrockModelOptions } from '$/sdk/models/bedrock.js'
 import { OpenAIModel, type OpenAIModelOptions } from '$/sdk/models/openai.js'
@@ -10,7 +9,7 @@ import { OpenAIModel, type OpenAIModelOptions } from '$/sdk/models/openai.js'
 export const bedrock = {
   name: 'BedrockModel',
   get skip() {
-    return !isCI() && inject('provider-bedrock').shouldSkip
+    return inject('provider-bedrock').shouldSkip
   },
   createModel: (options: BedrockModelOptions = {}): BedrockModel => {
     const credentials = inject('provider-bedrock').credentials
@@ -31,7 +30,7 @@ export const bedrock = {
 export const openai = {
   name: 'OpenAIModel',
   get skip() {
-    return !isCI() && inject('provider-openai').shouldSkip
+    return inject('provider-openai').shouldSkip
   },
   createModel: (config: OpenAIModelOptions = {}): OpenAIModel => {
     const apiKey = inject('provider-openai').apiKey
--- a/test/integ/bash.test.ts
+++ b/test/integ/bash.test.ts
@@ -4,7 +4,7 @@ import { bash } from '$/sdk/vended-tools/bash/index.js'
 import { getMessageText } from './__fixtures__/model-test-helpers.js'
 import { bedrock } from './__fixtures__/model-providers.js'
 
-describe.skipIf(bedrock.skip || process.platform === 'win32')('Bash Tool Integration', { timeout: 60000 }, () => {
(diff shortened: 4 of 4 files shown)


dbschmigelski (dismissed):

test/integ/__fixtures__/_setup-global.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  for (const provider of requiredProviders) {
+    if (!process.env[provider]) {
+      throw new Error(`Missing required environment variables for ${provider}`)
+    }
+  }
+}
+
+/**
+ * Perform shared setup for the integration tests unless it's already been.
```
Unshure: nit: "... already been run." ?

test/integ/__fixtures__/_setup-global.ts
```diff
@@ -0,0 +1,111 @@
+/**
+ * Global setup that runs once before all integration tests and possibly runs in the *parent* process
```
Unshure: This is supposed to run as part of the vitest setup right? This file runs before of the tests node/browser environment, and in the vitest setup env?

test/integ/agent.test.ts
```diff
@@ -33,21 +31,7 @@ const calculatorTool = tool({
...
-  },
-  {
-    name: 'OpenAIModel',
-    skip: shouldSkipOpenAITests(),
-    createModel: () => new OpenAIModel(),
-  },
-]
-
-describe.each(providers)('Agent with $name', ({ name, skip, createModel }) => {
+describe.each(allProviders)('Agent with $name', ({ name, skip, createModel }) => {
```
Unshure: nit
```suggestion
describe.each(allProviders)('Agent with $modelProviderName', ({ modelProviderName, skip, createModel }) => {
```

JackYPCOnline (dismissed):

test/integ/__fixtures__/model-providers.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+import { isCI } from './test-helpers.js'
+import { inject } from 'vitest'
+import { BedrockModel, type BedrockModelOptions } from '$/sdk/models/bedrock.js'
+import { OpenAIModel, type OpenAIModelOptions } from '$/sdk/models/openai.js'
+
+export const bedrock = {
+  name: 'BedrockModel',
+  get skip() {
```
Unshure: I never knew about object getters and setters! https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_objects#defining_getters_and_setters

There are so many little cool tricks like this in ts/js ❤️

test/integ/__fixtures__/model-providers.ts
```diff
@@ -0,0 +1,53 @@
...
+        credentials: credentials,
+      },
+    })
+  },
+}
+
+export const openai = {
+  name: 'OpenAIModel',
+  get skip() {
+    return !isCI() && inject('provider-openai').shouldSkip
```
Unshure: Should we remove the `shouldSkip` logic from the `provider-openai` and move it here? We are basically double checking `isCI` at the moment.

test/integ/browser/agent.browser.test.ts
```diff
@@ -33,36 +31,10 @@ const calculatorTool = tool({
...
-        clientConfig: {
-          dangerouslyAllowBrowser: true,
-        },
-      }),
-  },
-]
-
-describe.each(providers)('Agent Browser Tests with $name', async ({ name, createModel }) => {
-  describe(`${name} Browser Integration`, () => {
+describe.each(allProviders)('Agent Browser Tests with $name', async ({ name, skip, createModel }) => {
```
Unshure: nit
```suggestion
describe.each(allProviders)('Agent Browser Tests with $modelProviderName', async ({ modelProviderName, skip, createModel }) => {
```

test/integ/__fixtures__/_setup-global.ts
zastrowm: Will remove; this no longer has anything related to "one time setup"; that was from an earlier revision

test/integ/__fixtures__/_setup-test.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Global setup that runs once before all integration tests and possibly runs in the *parent* process
```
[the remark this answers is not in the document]
zastrowm: Yeah, the same text is an oversight; updating.

parent/host process => `provide` the value
child/client process => `inject` (get) the value

test/integ/__fixtures__/_setup-global.ts
zastrowm: This runs in the parent process of the "tests" - I believe/think that's the vitest main process, but it is the one that has access to the environment variables

test/integ/agent.test.ts
zastrowm: If we do this, then we need the provider property name to be `modelProviderName` which I don't really like as it's more verbose

test/integ/browser/agent.browser.test.ts
zastrowm: Same comment before; this ends up requiring a rename of the property in the provider, which I'm not a fan of

test/integ/vitest.d.ts
```diff
@@ -0,0 +1,17 @@
...
+
+declare module 'vitest' {
+  export interface ProvidedContext {
+    isCI: boolean
+    isBrowser: boolean
+    ['provider-openai']: {
+      shouldSkip: boolean
+      apiKey: string | undefined
+    }
+    ['provider-bedrock']: {
```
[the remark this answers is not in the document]
zastrowm: Part of the problem is the different types that each one needs, but ACK for future expansion

vitest.config.ts
```diff
@@ -71,9 +60,10 @@ export default defineConfig({
           include: ['test/integ/**/*.test.ts'],
           exclude: ['test/integ/**/*.browser.test.ts'],
           name: { label: 'integ-node', color: 'magenta' },
-          testTimeout: 30000,
+          testTimeout: 60 * 1000,
```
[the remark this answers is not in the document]
zastrowm: Yes; I could have made a separate PR for this, but locally I was seeing the same issue, so figured it made sense

PR #361, opened 2025-12-17
Add release-notes SOP for generating release notes

zastrowm:
## Description

Add an agent to generate more detailed release notes for a given release.  This is applicable to both Python & TypeScript though all of the tests/examples that I ran were for Python (since they have more interesting stuff right now)

Flow:

 - Create an issue for release notes to be created for, mentioning the tags/branches you want it for.  Or point it at a GitHub release
 - Agent will confirm with you which items are worth calling out in the two sections (features and major bug fixes)
 - Agent will then generate release notes with code samples (that are verified via tests).  Agent will post the release notes to the issue

The SOP script was modified to:

 - Include the "release notes" & "release-notes" command
 - Allow additional text to be included after the command (so that you can trigger the agents with a comment and explaining text)

## Examples:

 - https://github.com/zastrowm/sdk-typescript/issues/31
 - https://github.com/zastrowm/sdk-typescript/issues/30
 - https://github.com/zastrowm/sdk-typescript/issues/29

## Type of Change

New feature

## Testing

- [X] I ran `npm run check`

## Checklist

----

### DIFF
--- a/.github/scripts/javascript/process-input.cjs
+++ b/.github/scripts/javascript/process-input.cjs
@@ -10,7 +10,7 @@ async function getIssueInfo(github, context, inputs) {
     : context.payload.issue.number.toString();
   const command = context.eventName === 'workflow_dispatch'
     ? inputs.command
-    : (context.payload.comment.body.match(/^\/strands\s*(.*)$/)?.[1]?.trim() || '');
+    : (context.payload.comment.body.match(/^\/strands\s*(.*?)$/m)?.[1]?.trim() || '');
 
   console.log(`Event: ${context.eventName}, Issue ID: ${issueId}, Command: "${command}"`);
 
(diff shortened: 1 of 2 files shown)


.github/scripts/javascript/process-input.cjs
```diff
@@ -86,8 +89,20 @@ module.exports = async (context, github, core, inputs) => {
...
+    let mode;
+    if (command.startsWith('release-notes') || command.startsWith('release notes')) {
+      mode = 'release-notes';
+    } else if (command.startsWith('implement')) {
+      mode = 'implementer';
+    } else if (command.startsWith('refine')) {
+      mode = 'refiner';
+    } else {
+      // Default behavior when no explicit command: PR -> implementer, Issue -> refiner
+      mode = isPullRequest ? 'implementer' : 'refiner';
```
afarntrog: non-blocker: the order of precedence is changing. before this pr, even if the command started with something as long as it was a `isPullRequest` then it would use the `implementer` mode. In this PR the order of precedence is given to the `command.startsWith`. I think it's actually better this way but calling it out.

zastrowm: Great callout; I'm now interested to know how it behaves in a PR if you trigger it using one of the other ones.  

> I think it's actually better this way but calling it out.

 In theory I **do** like the explicit command taking priority, but not I'm not sure in practice it makes sense.   After merge I'll play around and see how I like it

PR #397, opened 2026-01-13
Fix audit errors + split out audit to a separate workflow

zastrowm:
## Description

Ran `npm audit fix` to update dependencies and then split out auditing to a separate workflow

NPM audit is blocking PRs & test statuses because it's done as part of running the tests.  Instead do it as a separate workflow, which also cuts down on redundant checks since it's only done once instead of per OS/node-version (which don't matter because [`npm run audit` is OS-independent](https://docs.npmjs.com/cli/v9/commands/npm-audit)

Also needed to add an `as` cast to fix an MCP issue - this is now issue #398

## Type of Change

Bug fix

## Testing

- [X] I ran `npm run check`

## Checklist

----

### DIFF
--- a/.github/workflows/pr-and-push.yml
+++ b/.github/workflows/pr-and-push.yml
@@ -13,6 +13,13 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  call-security-audit:
+    uses: ./.github/workflows/security-audit.yml
+    permissions:
+      contents: read
+    with:
+      ref: ${{ github.event.pull_request.head.sha }}
+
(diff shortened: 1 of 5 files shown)


zastrowm:
> Should we block prs if the audit check fails? I guess we should because if the pr brings in new dependencies, we should check if they pass our audit check.

I'd rather keep it optional as I don't want all PRs to start failing because main has a dependency failing `npm audit` and I'd rather avoid the "override"s becoming a normal response to that

For new dependencies, I think we can use the npm run audit as a signal IMHO

PR #498, opened 2026-02-05
feat: add ModelThrottledError for rate limiting

dbschmigelski:
This is the same PR as https://github.com/strands-agents/sdk-typescript/pull/436

Thank you to @lseguin1337 who will be given commit credit. We are reraising this because we needed rebase to address a security concern. Unfortunately, we weren't given maintainer edit permission.

### DIFF
--- a/src/models/anthropic.ts
+++ b/src/models/anthropic.ts
@@ -198,17 +198,15 @@ export class AnthropicModel extends Model<AnthropicModelConfig> {
     } catch (unknownError) {
       const error = normalizeError(unknownError)
 
-      // Check for rate limit errors - Anthropic SDK throws errors with status 429
-      // This matches Python SDK behavior: `except anthropic.RateLimitError as e`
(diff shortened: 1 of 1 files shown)


src/errors.ts
```diff
@@ -111,6 +111,46 @@ export class ConcurrentInvocationError extends Error {
...
 }
 
+/**
+ * Error thrown when a model provider returns a throttling or rate limit error.
+ *
+ * This error indicates that the model API has rate limited the request. Users can
+ * handle this error in hooks to implement custom retry strategies using the
+ * `AfterModelCallEvent.retryModelCall` mechanism.
+ *
+ * @example
```
Unshure: nit: I would avoid the example here, it doesnt really add much, and would just add tokens to an agent if it tried to read it.

mehtarac (approved):

mehtarac (approved):

PR #520, opened 2026-02-11
feat: sessionManager - Interface Design & Storage Implementation

JackYPCOnline:
## Description

Implements a complete session management storage layer with pluggable backends for snapshot-based conversation persistence.

### Key Components Added:

**Storage Interface & Implementations:**
- `SnapshotStorage` interface: Defines contract for storage backends with standardized method signatures
- `FileStorage`: File system implementation with atomic writes, directory management, and ENOENT handling
- `S3Storage`: AWS S3 implementation with configurable prefixes, regions, and client options
- `SessionStorage`: Configuration type for pluggable storage backends

**Core APIs:**
- `saveSnapshot(params)`: Persists snapshots to storage with optional latest marking
- `loadSnapshot(params)`: Loads snapshots by ID or latest when undefined
- `listSnapshotIds(params)`: Returns sorted list of available snapshot IDs
- `loadManifest(params)`: Loads metadata with automatic defaults for missing manifests
- `saveManifest(params)`: Persists snapshot metadata and versioning info

**Type System:**
- `Scope`: Support for single agent (`{ kind: 'agent', agentId }`) and multi-agent (`{ kind: 'multiAgent', multiAgentId }`) contexts
- `Snapshot`: Immutable conversation state with schema versioning, messages, state, and timestamps
- `SnapshotManifest`: Metadata tracking with schema version, next ID allocation, and update timestamps
- `SnapshotTriggerParams` & `SnapshotTriggerCallback`: Framework for user-defined snapshot triggers
- `S3StorageConfig`: Configuration options for S3 backend

**Enhanced Features:**
- **Path safety**: Enhanced `validateIdentifier` with strict character restrictions (`[a-z0-9_-]` only)
- **Atomic operations**: File writes use temporary files with atomic rename
- **Error handling**: Custom `SessionError` with cause chaining for better debugging
- **Schema versioning**: Forward compatibility through version tracking
- **Immutable history**: Snapshots stored with zero-padded sequential IDs

**File Structure:**

**File Structure:**
```
sessions/<session_id>/
  scopes/
    agent/<agent_id>/
      snapshots/
        snapshot_latest.json
        manifest.json
        immutable_history/
          snapshot_00001.json
          snapshot_00002.json
```

## Related Issues

https://github.com/orgs/strands-agents/projects/9?pane=issue&itemId=136093391&issue=strands-agents%7Csdk-typescript%7C80

## Type of Change

Bug fix
New feature
Breaking change
Documentation update
Other (please describe):

## Testing

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,7 @@
       },
       "peerDependencies": {
         "@anthropic-ai/sdk": "^0.71.2",
-        "@aws-sdk/client-s3": "^3.893.0",
+        "@aws-sdk/client-s3": "^3.943.0",
         "@google/genai": "^1.40.0",
         "@modelcontextprotocol/sdk": "^1.25.2",
         "openai": "^6.7.0",
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
   },
   "peerDependencies": {
     "@anthropic-ai/sdk": "^0.71.2",
-    "@aws-sdk/client-s3": "^3.893.0",
+    "@aws-sdk/client-s3": "^3.943.0",
     "@google/genai": "^1.40.0",
     "@modelcontextprotocol/sdk": "^1.25.2",
     "openai": "^6.7.0",


src/session/s3-snapshot-storage.ts
```diff
@@ -0,0 +1,180 @@
...
+      snapshotId === null
+        ? this.getLatestSnapshotKey(sessionId, scope)
+        : this.getHistorySnapshotKey(sessionId, scope, snapshotId)
+    return this.readJSON<Snapshot>(key)
+  }
+
+  /**
+   * Lists all snapshot IDs for a session scope
+   */
+  async listSnapshot(sessionId: string, scope: Scope): Promise<number[]> {
```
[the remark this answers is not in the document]
JackYPCOnline: Actually for now we don't need it, but we can add optional parameters in future if there is a use case.
Added TODO section

src/session/file-storage.ts
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  /**
+   * Loads manifest or returns default if not found
+   */
+  async loadManifest(params: { sessionId: string; scope: Scope }): Promise<SnapshotManifest> {
+    const path = this.getPath(params.sessionId, params.scope, MANIFEST)
+    return (
+      (await this.readJSON<SnapshotManifest>(path)) ?? {
+        schemaVersion: 1,
```
JackYPCOnline: Will use a Static varible instead of magic number in next PR

src/__fixtures__/mock-storage-provider.ts
zastrowm: Can you update the PR description with the new APIs that are being introduced?

zastrowm (approved):
If everything here is effectively private (because we don't export it at the top level) and we're committing to follow-ups, I'm okay deferring some of the comments

src/session/storage.ts
```diff
@@ -42,7 +42,7 @@ export interface SnapshotStorage {
   /**
    * Loads a snapshot from storage.
    */
-  loadSnapshot(params: { sessionId: string; scope: Scope; snapshotId: string | null }): Promise<Snapshot | null>
+  loadSnapshot(params: { sessionId: string; scope: Scope; snapshotId: string | undefined }): Promise<Snapshot | null>
```
zastrowm: This should also return undefined I think? Unless we have an existing null convention

src/types/validation.ts
```diff
@@ -12,3 +12,19 @@ export function ensureDefined<T>(value: T | null | undefined, fieldName: string)
...
+
+/**
+ * Validates that an identifier contains only allowed characters.
+ * Allowed characters: lowercase letters (a-z), numbers (0-9), hyphens (-), and underscores (_)
+ *
+ * @param id - The identifier to validate
+ * @returns The validated identifier
+ * @throws Error if identifier contains invalid characters
+ */
+export function validateIdentifier(id: string): string {
```
pgrayy: This seems a bit misplaced in the types module. I would consider moving to session.

src/session/storage.ts
```diff
@@ -0,0 +1,72 @@
...
+ *         immutable_history/
+ *           snapshot_00001.json
+ *           snapshot_00002.json
+ * ```
+ */
+export interface SnapshotStorage {
+  /**
+   * Persists a snapshot to storage.
+   */
+  saveSnapshot(params: { sessionId: string; scope: Scope; isLatest: boolean; snapshot: Snapshot }): Promise<void>
```
pgrayy: Do we use the inline typing pattern elsewhere in code? It might be worth creating dedicated param types. Also, we you be able to add docstrings explaining these parameters.

src/session/file-storage.ts
```diff
@@ -0,0 +1,167 @@
...
+   * @param baseDir - Base directory path for storing snapshots
+   */
+  constructor(baseDir: string) {
+    this._baseDir = baseDir
+  }
+
+  /**
+   * Generates file path for session scope snapshots
+   */
+  private getPath(sessionId: string, scope: Scope, filename: string): string {
```
pgrayy: Convention for private methods is to add _ prefix.

src/session/storage.ts
JackYPCOnline: I'd keep inline types for these storage methods because:

Parameters are most likly unique, and they're implementation details, users can't access it. I created dedicated types  `SnapshotTriggerParams` since user will use it.

src/session/file-storage.ts
```diff
@@ -0,0 +1,167 @@
...
+    scope: Scope
+    isLatest: boolean
+    snapshot: Snapshot
+  }): Promise<void> {
+    await this.writeJSON(
+      this.getHistorySnapshotPath(params.sessionId, params.scope, params.snapshot.snapshotId),
+      params.snapshot
+    )
+    if (params.isLatest) {
+      await this.writeJSON(this.getLatestSnapshotPath(params.sessionId, params.scope), params.snapshot)
```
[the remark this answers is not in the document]
JackYPCOnline: Yes, this makes sense and It's on my plan. For basic version I recommand keep it as is. I will iterate it after we merged session manager impl

src/session/file-storage.ts
```diff
@@ -0,0 +1,167 @@
...
+   */
+  constructor(baseDir: string) {
+    this._baseDir = baseDir
+  }
+
+  /**
+   * Generates file path for session scope snapshots
+   */
+  private getPath(sessionId: string, scope: Scope, filename: string): string {
+    validateIdentifier(sessionId)
```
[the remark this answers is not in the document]
JackYPCOnline: This is where we do in python SDK, I think we should add validation here since FileStorage/S3Storage is user instanciated? This will also prevent secruity issues?

src/session/s3-storage.ts
```diff
@@ -0,0 +1,198 @@
...
+  /** AWS region (default: us-east-1). Cannot be used with s3Client */
+  region?: string
+  /** Pre-configured S3 client. Cannot be used with region */
+  s3Client?: S3Client
+}
+
+/**
+ * S3-based implementation of SnapshotStorage for persisting session snapshots
+ */
+export class S3Storage implements SnapshotStorage {
```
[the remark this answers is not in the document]
pgrayy: Okay great, this makes sense.

src/session/file-storage.ts
```diff
@@ -0,0 +1,167 @@
...
+  async listSnapshotIds(params: { sessionId: string; scope: Scope }): Promise<string[]> {
+    const dirPath = this.getPath(params.sessionId, params.scope, IMMUTABLE_HISTORY)
+    try {
+      const files = await fs.readdir(dirPath)
+      return files
+        .map((file) => file.match(SNAPSHOT_REGEX)?.[1])
+        .filter((id): id is string => id !== undefined)
+        .sort((a, b) => parseInt(a) - parseInt(b))
+    } catch (error: unknown) {
+      if (error && typeof error === 'object' && 'code' in error && error.code === 'ENOENT') {
```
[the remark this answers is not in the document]
pgrayy: Note there is a duplication of this already down below in readJson. Since this is private, can be handled in a follow up.

src/session/types.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Manifest tracks snapshot metadata and ID allocation.
+ * Stored alongside snapshots to manage versioning.
+ */
+export interface SnapshotManifest {
+  /** Schema version for forward/backward compatibility */
+  schemaVersion: string
+  /** Next available snapshot ID for allocation */
+  nextSnapshotId: string
+  /** ISO 8601 timestamp of last manifest update */
+  updatedAt: string
```
pgrayy: Calling out so it doesn't get lost, but I think we should switch all the At fields to Date and use the JSON.stringify replacer to convert to ISO string when needed.

pgrayy (approved):

PR #544, opened 2026-02-17
feat: wrap raw data objects in event wrappers for agent stream

### LINKED ISSUE #288: Ensure consistency between streamed events & hooks
We have an inspiration goal to unify:

 - streamed-events (coming from `Agent.stream`) 
 - (and) hooks which are subscription based
 
There are currently a couple of gaps yet between the two that we should resolve:

 - Naming - both events are named with an Event suffix, but due to collisions, some hooks have a Hook suffix
 - Missing hooks:
    - `ModelStreamEventHook` - intentionally removed for now until we resolve how to emit this one. It has two core problems:
       - If we simply emit ModelStreamEventHook then we emit duplicate model events (one wrapped in ModelStreamEventHook, one raw)
       - It includes both content blocks *and* model events, which is odd and nothing else does that right now
    - `ContentBlocks` - both tools and ModelContentBlocks
    - Tool Streaming Events
 - MessageAddedEvent - consider if we can stop special casing this one

There might be additional differences that should be explored. As part of this task, we should take a holistic look at what events are emitted and what hooks are emitted and see if they align or how they should align.

### LINKED ISSUE #454: [V1] Protocols - Streamed events should be Wrapper Events, not raw Data
### Summary

Right now, TypeScript is yielding an odd combination of streaming events.  Some are "a snapshot in time" - like delta events; while others are raw data - like Message and ContentBlock.  Emitting raw data couples the streamed event to the underlying data, creating a tight coupling between what we stream and the data-structured, preventing us from adding helpers or additional data that isn't tied to the data itself.

We should be wrapping all raw data (like Message & ContentBlock) in an explicit event so that we can continue to evolve the api.

Related issues: #23, #288

https://github.com/zastrowm/sdk-typescript/issues/23#issuecomment-3638995110 has specific examples:

1. LIFECYCLE EVENTS ✅ (properly describe state transitions)
    - BeforeInvocationEvent - "agent invocation is starting"
    - AfterInvocationEvent - "agent invocation completed"
    - BeforeModelCallEvent - "about to call model"
    - AfterModelCallEvent - "model call completed"
    - ModelMessageStartEvent - "model message streaming started"
    - ModelMessageStopEvent - "model message streaming stopped"
    - ModelContentBlockStartEvent - "content block streaming started"
    - ModelContentBlockStopEvent - "content block streaming stopped"
    - BeforeToolsEvent, AfterToolsEvent, BeforeToolCallEvent, AfterToolCallEvent

2. DELTA EVENTS ✅ (properly wrapped incremental updates)
   - ModelContentBlockDeltaEvent

3. METADATA EVENTS ✅ (properly wrapped metadata)
    - ModelMetadataEvent - Contains usage, metrics, trace
    - MessageAddedEvent - "message was added" (contains the message)

4. DATA OBJECTS ❌ (emitted directly WITHOUT event wrapper) **Problem: These are complete data objects, not events!**
    - ContentBlock (TextBlock, ToolUseBlock, ReasoningBlock)
    - ToolResultBlock
    - AgentResult

dbschmigelski:
## Summary

Resolves #454, resolves #288

`agent.stream()` yielded a mix of `HookEvent` instances and raw data objects (`ContentBlock`, `ToolResultBlock`, `AgentResult`, `ModelStreamEvent`, `ToolStreamEvent`). This PR wraps all raw yields in event classes. **Every member of `AgentStreamEvent` is now a `HookEvent`**, providing consistent structure, agent context, and hookability across the entire stream.

### New event classes

All extend `HookEvent`, all are both yielded in the stream and hookable:

- `ContentBlockCompleteEvent` — wraps completed content blocks (TextBlock, ToolUseBlock, ReasoningBlock)
- `ModelMessageEvent` — wraps the fully assembled model message after streaming completes
- `ToolResultEvent` — wraps tool execution results
- `ToolStreamObserverEvent` — wraps tool streaming progress events
- `AgentResultEvent` — wraps the final agent result

### `ModelStreamEventHook` → `ModelStreamObserverEvent`

Renamed for naming consistency (all events now use "Event" suffix). Resolved the two concerns from #288 that blocked public export:

1. **Mixed content types** — Previously typed `ModelStreamEvent | ContentBlock`, bundling streaming deltas and completed blocks in one event. Now narrowed to just `ModelStreamEvent`. Completed blocks have their own `ContentBlockCompleteEvent`.

2. **Duplicate events / raw data in stream** — Previously `ModelStreamEvent` appeared raw in the stream alongside `HookEvent` instances. Now wrapped in `ModelStreamObserverEvent`, consistent with all other wrapping.

Now publicly exported.

### Design: wrapping at the agent boundary

Raw data types from lower layers (`ModelStreamEvent` from the model, `ToolStreamEvent` from tools) don't know about agents or hooks. Rather than modifying these lower-layer types to extend `HookEvent` (which would leak agent concerns into the model/tool layer), we wrap them at the agent boundary:

- `ModelStreamObserverEvent` wraps `ModelStreamEvent` — the agent iterates `model.streamAggregated()` and wraps each delta
- `ToolStreamObserverEvent` wraps `ToolStreamEvent` — the agent iterates `tool.stream()` manually (replacing `yield*` with explicit iteration) and wraps each progress event

This keeps the tool authoring interface unchanged — tools construct `ToolStreamEvent` without any `agent` field or `HookEvent` dependency.

### Design: why `ModelStreamObserverEvent` and `ContentBlockCompleteEvent` are separate

The model's `streamAggregated()` yields two kinds of output: `ModelStreamEvent` (transient streaming deltas — partial data arriving while the model generates) and `ContentBlock` (fully assembled results after all deltas accumulate). These represent different granularities with different semantics, so they are wrapped in distinct event classes rather than combined.

### Design: `ContentBlockCompleteEvent` vs `ModelMessageEvent`

These serve different granularities:

- `ContentBlockCompleteEvent` fires as each individual content block finishes during streaming (e.g., a TextBlock completes while a ToolUseBlock is still accumulating). It gives consumers the assembled block immediately, without having to accumulate deltas themselves.
- `ModelMessageEvent` fires once after all blocks are assembled into the complete message.

The consumer already gets `modelContentBlockStopEvent` (via `ModelStreamObserverEvent`) which signals a block ended, and `ModelMessageEvent` which has all finished blocks. `ContentBlockCompleteEvent` sits in between — it provides the fully assembled block as soon as it's ready, without waiting for the entire message.

Python has no equivalent — it silently folds completed blocks into message state, and consumers only see them via `ModelMessageEvent`. The reason `ContentBlockCompleteEvent` exists in TS is that `model.streamAggregated()` yields completed `ContentBlock` objects alongside `ModelStreamEvent` deltas. Before this PR, those raw `ContentBlock`s were yielded directly into the stream. We wrapped them in `ContentBlockCompleteEvent` to maintain the "everything is a HookEvent" invariant, preserving data that was already in the stream. Removing it entirely (to match Python) would be a separate breaking change.

### `MessageAddedEvent` special-casing removed

Previously, `MessageAddedEvent` hooks were invoked early inside `_appendMessage()` and then skipped in the `stream()` dispatch loop. This was unnecessary — no hook subscriber depends on early invocation (the conversation manager uses `AfterInvocationEvent`/`AfterModelCallEvent`, not `MessageAddedEvent`), and `yield*` delegation ensures hooks run before the generator resumes regardless. `MessageAddedEvent` now follows the same dispatch path as every other event.

### Breaking changes

Consumers filtering stream events by `type` need to update:

| Before (raw) | After (wrapped) |
|---|---|
| `'modelContentBlockDeltaEvent'` etc. | `'modelStreamObserverEvent'` → access `.event` |
| `'textBlock'` / `'toolUseBlock'` / `'reasoningBlock'` | `'contentBlockCompleteEvent'` → access `.contentBlock` |
| `'toolStreamEvent'` | `'toolStreamObserverEvent'` → access `.toolStreamEvent` |
| `'toolResultBlock'` | `'toolResultEvent'` → access `.toolResult` |
| Raw `AgentResult` | `'agentResultEvent'` → access `.result` |
| `'modelStreamEventHook'` | `'modelStreamObserverEvent'` |

### Python SDK alignment

| Python SDK | TypeScript (this PR) | Notes |
|---|---|---|
| `ModelMessageEvent` | `ModelMessageEvent` | Direct equivalent |
| `ToolResultEvent` | `ToolResultEvent` | Direct equivalent |
| `AgentResultEvent` | `AgentResultEvent` | Direct equivalent |
| `ToolStreamEvent` | `ToolStreamObserverEvent` | TS wraps in HookEvent at agent boundary |
| `TextStreamEvent` / `ModelStreamChunkEvent` etc. | `ModelStreamObserverEvent` | TS wraps in HookEvent at agent boundary |
| (no equivalent) | `ContentBlockCompleteEvent` | TS-only. Python folds completed blocks silently into message state; they only surface as part of `ModelMessageEvent`. TS exposes them individually as they complete because `streamAggregated()` already yielded them — we wrapped rather than dropped. |

## Test plan

- [x] Full quality gate passes (`npm run check`) — 844 tests, 89.6% coverage
- [x] New events have unit tests (type discriminator, readonly, reverse callbacks)
- [x] `ModelStreamObserverEvent` test verifies it's both yielded in stream and fired as hook
- [x] Integration tests updated (notebook, file-editor)
- [ ] Manual integration test against live model

### DIFF
--- a/src/hooks/events.ts
+++ b/src/hooks/events.ts
@@ -40,7 +40,7 @@ import type { ModelStreamEvent } from '../models/streaming.js'
  *   *Completion events* — wrap finished data after processing completes.
  *   - Naming: descriptive `<Noun>Event`, payload field matches data type
  *     (`.result` for results, `.message` for messages, `.contentBlock` for content blocks).
- *   - Examples: {@link ContentBlockCompleteEvent}, {@link ModelMessageEvent},
+ *   - Examples: {@link ContentBlockEvent}, {@link ModelMessageEvent},
  *     {@link ToolResultEvent}, {@link AgentResultEvent}
  *
  * ## Field naming conventions
@@ -278,7 +278,7 @@ export class AfterModelCallEvent extends HookableEvent {
 /**
  * Event triggered for each streaming event from the model.
  * Wraps a {@link ModelStreamEvent} (transient streaming delta) during model inference.
- * Completed content blocks are handled separately by {@link ContentBlockCompleteEvent}
+ * Completed content blocks are handled separately by {@link ContentBlockEvent}
  * because they represent different granularities: partial deltas vs fully assembled results.
  */
 export class ModelStreamUpdateEvent extends HookableEvent {
@@ -303,8 +303,8 @@ export class ModelStreamUpdateEvent extends HookableEvent {
  * These represent different granularities with different semantics, so they are
  * wrapped in distinct event classes rather than combined into a single event.
  */
-export class ContentBlockCompleteEvent extends HookableEvent {
-  readonly type = 'contentBlockCompleteEvent' as const
+export class ContentBlockEvent extends HookableEvent {
+  readonly type = 'contentBlockEvent' as const
   readonly agent: AgentData
   readonly contentBlock: ContentBlock
 
--- a/src/agent/agent.ts
+++ b/src/agent/agent.ts
@@ -43,7 +43,7 @@ import {
   HookableEvent,
   MessageAddedEvent,
   ModelStreamUpdateEvent,
-  ContentBlockCompleteEvent,
+  ContentBlockEvent,
   ModelMessageEvent,
   ToolResultEvent,
   AgentResultEvent,
@@ -567,7 +567,7 @@ export class Agent implements AgentData {
    * - **ModelStreamEvent**: Transient streaming deltas (partial data while generating).
    *   Wrapped in {@link ModelStreamUpdateEvent} before yielding.
    * - **ContentBlock**: Fully assembled results (after all deltas accumulate).
-   *   Wrapped in {@link ContentBlockCompleteEvent} before yielding.
+   *   Wrapped in {@link ContentBlockEvent} before yielding.
    *
    * These are separate event classes because they represent different granularities
    * (partial deltas vs finished blocks). Both are yielded in the stream and hookable.
@@ -590,8 +590,8 @@ export class Agent implements AgentData {
         // ModelStreamEvent: wrap in ModelStreamUpdateEvent
         yield new ModelStreamUpdateEvent({ agent: this, event })
       } else {
-        // ContentBlock: wrap in ContentBlockCompleteEvent
-        yield new ContentBlockCompleteEvent({ agent: this, contentBlock: event })
+        // ContentBlock: wrap in ContentBlockEvent
+        yield new ContentBlockEvent({ agent: this, contentBlock: event })
       }
       result = await streamGenerator.next()
     }
--- a/src/hooks/__tests__/events.test.ts
+++ b/src/hooks/__tests__/events.test.ts
@@ -11,7 +11,7 @@ import {
   BeforeToolsEvent,
   MessageAddedEvent,
   ModelStreamUpdateEvent,
-  ContentBlockCompleteEvent,
+  ContentBlockEvent,
   ModelMessageEvent,
   ToolResultEvent,
   ToolStreamUpdateEvent,
@@ -382,14 +382,14 @@ describe('ModelStreamUpdateEvent', () => {
   })
 })
 
-describe('ContentBlockCompleteEvent', () => {
+describe('ContentBlockEvent', () => {
   it('creates instance with correct properties', () => {
     const agent = new Agent()
     const contentBlock = new TextBlock('Hello')
-    const event = new ContentBlockCompleteEvent({ agent, contentBlock })
+    const event = new ContentBlockEvent({ agent, contentBlock })
 
     expect(event).toEqual({
-      type: 'contentBlockCompleteEvent',
+      type: 'contentBlockEvent',
       agent: agent,
       contentBlock: contentBlock,
     })
--- a/src/hooks/index.ts
+++ b/src/hooks/index.ts
@@ -26,7 +26,7 @@ export {
   BeforeModelCallEvent,
   AfterModelCallEvent,
   ModelStreamUpdateEvent,
-  ContentBlockCompleteEvent,
+  ContentBlockEvent,
   ModelMessageEvent,
   ToolResultEvent,
   ToolStreamUpdateEvent,
--- a/src/index.ts
+++ b/src/index.ts
@@ -161,7 +161,7 @@ export {
   AfterModelCallEvent,
   BeforeToolsEvent,
   AfterToolsEvent,
-  ContentBlockCompleteEvent,
+  ContentBlockEvent,
   ModelMessageEvent,
   ToolResultEvent,
   ToolStreamUpdateEvent,
--- a/src/types/agent.ts
+++ b/src/types/agent.ts
@@ -11,7 +11,7 @@ import type {
   AfterToolCallEvent,
   MessageAddedEvent,
   ModelStreamUpdateEvent,
-  ContentBlockCompleteEvent,
+  ContentBlockEvent,
   ModelMessageEvent,
   ToolResultEvent,
   ToolStreamUpdateEvent,
@@ -110,7 +110,7 @@ export class AgentResult {
  */
 export type AgentStreamEvent =
   | ModelStreamUpdateEvent
-  | ContentBlockCompleteEvent
+  | ContentBlockEvent
   | ModelMessageEvent
   | ToolStreamUpdateEvent
   | ToolResultEvent
--- a/test/integ/file-editor.test.node.ts
+++ b/test/integ/file-editor.test.node.ts
@@ -50,9 +50,7 @@ describe.skipIf(bedrock.skip)('FileEditor Tool Integration', () => {
     const { items: events } = await collectGenerator(agent.stream(`View the file at ${testFile}`))
 
     // The agent should have received the file content
-    const textBlocks = events.filter(
-      (e: any) => e.type === 'contentBlockCompleteEvent' && e.contentBlock.type === 'textBlock'
-    )
+    const textBlocks = events.filter((e: any) => e.type === 'contentBlockEvent' && e.contentBlock.type === 'textBlock')
     expect(textBlocks.length).toBeGreaterThan(0)
   }, 60000)
 
@@ -99,9 +97,7 @@ describe.skipIf(bedrock.skip)('FileEditor Tool Integration', () => {
     expect(toolResults.length).toBeGreaterThan(0)
 
     // The model should have handled the error gracefully
-    const textBlocks = events.filter(
-      (e: any) => e.type === 'contentBlockCompleteEvent' && e.contentBlock.type === 'textBlock'
-    )
+    const textBlocks = events.filter((e: any) => e.type === 'contentBlockEvent' && e.contentBlock.type === 'textBlock')
     expect(textBlocks.length).toBeGreaterThan(0)
   }, 60000)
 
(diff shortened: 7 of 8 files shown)


src/__fixtures__/mock-hook-provider.ts
```diff
@@ -8,7 +8,12 @@ import {
   AfterToolCallEvent,
   BeforeModelCallEvent,
   AfterModelCallEvent,
-  ModelStreamEventHook,
+  ModelStreamObserverEvent,
+  ContentBlockCompleteEvent,
+  ModelMessageEvent,
+  ToolResultEvent,
+  ToolStreamObserverEvent,
```
zastrowm: We should name shop this; Observer seems off to me; were there any other good alternatives you came up with?

`ToolStreamObserverEvent` -> `ToolStreamDeltaEvent`; `ToolStreamedEvent`; `ToolStreamEmitEvent`

`ModelStreamObserverEvent` ->  `ModelDeltaStreamEvent`, `ModelStreamEmitEvent`

src/hooks/events.ts
```diff
@@ -219,23 +219,121 @@ export class AfterModelCallEvent extends HookEvent {
...
+    this.agent = data.agent
+    this.toolStreamEvent = data.toolStreamEvent
+  }
+}
+
+/**
+ * Event triggered as the final event in the agent stream.
+ * Wraps the agent result containing the stop reason and last message.
+ */
+export class AgentResultEvent extends HookEvent {
```
zastrowm: For naming; can we come up with a set of rules and use that to name them?

- `ToolStreamObserverEvent` seems to indicate some sort of observation but `Observer` is odd naming
- `ModelMessageEvent` comes off as this is an event for a thing, but no comment as to a "lifecycle"
- `AgentResultEvent` does match ^ `ModelMessageEvent`
- `ContentBlockCompleteEvent` comes off as "This thing is now complete, here's an event"
- `BeforeToolCallEvent` is a lifecycle before/after thing
- `MessageAddedEvent` is a lifecycle after thing

I can't come up with a good rule that matches all of these, but there's something *inconsistent* right now.

src/hooks/events.ts
```diff
@@ -219,23 +219,121 @@ export class AfterModelCallEvent extends HookEvent {
...
+/**
+ * Event triggered for each streaming progress event from a tool during execution.
+ * Wraps a {@link ToolStreamEvent} with agent context, keeping the tool authoring
+ * interface unchanged — tools construct `ToolStreamEvent` without knowledge of agents
+ * or hooks, and the agent layer wraps them at the boundary.
+ *
+ * Both yielded in the agent stream and hookable, consistent with
+ * {@link ModelStreamObserverEvent} which wraps model streaming events the same way.
+ */
+export class ToolStreamObserverEvent extends HookEvent {
```
zastrowm: For TS 1.0; we should consider renaming `HookEvent` maybe.  `LifecycleEvent`; `ObservableEvent`.

`HookEvent` doesn't make as much sense if you're just streaming it IMHO

src/hooks/events.ts
```diff
@@ -219,23 +219,121 @@ export class AfterModelCallEvent extends HookEvent {
...
-export class ModelStreamEventHook extends HookEvent {
-  readonly type = 'modelStreamEventHook' as const
+export class ModelStreamObserverEvent extends HookEvent {
+  readonly type = 'modelStreamObserverEvent' as const
   readonly agent: AgentData
-  readonly event: ModelStreamEvent | ContentBlock
+  readonly event: ModelStreamEvent
 
-  constructor(data: { agent: AgentData; event: ModelStreamEvent | ContentBlock }) {
+  constructor(data: { agent: AgentData; event: ModelStreamEvent }) {
```
zastrowm: Not a blocker, but food for thought; do we actually need `agent` on HookEvent? 

Could hooks just have an agent associated with it when it's fired and it becomes an additional parameter to callbacks?  E.g. there's no need for each event to carry around.

I think the primary thing this would unlock is that events could be defined without having to provide an agent - making HookEvent even lower level. Do we gain any benefits (maybe we don't have to wrap as many things) by doing so?

src/hooks/events.ts
```diff
@@ -219,23 +219,121 @@ export class AfterModelCallEvent extends HookEvent {
...
  *
- * Currently private pending https://github.com/strands-agents/sdk-typescript/issues/288
+ * This event wraps only {@link ModelStreamEvent} (transient streaming deltas — partial
+ * data arriving while the model generates). Completed content blocks are handled
+ * separately by {@link ContentBlockCompleteEvent} because they represent different
+ * granularities: partial deltas vs fully assembled results.
  */
-export class ModelStreamEventHook extends HookEvent {
-  readonly type = 'modelStreamEventHook' as const
+export class ModelStreamObserverEvent extends HookEvent {
```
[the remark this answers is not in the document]
zastrowm: > I still question whether we want everything hookable by default

Got a two liner as to why, or pros & cons?  I'm not necessarily convinced that StreamEvent = HookEvent, but I do believe that anything that is streamable should be observable by hooks, and TBH vice-versa

src/hooks/events.ts
```diff
@@ -1,14 +1,72 @@
...
+ * | `.message`     | Message object                               |
+ * | `.contentBlock`| Content block object                         |
+ */
+
+/**
+ * Base class for all events yielded by `agent.stream()`.
+ * Carries no hookability — subclasses that should be hookable extend {@link HookableEvent} instead.
  */
-export abstract class HookEvent {
+export abstract class StreamEvent {}
```
dbschmigelski: possible discussion regarding moving `abstract readonly type: string` here

zastrowm: I think we should do it

Eventually I also anticipate that we'll want serialization methods on here like we do for `Message`, but that's a future concern

src/hooks/events.ts
```diff
@@ -1,14 +1,72 @@
...
+ * Two sub-categories:
+ *
+ *   *Update events* — wrap transient streaming data from lower layers.
+ *   - Naming: `<Source>StreamUpdateEvent`, payload field: `.event`
+ *   - Examples: {@link ModelStreamUpdateEvent}, {@link ToolStreamUpdateEvent}
+ *
+ *   *Completion events* — wrap finished data after processing completes.
+ *   - Naming: descriptive `<Noun>Event`, payload field matches data type
+ *     (`.result` for results, `.message` for messages, `.contentBlock` for content blocks).
+ *   - Examples: {@link ContentBlockCompleteEvent}, {@link ModelMessageEvent},
```
zastrowm: `ContentBlockCompleteEvent` stands out to me; `ContentBlockEvent` would make it fit better I think.  Is that possible or is it already used?

Otherwise I really like these rules

src/hooks/events.ts
```diff
@@ -219,28 +277,120 @@ export class AfterModelCallEvent extends HookEvent {
...
+ * Event triggered when a content block completes during model inference.
+ * Wraps completed content blocks (TextBlock, ToolUseBlock, ReasoningBlock) from model streaming.
+ * This is intentionally separate from {@link ModelStreamUpdateEvent}. The model's
+ * `streamAggregated()` yields two kinds of output: {@link ModelStreamEvent} (transient
+ * streaming deltas — partial data arriving while the model generates) and
+ * {@link ContentBlock} (fully assembled results after all deltas accumulate).
+ * These represent different granularities with different semantics, so they are
+ * wrapped in distinct event classes rather than combined into a single event.
+ */
+export class ContentBlockCompleteEvent extends HookableEvent {
```
zastrowm: Per my other comment, can this just be ContentBlockEvent?

zastrowm (changes requested):

zastrowm (approved):

src/models/streaming.ts
```diff
@@ -23,6 +23,25 @@ export type ModelStreamEvent =
   | ModelMessageStopEventData
   | ModelMetadataEventData
 
+/** Set of all ModelStreamEvent type discriminators. */
+const modelStreamEventTypes: ReadonlySet<string> = new Set<ModelStreamEvent['type']>([
```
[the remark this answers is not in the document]
dbschmigelski: Yeah exactly I do think the fact that the check with an isModelStreamEvent() type guard backed by a Set<ModelStreamEvent['type']> which is exhaustive over the union's type discriminators, so if a new event type is added to ModelStreamEvent without updating the set, TypeScript will flag it as a compile error.

PR #545, opened 2026-02-18
fix: make callback the first parameter for adding hook callback

Unshure:
## Description
**Breaking Change** This reorders the parameters for the HookProvider `addCallback` method to align with the feedback in the python sdk pr: https://github.com/strands-agents/sdk-python/pull/1706

## Related Issues

N/A

## Documentation PR

TODO

## Type of Change

Breaking change

## Testing

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -321,12 +321,12 @@ When testing hook behavior, you **MUST** use `agent.hooks.addCallback()` for reg
 // ✅ CORRECT - Use agent.hooks.addCallback() for single callbacks
 const agent = new Agent({ model, tools: [tool] })
 
-agent.hooks.addCallback(BeforeToolCallEvent, (event: BeforeToolCallEvent) => {
+agent.hooks.addCallback((event: BeforeToolCallEvent) => {
   event.toolUse = {
     ...event.toolUse,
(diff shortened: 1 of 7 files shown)


zastrowm:
> > Can we just remove HookRegistry and migrate to plugins? IMHO there's no need for `agent.hooks` in typescript if we're adding callbacks and plugins to the top level.
> > https://github.com/strands-agents/sdk-typescript/blob/55d29d5540653ffa9c767269202c0f518ea1d053/src/hooks/registry.ts#L16
> > 
> > is only those 3 methods and I'd rather avoid a breaking change and a migration later to Plugins if we can do it all in one go
> 
> @zastrowm Are we adding `callbacks` to the top level? I thought we would still expose `agent.hook` just like in python?

In TS, we don't expose the ability to fire hooks and kept more apis as an implementation detail. I haven't seen any usage of folks firing custom events yet, and given the number of changes of hooks went through in python I figured it's better not to expose these items until we have use cases

So in TS, you already don't have access to callbacks; the only apis we have are addCallback, addHook, removeHook; and all 3 are effectively not needed once we have addHook and plugins on agent

PR #568, opened 2026-02-24
feat: add CitationsBlock for document citation support

### LINKED ISSUE #487: [V1] Protocols - Content Blocks
## Summary

Add support for missing content block input types (e.g., ImageContent) to enable multimodal agent interactions.

## Usage

```typescript
import { Agent, ImageContent, TextContent } from '@strands-agents/sdk'

const agent = new Agent({ model })

const result = await agent.invoke([
  new TextContent('What is in this image?'),
  new ImageContent({
    source: {
      type: 'base64',
      mediaType: 'image/png',
      data: base64ImageData,
    },
  }),
])
```

## Documentation

https://strandsagents.com/latest/documentation/docs/user-guide/concepts/agents/prompts/

## Reference

https://github.com/strands-agents/sdk-python/blob/main/src/strands/types/content.py

dbschmigelski:
## Description
Resolves #487

Adds `CitationsBlock` for document citation support, achieving parity with the Python SDK (strands-agents/sdk-python#631). Citations are returned by models when document citations are enabled via `citations: { enabled: true }` on `DocumentBlock` inputs. They are output-only blocks that appear in conversation history and must be serializable.

---

### Provider-agnostic type design

Rather than mirroring Bedrock's wire format directly, citation types use SDK-native patterns. `CitationLocation` is a discriminated union with a `type` field — matching how `ContentBlockDelta` already works — so consumers can use `switch(location.type)` instead of `'key' in obj` checks. The Bedrock-specific object-key format (`{ documentChar: { ... } }`) is mapped bidirectionally in `bedrock.ts`, keeping the public types provider-agnostic. This means other providers can add citation support by implementing their own mapping without changing the shared types.

We also renamed Bedrock's `searchResultLocation` to `searchResult` in the SDK type since the "Location" suffix is redundant inside `CitationLocation`.

### Union type convention for extensibility

`CitationSourceContent` and `CitationGeneratedContent` are modeled as `type` aliases rather than interfaces. This follows the same pattern as `DocumentSourceData` in `media.ts` — when a type maps to a Bedrock API UNION, we use a `type` alias so additional variants (e.g., `{ image: ... }`) can be added as a union expansion rather than requiring interface changes. This convention is now documented in `AGENTS.md`.

### Required fields over optional

Bedrock's API docs mark most citation fields as "Not Required", but we kept them required in the SDK types. Integration tests confirm Bedrock always returns `documentIndex`, `start`, and `end` for document locations. Starting with required fields is the safer choice — relaxing to optional is non-breaking, but tightening later would break consumers.

### Multi-turn citation support

Citations are filtered and sent back to Bedrock in conversation history, matching the Python SDK behavior. This enables multi-turn conversations where the model can reference previous citations. The formatting pipeline in `_formatContentBlock` strips to Bedrock-supported fields only (`location`, `sourceContent.text`, `title`) and handles the SDK-to-Bedrock type mapping on the way out.

### Python SDK alignment

| Python SDK | TypeScript (this PR) |
|---|---|
| CitationsContentBlock in strands/types/citations.py | CitationsBlock in src/types/citations.ts |
| citationsContent field in ContentBlock | citationsContent wrapper in ContentBlockData union |
| Filtered and sent back to Bedrock for multi-turn | Same — filtered to Bedrock-supported fields and sent back |
| Object-key citation locations (Bedrock wire format) | type-field discriminated union (mapped in bedrock.ts) |

## Related Issues

https://github.com/strands-agents/sdk-python/pull/631

## Type of Change

Bug fix
New feature
Breaking change
Documentation update
Other (please describe):

## Testing

- [ ] I ran `npm run check`

## Checklist
- [ ] I have read the CONTRIBUTING document
- [ ] My changes generate no new warnings

----

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2014,11 +2014,10 @@
       }
     },
     "node_modules/@hono/node-server": {
-      "version": "1.19.9",
-      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.9.tgz",
-      "integrity": "sha512-vHL6w3ecZsky+8P5MD+eFfaGTyCeOHUIFYMGpQGbrBTSmNNoxv0if69rEZ5giu36weC5saFuznL411gRX7bJDw==",
+      "version": "1.19.11",
+      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.11.tgz",
+      "integrity": "sha512-dr8/3zEaB+p0D2n/IUrlPF1HZm586qgJNXK1a9fhg/PzdtkK7Ksd5l312tJX2yBuALqDYBlG20QEbayqPyxn+g==",
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18.14.1"
       },
@@ -2191,6 +2190,7 @@
       "integrity": "sha512-3giAOQvZiH5F9bMlMiv8+GSPMeqg0dbaeo58/0SlA9sxSqZhnUtxzX9/2FzyhS9sWQf5S0GJE0AKBrFqjpeYcg==",
       "dev": true,
       "license": "Apache-2.0",
+      "peer": true,
       "engines": {
         "node": ">=8.0.0"
       }
@@ -3700,6 +3700,7 @@
       "integrity": "sha512-klQbnPAAiGYFyI02+znpBRLyjL4/BrBd0nyWkdC0s/6xFLkXYQ8OoRrSkqacS1ddVxf/LDyODIKbQ5TgKAf/Fg==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@typescript-eslint/scope-manager": "8.56.1",
         "@typescript-eslint/types": "8.56.1",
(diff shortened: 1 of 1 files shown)


src/types/messages.ts
```diff
@@ -115,6 +117,7 @@ export type ContentBlockData =
   | { image: ImageBlockData }
   | { video: VideoBlockData }
   | { document: DocumentBlockData }
+  | { citationsContent: CitationsBlockData }
```
pgrayy: Nit: Maybe consider calling citations for simplicity. We have guardContent though so there is precedent for the Content suffix.

src/models/model.ts
```diff
@@ -228,23 +233,28 @@ export abstract class Model<T extends BaseModelConfig = BaseModelConfig> {
...
-                accumulatedToolInput += event.delta.input
-                break
-              case 'reasoningContentDelta':
-                if (event.delta.text) accumulatedReasoning.text = (accumulatedReasoning.text ?? '') + event.delta.text
-                if (event.delta.signature) accumulatedReasoning.signature = event.delta.signature
-                if (event.delta.redactedContent) accumulatedReasoning.redactedContent = event.delta.redactedContent
-                break
+          case 'modelContentBlockDeltaEvent': {
+            const delta = event.delta
+            if (delta.type === 'textDelta') {
```
pgrayy: How come the switch to if/else?

pgrayy (approved):

Issue #395, opened 2026-01-13
[FEATURE] Tools can return DocumentBlock, ImageBlock, VideoBlock


dpruessner:
### Problem Statement

Multi-modal agents need tools that can return rich media content (documents, images, videos) directly to the model. The Bedrock Conversation API supports DocumentBlock, ImageBlock, and VideoBlock in tool results, enabling more efficient processing than converting media to JSON or text. For example, returning HTML as a DocumentBlock is safer and more performant than embedding it as a string.

### Proposed Solution

This PR enables tools to return `DocumentBlock`, `ImageBlock`, and `VideoBlock` content directly, unlocking efficient multi-modal tool capabilities.

### **Changes Made:**

1. Expanded Type System
- Added `ToolReturnValue` union type: `JSONValue | DocumentBlock | ImageBlock | VideoBlock`
- Updated `tool()` helper constraints to accept media blocks while preserving type inference
- Extended `ToolResultContentData` to support direct media block types

2. Enhanced Tool Processing
- Updated `FunctionTool._wrapInToolResult()` to handle media blocks natively
- Added proper Bedrock API formatting for `DocumentBlock`, `ImageBlock`, `VideoBlock` in tool results
- Maintained backward compatibility with existing string/JSON returns

3. Improved Message Processing
- Enhanced `contentBlockFromData()` to process media blocks in tool results
- Added proper type detection for document/image/video content

### **Usage:**

```typescript
const docTool = tool({
  name: 'create_document',
  description: 'Creates a document',
  inputSchema: z.object({ content: z.string() }),
  callback: (input) => {
    return new DocumentBlock({  // ✅ Now supported!
      name: 'RESULT',
      format: 'html',
      source: { bytes: new TextEncoder().encode(input.content) }
    })
  }
})
```

### **Benefits:**
- **Efficient**: Media content flows directly to model API without conversion (and can use S3 sources)
- **Backward compatible**: Existing tools continue working unchanged
- **Performant**: Eliminates JSON/string conversion overhead for large media

Testing: Added test coverage for new `Tool` return types and runtime validation.

### Use Case

A user wants an authoritative answer from a question in a Knowledge Base that has indexed PDFs.  When the Agent identifies a possible source document, it needs to analyze the document to provide a complete response.  Agent will call a `load_document` tool with the document descriptor, and the *Tool* will return a DocumentBlock with the PDF contents (or an S3 source).  The multi-modal LLM can now give a complete and accurate response.

PR #576, opened 2026-02-26
feat: support documentblock, imageblock, videoblock in model providers that support it

mehtarac:
## Description

Added support for tools to return media content (images, documents, videos) across all model providers, with provider-specific handling based on API capabilities.

## Related Issues

https://github.com/strands-agents/sdk-typescript/issues/395

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/src/tools/function-tool.ts
+++ b/src/tools/function-tool.ts
@@ -359,8 +359,10 @@ export class FunctionTool extends Tool implements InvokableTool<unknown, JSONVal
       return value
     }
 
-    // Plain data objects — delegate to shared converter
+    // Plain data objects — require exactly one key to match the discriminated
+    // union shape; multi-key objects fall through to JsonBlock instead.
     try {
+      if (Object.keys(value as object).length !== 1) return undefined
       return toolResultContentFromData(value as ToolResultContentData)
     } catch {
       return undefined
--- a/src/tools/__tests__/tool.test.ts
+++ b/src/tools/__tests__/tool.test.ts
@@ -183,6 +183,30 @@ describe('FunctionTool', () => {
         })
       })
 
+      it('treats objects with extra keys beyond a content block key as JSON', async () => {
+        const tool = new FunctionTool({
+          name: 'extraKeyTool',
+          description: 'Returns object with text key plus extra keys',
+          inputSchema: { type: 'object' },
+          callback: (): { text: string; extra: string } => ({ text: 'abc', extra: '123' }),
+        })
+
+        const toolUse = { name: 'extraKeyTool', toolUseId: 'test-extra', input: {} }
+        const { result } = await collectGenerator(tool.stream(createMockContext(toolUse)))
+
+        expect(result).toEqual({
+          type: 'toolResultBlock',
+          toolUseId: 'test-extra',
+          status: 'success',
+          content: [
+            expect.objectContaining({
+              type: 'jsonBlock',
+              json: { text: 'abc', extra: '123' },
+            }),
+          ],
+        })
+      })
+
       it('passes input to callback exactly as provided to stream', async () => {
         const inputData = { name: 'test', value: 42, nested: { key: 'value' } }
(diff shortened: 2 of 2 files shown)


src/types/messages.ts
```diff
@@ -371,6 +371,12 @@ export class ToolResultBlock implements ToolResultBlockData, JSONSerializable<{
         return new TextBlock(contentItem.text)
       } else if ('json' in contentItem) {
         return new JsonBlock(contentItem)
+      } else if ('image' in contentItem) {
+        return ImageBlock.fromJSON(contentItem as { image: ImageBlockData })
```
pgrayy: For consistency, can we do `return new ImageBlock(contentItem.image)`? Same question for the other media types.

src/types/messages.ts
```diff
@@ -354,7 +354,7 @@ export class ToolResultBlock implements ToolResultBlockData, JSONSerializable<{
       toolResult: {
         toolUseId: this.toolUseId,
         status: this.status,
-        content: this.content.map((block) => block.toJSON()),
+        content: this.content.map((block) => block.toJSON() as ToolResultContentData),
```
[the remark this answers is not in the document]
pgrayy: Instead of casting, I think we should fix ToolResultContentData:
```typescript
export type ToolResultContentData =
  | TextBlockData
  | JsonBlockData
  | { image: ImageBlockData }
  | { video: VideoBlockData }
  | { document: DocumentBlockData }
```

This actually matches the same pattern for [ContentBlockData](https://github.com/strands-agents/sdk-typescript/blob/main/src/types/messages.ts#L110C13-L110C29) and so we are promoting consistency as well.

src/tools/function-tool.ts
```diff
@@ -246,8 +258,18 @@ export class FunctionTool extends Tool {
         })
       }
 
-      // Handle arrays by wrapping in object { $value: array }
+      // Handle arrays
       if (Array.isArray(value)) {
+        // Check if array contains only ToolResultContent instances
+        if (value.length > 0 && value.every((item) => this._isToolResultContent(item))) {
```
pgrayy: This condition will fail for arrays with mixed types. For example, maybe you have a TextBlock as the first item and a plain number as the second. I think this is okay though. It's easy enough for a user to follow this contract.

src/models/openai.ts
```diff
@@ -681,50 +655,51 @@ export class OpenAIModel extends Model<OpenAIModelConfig> {
...
+          if (imageParts.length > 0) {
+            logger.warn(
+              `tool_call_id=<${toolResult.toolUseId}> | moving images from tool result to separate user message for openai compatibility`
+            )
+          }
 
-            // Prepend error indicator if status is error
-            const finalContent = toolResult.status === 'error' ? `[ERROR] ${contentText}` : contentText
+          // Validate content is not empty
+          if (!textContent || textContent.trim().length === 0) {
```
pgrayy: In Python, it looks like we inject a text message to prevent errors ([src](https://github.com/strands-agents/sdk-python/blob/main/src/strands/models/openai.py#L262)).

src/models/gemini/adapters.ts
pgrayy: We should add tests.

src/types/messages.ts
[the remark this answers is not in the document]
pgrayy: Is the cast still necessary now that you updated ToolResultContentData?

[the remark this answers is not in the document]
pgrayy: Makes sense. And this pattern exists elsewhere ([example](https://github.com/mehtarac/sdk-typescript/blob/3f632512e1d5d2458e0662ed28819e7a6362bc11/src/types/messages.ts#L72)).

src/tools/function-tool.ts
pgrayy: Can you add unit tests for the updates.

Issue #69, opened 2025-10-28
[PARENT] OTEL Telemetry


Unshure:
Implement OpenTelemetry-based telemetry system for comprehensive tracing of agent operations. Capture detailed traces at agent, cycle, model invoke, and tool execution levels with standardized attributes.

## Work Required:
- Add OpenTelemetry SDK as dependency for tracing collection and export
- Implement telemetry integration in Agent class with optional trace_attributes during initialization
- Add Agent-Level tracing with attributes: gen_ai.system, gen_ai.agent.name, gen_ai.user.message, gen_ai.choice, system_prompt, gen_ai.request.model, start/end times, token usage
- Add Cycle-Level tracing in agent with attributes: agent_loop.cycle_id, user/assistant messages, end_time, choice.message, choice.tool.result
- Add Model Invoke tracing in model providers with attributes: gen_ai.operation.name, formatted prompts, model ID, start/end times, response, token usage
- Add Tool-Level tracing with attributes: tool.status, gen_ai.tool.name, gen_ai.tool.call.id, gen_ai.operation.name, start/end times, formatted results
- Use OpenTelemetry's built-in configuration mechanism for trace export and sampling
- Integrate tracing into Agent, agent loop, model providers, and tool execution
- Create unit tests for telemetry data collection and attribute accuracy
- Create integration test that validates complete trace spans across agent execution

## Exit Criteria:
A working OpenTelemetry-based telemetry system that captures comprehensive traces at all execution levels with proper attributes, integrates seamlessly with existing components, and exports traces using OpenTelemetry standards.

## Dependencies:
- task-09-implement-core-agent-class

Issue #492, opened 2026-02-04
[V1] Telemetry - [#8] Follow ups


lizradway:
## Fast follows to OTEL implementation
More items will be added as encountered, this is a living incomplete list:
- https://github.com/strands-agents/sdk-typescript/issues/447
- Implement exporter config union types to allow users to pass optional configurations to exporters (see https://github.com/lizradway/sdk-typescript/pull/1/changes#r2764605615 for context)
- Allow custom propagators/context managers that can override [default NodeTracerProvider Registration](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-node/src/NodeTracerProvider.ts#L96) (ie. in config.ts 
```
  _provider.register({
    contextManager: config.contextManager,
    propagator: config.propagator,
  })
```
- ~~Consider moving mcp instrumentation into mcp directory from tools directory~~
- ~~Build MCP instrumentation logic directly into callTool rather than wrap with an anonymous function~~
- ~~Add `getTracer` API that wraps global trace_api~~
- Explore accumulated message session persistence

### Documentation

https://strandsagents.com/latest/documentation/docs/user-guide/observability-evaluation/traces/#local-development-setup

### Reference

https://github.com/strands-agents/sdk-python/tree/main/src/strands/telemetry

PR #579, opened 2026-02-26
feat: add rebased telemetry implementation

lizradway:
## Description
- Add approved telemetry implementation from [`/liz/otel` branch](https://github.com/strands-agents/sdk-typescript/tree/liz/otel) rebased with structured output changes
- The only rebasing with real ramifications was done under `agent.ts`

#### Structural: _executeAgentLoop inlined into _stream
- The original branch extracted the agent loop into a separate `_executeAgentLoop` method, with `_stream` handling the agent span lifecycle around it.
- After rebase, structured output's `context.registerTool()` / `context.cleanup()` must wrap the loop, and the agent span must wrap both.
- Keeping them as separate methods would require threading the structured output context through or modifying scoping.
- Inlining the loop into `_stream` keeps a single try/finally that handles both telemetry span cleanup and structured output cleanup.
- Structured output logic and telemetry logic should be extracted out in a follow up as we consider modularization of the agent loop similarly to Python

Similarly:
- The original branch returned result from `_executeAgentLoop` and passed it to `endAgentSpan` in `_stream`'s `finally`.
- With the loop inlined, result is declared as `let result: AgentResult | undefined` at the top of `_stream` so the `finally` block can reference it for the agent span's response/stopReason attributes.

#### Force tool continue
- Structured output introduces a continue path (when forcing the model to use the structured output tool see line 473).
- The inner try/catch around each cycle only catches errors, so continue would skip `endAgentLoopSpan`. Added an explicit `endAgentLoopSpan(cycleSpan)` before the continue to prevent span leaks.
- Added test coverage for this interaction and other structured output interactions

#### _appendMessage is synchronous/ returns event for yielding
- The original /liz/otel branch had `async _appendMessage` that invoked hooks directly. 
- Main's event wrapper PR changed this to a synchronous method that returns a `MessageAddedEvent` for yielding — hooks are invoked centrally in `stream()`. The rebase uses the current synchronous pattern.

## Related Issues

https://github.com/strands-agents/sdk-typescript/issues/69

## Documentation PR

https://github.com/strands-agents/docs/pull/557

## Type of Change

New feature

## Testing

- [X] I ran `npm run check`

## Checklist

----


src/telemetry/config.ts
pgrayy: To simplify things, I would actually now recommend this module be left out. It is the responsibility of the application to import and config `@opentelemetry/sdk*`. Here we try to provide a convenience around the config and I'm not sure we have enough feedback to say that it is actually useful. My other concerns include:
1. It only supports configuring of a NodeTracerProvider and thus excludes browser users.
2. There aren't all that many configurations.

I think customers will be just fine configuring on their own for now. With that said, I would still keep this stashed away because it does demonstrate a nice, simple API that we could expose in the future. And at that time, we should also consider browser support as well.

src/telemetry/tracer.ts
```diff
@@ -0,0 +1,677 @@
...
+    for (const [key, value] of Object.entries(attributes)) {
+      if (value !== undefined && value !== null) {
+        try {
+          span.setAttribute(key, value)
+        } catch (err) {
+          logger.warn(`error=<${err}>, key=<${key}> | failed to set span attribute`)
+        }
+      }
+    }
+  }
```
poshinchen: We can remove this and use `setAttributes` from the span API.

test/integ/telemetry.test.ts
```diff
@@ -0,0 +1,702 @@
...
+  inputSchema: z.object({}),
+  callback: () => {
+    throw new Error('tool exploded')
+  },
+})
+
+describe.sequential('Telemetry Integration', () => {
+  beforeAll(() => {
+    exporter = new InMemorySpanExporter()
+    provider = new NodeTracerProvider()
```
pgrayy: Is there a provider we can use that works in both browser and node? Either way, we could follow up. Still want to check though, does this test file need to be named `telemetry.test.node.ts`?

poshinchen:
https://github.com/strands-agents/sdk-python/blob/c50457d1396dbeb8268f47d56faa5ad539f98ae7/src/strands/telemetry/tracer.py#L114

Follow-up: Could you add `is_langfuse` and [observation_type](https://github.com/strands-agents/sdk-python/blob/c50457d1396dbeb8268f47d56faa5ad539f98ae7/src/strands/telemetry/tracer.py#L688-L689) so that Langfuse won't aggregate the tokens incorrectly? We might need to add [these lines](https://github.com/strands-agents/sdk-python/blob/c50457d1396dbeb8268f47d56faa5ad539f98ae7/src/strands/telemetry/tracer.py#L252-L253) too so Langfuse can read the latest semantic conventions well.

src/tools/mcp-instrumentation.ts
```diff
@@ -0,0 +1,81 @@
...
+  if (_instrumentedClients.has(mcpClient)) {
+    return
+  }
+  _instrumentedClients.add(mcpClient)
+
+  // Store original callTool method
+  const originalCallTool = mcpClient.callTool.bind(mcpClient)
+
+  // Patch callTool to inject tracing context
+  mcpClient.callTool = async function (tool: McpTool, args: JSONValue): Promise<JSONValue> {
```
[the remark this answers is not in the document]
lizradway: Resolving so that I can merge, but adding into follow ups tracking issue: https://github.com/strands-agents/sdk-typescript/issues/492

PR #594, opened 2026-03-03
feat(multiagent): add graph multi-agent orchestration pattern

jsamuel1:
## Summary

Add a `Graph` class that provides deterministic, dependency-based multi-agent orchestration using the existing `Node`, `Edge`, and `Queue` primitives. This complements the `Swarm` pattern — Graph is deterministic/structured, Swarm is autonomous/self-organizing.

## Features

- **Parallel execution** of independent nodes via shared Queue
- **Conditional edges** with handler functions
- **Auto-detection of entry points** (nodes with no incoming edges)
- **Execution limits**: maxNodeExecutions, executionTimeout
- **Reset-on-revisit** for stateless cycle behavior in graphs with cycles
- **Full streaming support** with handoff events between batches
- **Fan-out/fan-in** patterns with dependency output propagation

## Design

Ports the `Graph` concept from the Python SDK while using the existing TypeScript primitives (`Node`, `AgentNode`, `Edge`, `Queue`, `MultiAgentState`) rather than creating parallel abstractions.

## Testing

19 unit tests covering:
- Constructor validation (duplicates, missing nodes, auto-detection)
- Linear chain execution
- Parallel/fan-out/fan-in execution
- Conditional edges
- Execution limits
- Streaming events
- Error handling

PR #618, opened 2026-03-09
feat: add AbortSignal support to Agent and BedrockModel

jsamuel1:
## Summary

Plumb `AbortSignal` through the invocation stack so callers can cancel in-flight requests:

- `InvokeOptions.signal` passed to `Agent.invokeModel`
- `StreamOptions.signal` passed to `Model.stream`
- `BedrockModel` forwards signal as `abortSignal` to AWS SDK client

## Usage

```typescript
const controller = new AbortController()
const result = await agent.invoke('Hello', { signal: controller.signal })

// Cancel from another context:
controller.abort()
```

## Follow-up needed after multi-agent PRs merge

Once the swarm (#587) and graph (#594) multi-agent PRs land, abort signal propagation should be added to both:

- **Swarm**: Currently accepts `signal?: AbortSignal` in `SwarmInvokeOptions` and checks `signal.aborted` at the top of the loop, but does NOT forward the signal to `node.agent.stream(input)`. It should pass `{ signal: options.signal }` to each agent invocation so in-flight model calls are actually cancelled.
- **Graph**: Has no abort signal support at all. Should accept a `signal` option, check it between node executions, and forward it to each agent invocation.


jsamuel1:
Closing — superseded by #781 (feat: add mid-execution cancellation) which was merged into main on April 7th and covers the same AbortSignal/cancellation ground more comprehensively.

PR #635, opened 2026-03-10
feat(bedrock): add guardLatestUserMessage guardrail option

Unshure:
## Motivation

When using Bedrock guardrails in multi-turn conversations, evaluating the entire conversation history with each request can be inefficient and costly. The Python SDK provides a `guardrail_latest_message` option that allows evaluating only the latest user message, improving performance and reducing token costs.

This PR adds the equivalent `guardLatestUserMessage` option to the TypeScript SDK's `BedrockGuardrailConfig`, bringing feature parity with the Python SDK implementation.

Resolves: #565

## Public API Changes

The `BedrockGuardrailConfig` interface now accepts an optional `guardLatestUserMessage` boolean:

```typescript
// Before: evaluate all messages with guardrails
const model = new BedrockModel({
  guardrailConfig: {
    guardrailIdentifier: 'my-guardrail-id',
    guardrailVersion: '1',
  },
})

// After: evaluate only the latest user message
const model = new BedrockModel({
  guardrailConfig: {
    guardrailIdentifier: 'my-guardrail-id',
    guardrailVersion: '1',
    guardLatestUserMessage: true,
  },
})
```

When enabled, the latest user message's text and image content are wrapped in `guardContent` blocks, signaling to Bedrock's guardrails to evaluate only that content.

The implementation finds the last user message containing text or image content (not just the last message in the array), ensuring correct behavior during tool execution cycles where `toolResult` messages (with `role='user'`) may follow the actual user input.

## Use Cases

- **Performance optimization**: Reduce latency in multi-turn conversations by avoiding re-evaluation of previously validated messages
- **Cost reduction**: Fewer tokens evaluated per request lowers guardrail evaluation costs
- **Avoiding redundant validation**: Messages that have already passed guardrail checks don't need to be re-evaluated

## Documentation
TODO

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5715,9 +5715,9 @@
       }
     },
     "node_modules/hono": {
-      "version": "4.12.7",
-      "resolved": "https://registry.npmjs.org/hono/-/hono-4.12.7.tgz",
-      "integrity": "sha512-jq9l1DM0zVIvsm3lv9Nw9nlJnMNPOcAtsbsgiUhWcFzPE99Gvo6yRTlszSLLYacMeQ6quHD6hMfId8crVHvexw==",
+      "version": "4.12.5",
+      "resolved": "https://registry.npmjs.org/hono/-/hono-4.12.5.tgz",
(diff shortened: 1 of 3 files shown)


Unshure:
/strands can you add an integ test following the tests defined here: https://github.com/strands-agents/sdk-python/blob/main/tests_integ/test_bedrock_guardrails.py#L297

PR #664, opened 2026-03-16
feat: strongly type the conversation-manager

zastrowm:
## Description

Previously, `ConversationManager` was just a `Plugin` — there  was no guidance for implementors. 

This PR introduces `ConversationManager` as a proper abstract base class with a single required method: `reduce(options: ReduceOptions): boolean`. The base class wires up overflow recovery automatically — when a `ContextWindowOverflowError` occurs, `reduce` is called and the model call is retried if reduction succeeded. If `reduce` returns `false`, the error propagates out of the agent loop uncaught, which is the correct behavior and is now clearly documented.

**Why just reduce?**: In deciding what the interface of ConversationManager should be, Patrick and I debated what the point of the ConversationManager was.  The bare minimum of a CM is that it handles the `ContextWindowOverflowError` - everything beyond that is more implementation defined.   Of the 3 implementations in Python (SlidingWindow, Summarizer, Null), only the sliding window is proactive and thus needs additional plumbing. Rather than trying to develop plumbing that works for all cases, let each subclass handle that by overriding initAgent.  We didn't want to limit what a CM could do, but we wanted to guide implementors to what they *need* to do - with everything else (proactive management) being the *extra*.

`SlidingWindowConversationManager` and `NullConversationManager` both extend the new base class. The sliding window's proactive management (trimming after each invocation) stays in its own `AfterInvocationEvent` hook via `initAgent` override, keeping `reduce` focused purely on overflow recovery.

### Public API Changes

`ConversationManager` is now an exported abstract class. `AgentConfig.conversationManager` is typed as `ConversationManager` instead of `Plugin`.

```typescript
class MyManager implements Plugin {
  readonly name = 'my:manager'
  initAgent(agent: AgentData): void {
    agent.addHook(AfterModelCallEvent, (event) => {
      if (event.error instanceof ContextWindowOverflowError) {
        // manually wire retry logic...
      }
    })
  }
}

// After: extend ConversationManager, implement reduce
class MyManager extends ConversationManager {
  readonly name = 'my:manager'

  reduce({ agent, error }: ReduceOptions): boolean {
    // called automatically on ContextWindowOverflowError
    // must return true if reduction succeeded, false to let the error propagate
    if (agent.messages.length <= 5) return false
    agent.messages.splice(0, messages.length - 5)
    return true
  }
}
```

Subclasses that also want proactive management override `initAgent`, call `super.initAgent(agent)` to preserve overflow recovery, then register additional hooks.

## Related Issues

N/A

## Documentation PR

https://github.com/strands-agents/docs/pull/664

## Type of Change

New feature
## Testing

- [X] I ran `npm run check`

## Checklist
- [X] I have updated the documentation accordingly: https://github.com/strands-agents/docs/pull/664

----

### DIFF
--- a/src/conversation-manager/sliding-window-conversation-manager.ts
+++ b/src/conversation-manager/sliding-window-conversation-manager.ts
@@ -5,11 +5,11 @@
  * that preserves tool usage pairs and avoids invalid window states.
  */
 
-import { ContextWindowOverflowError } from '../errors.js'
 import { Message, TextBlock, ToolResultBlock } from '../types/messages.js'
 import type { AgentData } from '../types/agent.js'
 import { AfterInvocationEvent } from '../hooks/events.js'
-import { ConversationManager, type ReduceOptions } from './conversation-manager.js'
+import { ConversationManager, type ConversationManagerReduceOptions } from './conversation-manager.js'
+import { logger } from '../logging/logger.js'
 
 /**
  * Configuration for the sliding window conversation manager.
@@ -85,7 +85,7 @@ export class SlidingWindowConversationManager extends ConversationManager {
    * @param options - The reduction options
    * @returns `true` if the history was reduced, `false` otherwise
    */
-  reduce({ agent, error }: ReduceOptions): boolean {
+  reduce({ agent, error }: ConversationManagerReduceOptions): boolean {
     return this._reduceContext(agent.messages, error)
   }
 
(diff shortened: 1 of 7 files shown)


src/conversation-manager/sliding-window-conversation-manager.ts
```diff
@@ -173,6 +176,7 @@ export class SlidingWindowConversationManager implements Plugin {
 
     // Overwrite message history
     messages.splice(0, trimIndex)
+    return true
```
pgrayy: Any chance this could end up splicing nothing? Or is _reduceContext always guaranteed to reduce?

PR #670, opened 2026-03-17
refactor: split agent interfaces into InvokableAgent and LocalAgent, rename MultiAgentBase to MultiAgent

pgrayy:
## Motivation

The existing `AgentBase` and `AgentData` interfaces conflate two orthogonal capabilities: invokability (invoke/stream) and local state access (messages, tools, hooks, app state). This creates problems today and blocks future work:

- `A2AAgent` is forced to construct fake `AgentData` objects (empty AppState, empty ToolRegistry, no-op addHook) just to emit `AgentResultEvent`, even though its state lives remotely.
- `AgentStreamEvent` includes `A2AStreamUpdateEvent`, but `Agent` never yields it, making the discriminated union inaccurate.
- `MultiAgentBase` uses a `type: 'multiAgent'` discriminator that does not describe what Graph and Swarm actually do.
- The future `BidiAgent` would be local but not invokable, fitting neither interface cleanly.
- `NodeStreamUpdateEvent.event` had type `AgentStreamEvent | A2AStreamEvent | ...`, which transitively required `@a2a-js/sdk` types. Since `InvokableAgent` is referenced by multiagent exports, this made `@a2a-js/sdk` effectively required for anyone using multi-agent orchestration.

Splitting along the invokable/local axis gives each agent type a precise contract and removes the need for faking fields. Decoupling `InvokableAgent.stream()` from A2A types removes the transitive dependency.

## Documentation PR

Not required. All changed interfaces (`InvokableAgent`, `LocalAgent`, `MultiAgent`) are internal and not exported for external implementation. As an agent SDK, we own the agent types that users interact with. The concrete classes (`Agent`, `A2AAgent`, `Graph`, `Swarm`) are the public API, and we should be the ones implementing new agent types rather than exposing interfaces for external implementation. These internal interfaces exist to share structure across our own agent types and to give hook/event consumers typed access to agent capabilities.

## Public API Changes

These interfaces are internal and not exported for external implementation.

### Interface renames

| Before | After | Rationale |
|--------|-------|-----------|
| `AgentBase` | `InvokableAgent` | "Base" implies it is the root of all agents, but `BidiAgent` (future) will not implement it. `InvokableAgent` communicates the actual capability: agents you can call `invoke()` and `stream()` on. |
| `AgentData` | `LocalAgent` | "Data" implies a passive data container, but this interface represents agents with locally accessible state, messages, tools, and hooks. `LocalAgent` distinguishes from `A2AAgent`, where state lives on a remote server. |
| `MultiAgentBase` | `MultiAgent` | "Base" is an implementation detail. The codebase already uses the `MultiAgent` prefix extensively (`MultiAgentState`, `MultiAgentStreamEvent`, `MultiAgentPlugin`, etc.), so `MultiAgent` is consistent with existing naming and describes what Graph and Swarm are. |

`InvokableAgent` carries `id`, `name?`, `description?`, `invoke()`, and `stream()`. `LocalAgent` carries `state`, `messages`, `toolRegistry`, and `addHook()`. `Agent` implements both. `A2AAgent` implements only `InvokableAgent`.

### InvokableAgent.stream() decoupled from A2A types

`InvokableAgent.stream()` now yields `StreamEvent` (empty abstract base class) instead of `AgentStreamEvent`. This removes the transitive `@a2a-js/sdk` dependency from multiagent consumers.

### NodeStreamUpdateEvent.inner (replaces .event)

`NodeStreamUpdateEvent.event` is renamed to `NodeStreamUpdateEvent.inner` with a new `NodeStreamUpdateInnerEvent` type. This is a wrapper object discriminated by `source`:

```typescript
type NodeStreamUpdateInnerEvent =
  | { readonly source: 'agent'; readonly event: AgentStreamEvent }
  | { readonly source: 'multiAgent'; readonly event: Exclude<MultiAgentStreamEvent, NodeStreamUpdateEvent> }
  | { readonly source: 'custom'; readonly event: StreamEvent }
```

Consumer pattern:

```typescript
for await (const event of graph.stream('hello')) {
  if (event.type === 'nodeStreamUpdateEvent') {
    const { source, event: innerEvent } = event.inner
    if (source === 'agent') {
      // innerEvent is AgentStreamEvent, narrows on .type
      if (innerEvent.type === 'modelStreamUpdateEvent') { ... }
    }
  }
}
```

### A2AExecutor narrowing

`A2AExecutor` receives `StreamEvent` from `InvokableAgent.stream()` and uses `instanceof` checks (`ModelStreamUpdateEvent`, `ContentBlockEvent`) to narrow events. This is an internal implementation detail, not consumer-facing.

### Type discriminators removed

`InvokableAgent` and `MultiAgent` no longer carry a `type` discriminator field. `Graph._resolveNodes` now uses `instanceof` checks for `Graph` and `Swarm` instead of checking `definition.type`.

### A2A event separation

`AgentStreamEvent` no longer includes `A2AStreamUpdateEvent`. A new `A2AResultEvent` replaces `AgentResultEvent` for `A2AAgent`, carrying only `{ result }` instead of requiring a `LocalAgent` reference. `A2AAgent` no longer needs to fake local state fields.

### File renames

`src/agent/agent-base.ts` deleted (interfaces moved to `src/types/agent.ts`). `src/multiagent/base.ts` renamed to `src/multiagent/multiagent.ts`.

## Breaking Changes

- `AgentBase` renamed to `InvokableAgent`
- `AgentData` renamed to `LocalAgent`
- `MultiAgentBase` renamed to `MultiAgent`
- `readonly type` field removed from `InvokableAgent` (was `'agent'`) and `MultiAgent` (was `'orchestrator'`/`'multiAgent'`)
- `AgentStreamEvent` no longer includes `A2AStreamUpdateEvent`
- `AgentBase` and `AgentData` no longer exported from the package entry point
- `InvokableAgent.stream()` yield type changed from `AgentStreamEvent` to `StreamEvent`
- `NodeStreamUpdateEvent.event` renamed to `NodeStreamUpdateEvent.inner` (type changed to `NodeStreamUpdateInnerEvent`)

Since the SDK is in preview and these interfaces are internal, the migration path is to update any direct references to the new names.

## Testing

- [x] I ran `npm run check`

## Checklist

### DIFF
--- a/src/a2a/events.ts
+++ b/src/a2a/events.ts
@@ -4,7 +4,7 @@
 
 import type { Message, Task, TaskStatusUpdateEvent, TaskArtifactUpdateEvent } from '@a2a-js/sdk'
 import { StreamEvent } from '../hooks/events.js'
-import type { AgentResult } from '../types/agent.js'
+import type { AgentResultEvent } from '../hooks/events.js'
 
 /**
  * Union of raw A2A protocol event types received during streaming.
@@ -37,9 +37,9 @@ export class A2AStreamUpdateEvent extends StreamEvent {
  */
 export class A2AResultEvent extends StreamEvent {
   readonly type = 'a2aResultEvent' as const
-  readonly result: AgentResult
+  readonly result: AgentResultEvent['result']
 
-  constructor(data: { result: AgentResult }) {
+  constructor(data: Pick<AgentResultEvent, 'result'>) {
     super()
     this.result = data.result
   }


AGENT-INTERFACE-PROPOSAL.md
```diff
@@ -0,0 +1,113 @@
...
+- `A2AAgent` implements `InvokableAgent` only
+- `BidiAgent` implements `LocalAgent` only
+
+### Orchestrator Interface
+
+Rename `MultiAgentBase` to `Orchestrator` and update the type discriminator:
+
+| Before | After |
+|--------|-------|
+| `MultiAgentBase` | `Orchestrator` |
```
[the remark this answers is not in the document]
pgrayy: Removed.

src/agent/agent.ts
```diff
@@ -150,26 +155,6 @@ export type AgentConfig = {
...
-
-/**
- * Options for a single agent invocation.
- */
-export interface InvokeOptions {
-  /**
-   * Zod schema for structured output validation, overriding the constructor-provided schema for this invocation only.
-   */
-  structuredOutputSchema?: z.ZodSchema
-}
```
pgrayy: Copied as is into `src/types/agent.ts` for organizational purposes.

src/agent/agent-base.ts
pgrayy: Renamed to InvokableAgent and moved to `src/types/agent.ts`.

PR #678, opened 2026-03-17
fix: swarm maxstep throws when finish normally

JackYPCOnline:
## Description

Fixes a bug where Swarm throws "swarm reached step limit" even when the swarm completes normally on its final allowed step.

Problem
The _checkSteps guard ran unconditionally after the main loop in _stream:
```
while (state.steps < this.config.maxSteps) {
  state.steps++
  const result = yield* this._streamNode(...)
  if (result.status === Status.FAILED || !handoff?.agentId) {
    break  // normal termination
  }
  // hand off...
}

this._checkSteps(state)  // throws if steps >= maxSteps

```
If a swarm with maxSteps=N runs exactly N steps and terminates normally (the last agent produces a final response with no handoff), state.steps equals maxSteps. The break exits the loop, then _checkSteps throws — even though the swarm completed successfully.

## Type of Change

Bug fix
New feature
Breaking change
Documentation update
Other (please describe):

## Testing

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/src/multiagent/swarm.ts
+++ b/src/multiagent/swarm.ts
@@ -213,7 +213,7 @@ export class Swarm implements MultiAgent {
         node = target
       }
 
-      this._checkSteps(state)
+      this._checkSteps(state, handoff)
     } finally {
       yield new AfterMultiAgentInvocationEvent({ orchestrator: this, state })
     }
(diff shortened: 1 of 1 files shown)


src/multiagent/swarm.ts
```diff
@@ -216,7 +216,11 @@ export class Swarm implements MultiAgent {
         node = target
       }
 
-      this._checkSteps(state)
+      // If the loop exited because steps were exhausted (not via break),
+      // the last agent still wanted to hand off — that's an error.
+      if (handoff?.agentId && state.steps >= this.config.maxSteps) {
```
pgrayy: Can you pass handoff into _checkSteps and perform the check there. Then can you document the behavior on the docstring.

PR #687, opened 2026-03-18
fix: fix agent retry pass in same arg

JackYPCOnline:
## Description

Fix user message duplication when invokeModel retries via AfterModelCallEvent.retry.

When a retry is triggered (e.g., by ConversationManager on context overflow), invokeModel recursed with the original args, causing _normalizeInput(args) to re-append the user's message to this.messages. After a single retry, the conversation history contained the same user message twice.

Both retry paths were affected:

Success path: hook inspects the response and requests retry

Error path: ConversationManager reduces context on ContextWindowOverflowError and requests retry

The fix passes undefined instead of args on retry, since the messages were already appended in the first call. forcedToolChoice is also forwarded to preserve structured output forcing behavior.

## Type of Change

Bug fix
New feature
Breaking change
Documentation update
Other (please describe):

## Testing

Added two unit tests in src/agent/__tests__/agent.hook.test.ts:

"does not duplicate user messages on error retry" — triggers retry via AfterModelCallEvent error path, asserts the user message appears exactly once in agent.messages

"does not duplicate user messages on success retry" — triggers retry via AfterModelCallEvent success path, asserts the same

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/src/agent/agent.ts
+++ b/src/agent/agent.ts
@@ -495,8 +495,16 @@ export class Agent implements LocalAgent, InvokableAgent {
         })
 
         try {
-          const modelResult = yield* this.invokeModel(currentArgs, forcedToolChoice)
-          currentArgs = undefined // Only pass args on first invocation
+          // Normalize input and append user messages on first invocation only
+          if (currentArgs !== undefined) {
+            const messagesToAppend = this._normalizeInput(currentArgs)
+            for (const message of messagesToAppend) {
+              yield this._appendMessage(message)
+            }
+            currentArgs = undefined
+          }
+
+          const modelResult = yield* this.invokeModel(forcedToolChoice)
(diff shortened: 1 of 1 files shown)


src/agent/agent.ts
```diff
@@ -723,7 +723,7 @@ export class Agent implements LocalAgent, InvokableAgent {
       yield afterModelCallEvent
 
       if (afterModelCallEvent.retry) {
-        return yield* this.invokeModel(args)
+        return yield* this.invokeModel(undefined, forcedToolChoice)
```
zastrowm: This is a bit confusing to read; I think we should just remove the Messages argument from invokeModel and do it at the one call site that needs it.  Then calls to `invokeModel(forcedToolChoice)` reads more naturally and we don't have to keep track of when we pass a message or not

Issue #675, opened 2026-03-17
[FEATURE] TS MCP tool multi modal support


mehtarac:
### Problem Statement

support multi modal tool results

### Use Case

-

PR #703, opened 2026-03-20
feat: add multi modal tool result support for mcp tool

mehtarac:
## Description

MCP tools can return multi-modal content — images, embedded resources, and mixed content — but McpTool only handled text content blocks,falling back to JsonBlock for everything else. This meant image data from MCP servers was passed to the model as raw JSON (base64 strings) instead of proper ImageBlock instances, preventing models from actually interpreting the visual content.

## Related Issues
https://github.com/strands-agents/sdk-typescript/issues/675

## Type of Change

Bug fix
New feature
Breaking change
Documentation update
Other (please describe):

## Testing

- [ ] I ran `npm run check`

## Checklist
- [ ] I have read the CONTRIBUTING document
- [ ] My changes generate no new warnings

----

### DIFF
--- a/src/__tests__/mcp.test.ts
+++ b/src/__tests__/mcp.test.ts
@@ -4,6 +4,7 @@ import type { Transport } from '@modelcontextprotocol/sdk/shared/transport.js'
 import { McpClient } from '../mcp.js'
 import { McpTool } from '../tools/mcp-tool.js'
 import { JsonBlock, type TextBlock, type ToolResultBlock } from '../types/messages.js'
+import { ImageBlock } from '../types/media.js'
 import type { LocalAgent } from '../types/agent.js'
 import type { ToolContext } from '../tools/tool.js'
 import { context, propagation, trace, TraceFlags } from '@opentelemetry/api'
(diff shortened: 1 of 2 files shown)


JiwaniZakir (commented):
The duplicate imports from the same module in `mcp-tool.ts` should be consolidated — `ImageBlock` and `decodeBase64` are each imported on separate lines from `'../types/media.js'`, and similarly `toMediaFormat` and `ImageFormat` are split across two lines from `'../mime.js'`. These should each be merged into single import statements.

Additionally, `_mapMcpContent` declares a return type of `ToolResultContent | undefined`, and the caller filters with `if (block)`, but based on all visible code paths in the switch statement (including the `default` branch returning a `JsonBlock`), the function never appears to actually return `undefined`. If that's truly the case, the `undefined` return type and the `if (block)` guard in the `for...of` loop are dead code and should be removed for clarity — or if there's a legitimate path returning `undefined` in the truncated portion of `_mapMcpImageContent`/`_mapMcpEmbeddedResource`, a test covering that case is missing.

The test `'handles mixed content types in a single result'` includes a resource item with no `mimeType` (`{ uri: 'file:///notes.txt', text: 'Some notes' }`), which is a useful edge case, but there's no corresponding test for a resource that has neither `text` nor `blob` — verifying that fallback path for `_mapMcpEmbeddedResource` would strengthen coverage.

mehtarac:
closing this PR and re-opened [here](https://github.com/strands-agents/sdk-typescript/pull/865) since this was a little stale

PR #708, opened 2026-03-20
feat: add toJSON() to all streaming events for wire-safe serialization

agent-of-mkmeral:
## Motivation

Every event from `agent.stream()` carries `agent: LocalAgent` — the full Agent instance with conversation history, app state, model config, and internal registries. When users send events over the wire via `JSON.stringify(event)`, the output is massively bloated:

| Conversation size | Per-event JSON | With toJSON() |
|---|---|---|
| 100 messages | ~54 KB | ~100-200 bytes |
| 1,000 messages | ~246 KB | ~100-200 bytes |

This makes SSE streaming, WebSocket forwarding, and any wire transport impractical without manual field picking.

Resolves: agent-of-mkmeral/strands-coder#41

## Public API Changes

All 16 streaming event classes now have a `toJSON()` method that `JSON.stringify()` calls automatically. Zero breaking changes — `event.agent` still works for direct access in hooks and in-process code.

```typescript
// Same code, dramatically different wire output
for await (const event of agent.stream('Hello')) {
  // Before: ~54KB per text delta (entire agent serialized)
  // After: ~120 bytes (just the delta)
  res.write(`data: ${JSON.stringify(event)}\n\n`)
}
```

What each event serializes to:

| Event | `toJSON()` output |
|---|---|
| `ModelStreamUpdateEvent` | `{ type, event }` |
| `ContentBlockEvent` | `{ type, contentBlock }` |
| `ModelMessageEvent` | `{ type, message, stopReason }` |
| `ToolResultEvent` | `{ type, result }` |
| `ToolStreamUpdateEvent` | `{ type, event }` |
| `AgentResultEvent` | `{ type, result }` |
| `MessageAddedEvent` | `{ type, message }` |
| `BeforeToolCallEvent` | `{ type, toolUse }` |
| `AfterToolCallEvent` | `{ type, toolUse, result, error? }` |
| `AfterModelCallEvent` | `{ type, stopData?, error? }` |
| `BeforeToolsEvent` / `AfterToolsEvent` | `{ type, message }` |
| Lifecycle events | `{ type }` |

Excluded from all events: `agent: LocalAgent`. Additionally excluded where applicable: `tool: Tool` (class with methods), mutable flags (`cancel`, `retry`). `Error` objects are converted to their `.message` string.

cc @mkmeral

### DIFF
--- a/src/hooks/__tests__/events.test.ts
+++ b/src/hooks/__tests__/events.test.ts
@@ -951,7 +951,10 @@ describe('toJSON serialization completeness', () => {
       { name: 'BeforeInvocationEvent', event: new BeforeInvocationEvent({ agent }) },
       { name: 'AfterInvocationEvent', event: new AfterInvocationEvent({ agent }) },
       { name: 'BeforeModelCallEvent', event: new BeforeModelCallEvent({ agent }) },
-      { name: 'AfterModelCallEvent', event: Object.assign(new AfterModelCallEvent({ agent, stopData, error }), { retry: true }) },
+      {
+        name: 'AfterModelCallEvent',
+        event: Object.assign(new AfterModelCallEvent({ agent, stopData, error }), { retry: true }),
+      },
       { name: 'MessageAddedEvent', event: new MessageAddedEvent({ agent, message }) },
       { name: 'ModelStreamUpdateEvent', event: new ModelStreamUpdateEvent({ agent, event: streamEvent }) },
       { name: 'ContentBlockEvent', event: new ContentBlockEvent({ agent, contentBlock }) },
@@ -960,7 +963,10 @@ describe('toJSON serialization completeness', () => {
       { name: 'ToolStreamUpdateEvent', event: new ToolStreamUpdateEvent({ agent, event: toolStreamEvent }) },
       { name: 'AgentResultEvent', event: new AgentResultEvent({ agent, result: agentResult }) },
       { name: 'BeforeToolCallEvent', event: new BeforeToolCallEvent({ agent, toolUse, tool }) },
-      { name: 'AfterToolCallEvent', event: Object.assign(new AfterToolCallEvent({ agent, toolUse, tool, result, error }), { retry: true }) },
+      {
+        name: 'AfterToolCallEvent',
+        event: Object.assign(new AfterToolCallEvent({ agent, toolUse, tool, result, error }), { retry: true }),
+      },
       { name: 'BeforeToolsEvent', event: new BeforeToolsEvent({ agent, message }) },
       { name: 'AfterToolsEvent', event: new AfterToolsEvent({ agent, message }) },
     ]


mkmeral:
@agent-of-mkmeral do we send all of these events through stream? we don't need to do it for hooks, only streamed events

agent-of-mkmeral:
Good question @mkmeral! Yes — in the current architecture, **all 16 events flow through `stream()`**. There's no distinction between "hook-only" and "stream-only" events.

Here's the relevant code from [`agent.ts` lines 407-439](https://github.com/strands-agents/sdk-typescript/blob/main/src/agent/agent.ts#L407-L439):

```typescript
public async *stream(args, options) {
  const streamGenerator = this._stream(args, options)
  let result = await streamGenerator.next()

  while (!result.done) {
    const event = result.value
    // Hooks fire on the event
    if (event instanceof HookableEvent) {
      await this._hooksRegistry.invokeCallbacks(event)
    }
    this._printer?.processEvent(event)
    yield event  // ← every event is also yielded to the stream consumer
    result = await streamGenerator.next()
  }
  // ...
}
```

So when a user does `for await (const event of agent.stream(...))` and then `JSON.stringify(event)`, **any** of the 16 event types can hit the serialization path — including lifecycle events like `BeforeInvocationEvent`, `BeforeModelCallEvent`, etc.

That said, you raise a valid design point. We could scope `toJSON()` to only the events users typically care about serializing over the wire (the "data events"):

| **Commonly serialized (data events)** | **Rarely serialized (lifecycle/hook events)** |
|---|---|
| `ModelStreamUpdateEvent` | `InitializedEvent` |
| `ContentBlockEvent` | `BeforeInvocationEvent` / `AfterInvocationEvent` |
| `ModelMessageEvent` | `BeforeModelCallEvent` |
| `ToolResultEvent` | `BeforeToolCallEvent` / `AfterToolCallEvent` |
| `ToolStreamUpdateEvent` | `BeforeToolsEvent` / `AfterToolsEvent` |
| `AgentResultEvent` | `MessageAddedEvent` |
| `AfterModelCallEvent` (has error) | |

But even lifecycle events carry `agent: LocalAgent`, so a single `JSON.stringify(event)` on any event causes the bloat. Adding `toJSON()` to all events ensures no landmine regardless of which events the user serializes.

Would you prefer I:
1. **Keep as-is** — `toJSON()` on all 16 events (defensive, consistent)
2. **Scope down** — `toJSON()` only on the ~7 data events that users typically serialize, and add the rest later if needed

Happy to adjust either way! 👍

src/hooks/events.ts
```diff
@@ -89,6 +89,14 @@ export class InitializedEvent extends HookableEvent {
     super()
     this.agent = data.agent
   }
+
+  /**
+   * Serializes for wire transport, excluding the agent reference.
+   * Called automatically by JSON.stringify().
+   */
+  toJSON(): { type: 'initializedEvent' } {
```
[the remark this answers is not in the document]
agent-of-mkmeral: Done ✅ — Updated all 16 `toJSON()` return types to use `Pick<>`:

- **Lifecycle events:** `Pick<InitializedEvent, 'type'>`, etc.
- **Data events:** `Pick<ModelStreamUpdateEvent, 'type' | 'event'>`, etc.
- **Error-transforming events:** `Pick<AfterToolCallEvent, 'type' | 'toolUse' | 'result'> & { error?: string }` — uses `Pick` for the fields that pass through unchanged, plus an intersection for the `Error → string` transformation.

All quality gates pass: 1705/1705 tests ✅, type-check ✅, ESLint ✅, Prettier ✅

src/hooks/events.ts
```diff
@@ -218,6 +258,20 @@ export class AfterToolCallEvent extends HookableEvent {
   override _shouldReverseCallbacks(): boolean {
     return true
   }
+
+  /**
+   * Serializes for wire transport, excluding the agent reference, tool instance, and mutable retry flag.
+   * Converts Error to its message string for safe serialization.
+   * Called automatically by JSON.stringify().
+   */
+  toJSON(): Pick<AfterToolCallEvent, 'type' | 'toolUse' | 'result'> & { error?: string } {
```
zastrowm: For snapshot/Messages, we currently exclude errors, until we have a better serialization story for them; Should we do that here?  

If not, what if we start with `{ error: { message?: string } }` so that later we can add more information to it (for example the `type` or something)

src/hooks/__tests__/events.test.ts
```diff
@@ -592,3 +592,317 @@ describe('AfterToolsEvent', () => {
...
+    it('includes toolUse and excludes agent, tool, and cancel', () => {
+      const agent = new Agent()
+      const tool = new FunctionTool({
+        name: 'testTool',
+        description: 'Test',
+        inputSchema: {},
+        callback: () => 'result',
+      })
+      const toolUse = { name: 'testTool', toolUseId: 'id-1', input: { query: 'hello' } }
+      const event = new BeforeToolCallEvent({ agent, toolUse, tool })
```
zastrowm: Is there a good way to determine what *is* excluded?  For example, could we have a test with every event, and just verify that the class and the serialized json match *except* for agent, tool, etc.?

I'm a bit concerned of us adding new fields and not realizing that we needed to add it to the event too

pgrayy (dismissed):

Issue #549, opened 2026-02-19
Structured Output  - Follow-up work for structured output


mehtarac:
### Summary

Address the comments marked as follow-up

- [Comment #1](https://github.com/strands-agents/sdk-typescript/pull/402#discussion_r2819234922)
- [Comment #2](https://github.com/strands-agents/sdk-typescript/pull/402#discussion_r2819248100)
- [Comment #3](https://github.com/strands-agents/sdk-typescript/pull/402#discussion_r2819264422)
- [Comment #4](https://github.com/strands-agents/sdk-typescript/pull/402#discussion_r2822785945)
- [Comment #5](https://github.com/strands-agents/sdk-typescript/pull/402#discussion_r2822847107)
- [Comment #6](https://github.com/strands-agents/sdk-typescript/pull/402#discussion_r2805278509)

### Documentation

NA

### Reference

NA

PR #709, opened 2026-03-20
refactor: simplify structured output internals and fix infinite loop bug

pgrayy:
## Motivation

The structured output implementation had three issues:

- **Infinite loop with OpenAI models**: When the model voluntarily called the `strands_structured_output` tool in the `toolUse` branch, the agent loop did not check whether structured output was captured, so it continued looping indefinitely.
- **Unnecessary extra model call**: After successful structured output capture, the old code continued the loop, sent the tool result back to the model, and waited for a text response before exiting. This wasted a round-trip and tokens for no reason. The Python SDK already exits immediately after capturing structured output, and we now match that behavior.
- **Consolidation opportunity**: The implementation was spread across four files in a separate `structured-output/` directory with a stateful context object. This consolidates it into a single `StructuredOutputTool` under `src/tools/`, making the flow easier to follow.
- **Missing message**: If model ended on "end_turn", we did not append the associated assistant message before forcing the structured output call. 

Resolves: #549

## Public API Changes

`StructuredOutputException` is renamed to `StructuredOutputError` and now lives in `src/errors.ts` alongside the other SDK error classes. All other public interfaces (`structuredOutputSchema` on `AgentConfig`, `structuredOutput` on `AgentResult`, Zod schema support) are unchanged.

```typescript
// Before
import { StructuredOutputException } from '@strands-agents/sdk/structured-output/exceptions'

// After
import { StructuredOutputError } from '@strands-agents/sdk'
```

## Breaking Changes

`StructuredOutputException` no longer exists. Code that catches it by name should switch to `StructuredOutputError`. The class is exported from the main SDK entry point.

## Documentation PR

https://github.com/strands-agents/docs/pull/690

## Related Issues

https://github.com/strands-agents/sdk-typescript/issues/549
https://github.com/JackYPCOnline/sdk-typescript/issues/16

## Type of Change

Bug fix

## Testing

- [x] I ran `npm run check`
- [x] All 1632 unit tests pass
- [x] Integration test added and passing against Bedrock and Gemini

## Checklist

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -173,7 +173,6 @@ sdk-typescript/
 - **`src/plugins/`**: Plugin system for extending agent functionality
 - **`src/models/`**: Model provider implementations (Bedrock, OpenAI, future providers)
 - **`src/tools/`**: Tool definitions, types, and structured output validation with Zod schemas
-- **`src/tools/`**: Tool definitions and types for agent tool use
 - **`src/multiagent/`**: Multi-agent orchestration patterns (Graph for DAG execution, Swarm for handoff-based routing)
 - **`src/types/`**: Core type definitions used across the SDK
 - **`src/vended-tools/`**: Optional vended tools (not part of core SDK, independently importable)


src/agent/agent.ts
```diff
@@ -504,85 +506,68 @@ export class Agent implements LocalAgent, InvokableAgent {
...
-            yield this._appendMessage(modelResult.message)
-
-            // End cycle tracking
             this._meter.endCycle(cycleStartTime)
-
-            // End cycle span
             this._tracer.endAgentLoopSpan(cycleSpan)
 
-            const structuredOutput = context.getResult()
+            yield this._appendMessage(modelResult.message)
```
pgrayy: Was missing this before for structured output. Should call that out as another fix. If the model did not get structured output in one shot, this append was missed.

PR #710, opened 2026-03-22
fix: validate agent implements stream() in AgentNode constructor

jsamuel1:
## Problem

Passing an object that doesn't implement `InvokableAgent.stream()` to `AgentNode` (or indirectly via `Graph`/`Swarm` node definitions) produces a cryptic runtime error:

```
_ is not a function
```

This happens because `_resolveNodes()` casts unrecognized objects to `InvokableAgent` without validation, and the error only surfaces later when `AgentNode.handle()` calls `this._agent.stream()`.

## Solution

Add an early check in `AgentNode` constructor that verifies the agent has a `stream()` method. The error message includes the node id for easier debugging:

```
node_id=<marge> | agent does not implement stream(). Ensure the object passed as an agent has a stream() method.
```

## Testing

- Added unit test for the validation
- All existing tests pass (1633 tests, 0 failures)

### DIFF
--- a/src/multiagent/__tests__/nodes.test.ts
+++ b/src/multiagent/__tests__/nodes.test.ts
@@ -116,6 +116,20 @@ describe('AgentNode', () => {
     state = new MultiAgentState({ nodeIds: ['agent-1'] })
   })
 
+  describe('constructor', () => {
+    it('throws when agent does not implement stream()', () => {
+      const notAnAgent = { id: 'bad' } as any
+      expect(() => new AgentNode({ agent: notAnAgent })).toThrow('agent must implement InvokableAgent')
+    })
+
+    it('throws with fallback id when agent has neither stream nor id', () => {
+      const notAnAgent = {} as any
+      expect(() => new AgentNode({ agent: notAnAgent })).toThrow(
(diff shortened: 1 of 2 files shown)


JiwaniZakir (commented):
The validation in `nodes.ts` checks `typeof agent.stream !== 'function'` before calling `super()`, which is correct, but it then references `agent.id` in the error message without first verifying that `agent.id` is defined. If a caller passes an object with neither `stream` nor `id`, the error reads `node_id=<undefined>`, which is less helpful than it could be. A simple fallback like `agent.id ?? 'unknown'` would make the message more robust.

The test in `nodes.test.ts` is placed before the `beforeEach` block, which is unconventional — it inverts the typical setup-then-test structure and could confuse readers scanning the file. Moving it after `beforeEach` (or into a dedicated `describe('constructor validation', ...)` block) would be cleaner.

The test itself only covers the case where `id` is present but `stream` is absent. Adding a case where neither property exists would exercise the `agent.id` undefined path mentioned above and ensure the error message degrades gracefully in that scenario.

zastrowm:
@jsamuel1 out of curiosity, how did you encounter this?  Are you using JS directly?

Issue #725, opened 2026-03-23
[BUG] SessionManager: saveLatestOn 'trigger' with no snapshotTrigger silently never saves


pgrayy:
### Checks

- [x] I have updated to the latest minor and patch version of Strands
- [x] I have checked the documentation and this is not expected behavior
- [x] I have searched ./issues and there are no duplicates of my issue

### Strands Version

0.7.0

### Node.js Version

22.x

### Operating System

macOS

### Installation Method

npm

### Steps to Reproduce

```typescript
import { SessionManager, FileStorage } from "@strands-agents/sdk"

const session = new SessionManager({
  sessionId: "my-session",
  storage: { snapshot: new FileStorage() },
  saveLatestOn: "trigger",
  // no snapshotTrigger provided
})

const agent = new Agent({ sessionManager: session })
await agent.invoke("Hello")
// snapshot_latest is never saved, conversation state is silently lost
```

### Expected Behavior

The `SessionManager` constructor should throw a validation error when `saveLatestOn` is set to `"trigger"` but no `snapshotTrigger` callback is provided. This is a misconfiguration that leads to silent data loss.

### Actual Behavior

No error is thrown. The session manager initializes successfully but never saves `snapshot_latest` because the trigger check in `_onAfterAgentInvocation` short-circuits:

```typescript
if (this._snapshotTrigger?.({ agentData: agent })) {
  await this._saveImmutableAndLatest(agent)
}
```

With no trigger defined, this is always falsy, so no snapshots are ever persisted.

### Additional Context

Found during bug bash by @maczas.

### Possible Solution

Add validation in the constructor:

```typescript
if (this._saveLatestOn === "trigger" && !this._snapshotTrigger) {
  throw new Error("snapshotTrigger is required when saveLatestOn is \"trigger\"")
}
```

### Related Issues

N/A

PR #730, opened 2026-03-23
fix: add SessionManager guard rails and widen snapshot types to LocalAgent

pgrayy:
## Description

This PR adds guard rails to `SessionManager` and widens the snapshot/session APIs from `Agent` to `LocalAgent`, eliminating unsafe type casts.

**Overwrite warning**: When a session restore overwrites an agent that already has messages (e.g., a user passed initial messages to the Agent constructor alongside a session manager), a warning is logged so the developer knows their messages were replaced.

**Type widening**: `takeSnapshot`, `loadSnapshot`, `saveSnapshot`, `restoreSnapshot`, and internal helpers now accept `LocalAgent` instead of `Agent`. This removes all `as Agent` casts from `session-manager.ts`. To support this, `LocalAgent` gains two fields that `Agent` already implements:
- `readonly id: string`
- `systemPrompt?: SystemPrompt`

## Related Issues

Resolves strands-agents/sdk-typescript#725

Related: strands-agents/sdk-python#2531 (duplicate agent ID protection, tracked separately for cross-process design)

## Type of Change

Bug fix

## Testing

- [x] I ran `npm run check`

## Checklist

---

### DIFF
--- a/src/session/session-manager.ts
+++ b/src/session/session-manager.ts
@@ -72,10 +72,6 @@ export class SessionManager implements Plugin {
     this._storage = { snapshot: config.storage.snapshot }
     this._saveLatestOn = config.saveLatestOn ?? 'invocation'
     this._snapshotTrigger = config.snapshotTrigger
-
-    if (this._saveLatestOn === 'trigger' && !this._snapshotTrigger) {
-      throw new Error('snapshotTrigger is required when saveLatestOn is "trigger"')
-    }
   }
 
   /** Initializes the plugin by registering lifecycle hook callbacks. */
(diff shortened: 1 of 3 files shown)


src/session/session-manager.ts
```diff
@@ -72,6 +72,10 @@ export class SessionManager implements Plugin {
     this._storage = { snapshot: config.storage.snapshot }
     this._saveLatestOn = config.saveLatestOn ?? 'invocation'
     this._snapshotTrigger = config.snapshotTrigger
+
+    if (this._saveLatestOn === 'trigger' && !this._snapshotTrigger) {
```
JackYPCOnline: I recommand not adding this validation here, for example: some users they might want to custom their behavior, like persisting on specific hookEvent, then it makes sense to silently turn off auto snapshot_latest.

pgrayy: Could you elaborate more? An example might help me better understand.

Issue #94, opened 2025-10-29
Interrupt System


Unshure:
System to allow tools to be interrupted and resumed from

PR #784, opened 2026-04-02
feat(interrupt): implement interrupt system for human-in-the-loop workflows

### LINKED ISSUE #477: [V1] Interrupts - Before Tool Call Event
### Summary

Support raising interrupts from BeforeToolCallEvent hooks. Handle the propagation of the interrupt and response payloads through the agent loop.

This enables human-in-the-loop workflows where users can pause execution before a tool runs, collect approval or additional input, and then resume with the response.

### Usage

```typescript
import { Agent, BeforeToolCallEvent } from '@strands-agents/sdk'

const approvalHook = {
  registerCallbacks(registry) {
    registry.addCallback(BeforeToolCallEvent, (event) => {
      if (event.tool.name === 'delete_file') {
        event.interrupt({
          type: 'approval_required',
          message: `Approve deletion of ${event.input.path}?`,
          payload: { tool: event.tool.name, input: event.input }
        })
      }
    })
  }
}

const agent = new Agent({ hooks: [approvalHook] })

const result = await agent.invoke('Delete the temp files')

if (result.stopReason === 'interrupt') {
  // Collect user approval...
  const resumedResult = await agent.invoke('Delete the temp files', {
    interruptResponse: { approved: true }
  })
}
```

### Documentation

https://strandsagents.com/latest/documentation/docs/user-guide/concepts/interrupts/

### Reference

https://github.com/strands-agents/sdk-python/blob/main/src/strands/interrupt.py

### LINKED ISSUE #478: Interrupts - Before Tools Event
## Summary

Support raising interrupts from `BeforeToolsEvent` hooks to pause execution before any tools in a batch are executed.

## Usage

```typescript
import { Agent, BeforeToolsEvent } from '@strands-agents/sdk'

const agent = new Agent({
  hooks: {
    beforeTools: async (event: BeforeToolsEvent) => {
      const dangerousTools = event.toolUses.filter(t =>
        ['delete_file', 'execute_command'].includes(t.name)
      )

      if (dangerousTools.length > 0) {
        event.interrupt({
          name: 'batch_approval',
          reason: `Approve ${dangerousTools.length} dangerous tool calls?`,
          response: undefined,
        })
      }
    },
  },
})

const result = await agent.invoke('Delete all temp files and run cleanup')

if (result.stopReason === 'interrupt') {
  console.log('Interrupt:', result.interrupt)
  
  // Resume with user response
  const resumed = await agent.invoke({
    interruptResponse: {
      name: 'batch_approval',
      response: { approved: true },
    },
  })
}
```

## Documentation

N/A

## Reference

N/A

### LINKED ISSUE #479: Interrupts - Tool Call
## Summary

Implement a comprehensive interrupt system for human-in-the-loop workflows in the TypeScript SDK. This allows pausing agent execution to collect user input, confirmation, or clarification from:

1. **Within tool execution** (ZodTool/FunctionTool callbacks)
2. **`BeforeToolCallEvent` hooks** (before individual tool calls)
3. **`BeforeToolsEvent` hooks** (before batch tool execution)

> **Note**: This issue supersedes #477 and #478, which will be closed as duplicates once this is implemented.

### Interrupt from Tool Execution

```typescript
import { Agent, tool } from '@strands-agents/sdk'
import { z } from 'zod'

const transferMoney = tool({
  name: 'transfer_money',
  description: 'Transfer money between accounts',
  parameters: z.object({
    from: z.string(),
    to: z.string(),
    amount: z.number(),
  }),
  callback: async (input, context) => {
    if (input.amount > 1000) {
      // First call raises interrupt, second call returns user response
      const response = context.interrupt({
        name: 'confirm_transfer',
        reason: `Confirm transfer of $${input.amount}?`,
      })
      
      if (!response.confirmed) {
        return { success: false, reason: 'Transfer cancelled by user' }
      }
    }

    return { success: true, transactionId: '...' }
  },
})

const agent = new Agent({ tools: [transferMoney] })
const result = await agent.invoke('Transfer $5000 from checking to savings')

if (result.stopReason === 'interrupt') {
  console.log('Interrupts:', result.interrupts)
  
  // Resume with user confirmation
  const resumed = await agent.invoke([
    { interruptResponse: { interruptId: result.interrupts[0].id, response: { confirmed: true } } }
  ])
}
```

### Interrupt from BeforeToolCallEvent Hook

```typescript
import { Agent, BeforeToolCallEvent } from '@strands-agents/sdk'

const agent = new Agent({
  tools: [deleteFileTool],
})

agent.addHook(BeforeToolCallEvent, (event) => {
  if (event.toolUse.name === 'delete_file') {
    const response = event.interrupt({
      name: 'confirm_delete',
      reason: 'User confirmation required before deleting file',
    })
    
    if (response !== 'approved') {
      event.cancel = 'Deletion cancelled by user'
    }
  }
})

const result = await agent.invoke('Delete the temp files')

if (result.stopReason === 'interrupt') {
  const resumed = await agent.invoke([
    { interruptResponse: { interruptId: result.interrupts[0].id, response: 'approved' } }
  ])
}
```

### Interrupt from BeforeToolsEvent Hook

```typescript
import { Agent, BeforeToolsEvent } from '@strands-agents/sdk'

agent.addHook(BeforeToolsEvent, (event) => {
  const dangerousTools = event.message.content
    .filter((block) => block.type === 'toolUseBlock')
    .filter((block) => ['delete_file', 'execute_command'].includes(block.name))

  if (dangerousTools.length > 0) {
    const response = event.interrupt({
      name: 'batch_approval',
      reason: `Approve ${dangerousTools.length} dangerous tool calls?`,
    })
    
    if (!response.approved) {
      event.cancel = 'Batch cancelled by user'
    }
  }
})
```

---

### 1. Core Interrupt Types (`src/interrupt.ts`)

Create new file with:

```typescript
// Interrupt class - represents a single interrupt
export class Interrupt {
  readonly id: string
  readonly name: string
  readonly reason?: unknown
  response?: unknown
}

// InterruptException - thrown internally to halt execution
export class InterruptError extends Error {
  readonly interrupt: Interrupt
}

// InterruptState - tracks interrupt state on Agent instance
export class InterruptState {
  interrupts: Map<string, Interrupt>
  context: Map<string, unknown>
  activated: boolean
  
  activate(): void
  deactivate(): void
  resume(responses: InterruptResponseContent[]): void
}
```

### 2. Type Definitions (`src/types/interrupt.ts`)

```typescript
export interface InterruptResponse {
  interruptId: string
  response: unknown
}

export interface InterruptResponseContent {
  interruptResponse: InterruptResponse
}

// Parameters for interrupt() method
export interface InterruptParams {
  name: string
  reason?: unknown
}
```

### 3. Content Block Extension (`src/types/messages.ts`)

Add `InterruptResponseBlock` as a new content block type:

```typescript
export class InterruptResponseBlock {
  readonly type = 'interruptResponseBlock' as const
  readonly interruptId: string
  readonly response: unknown
}
```

Update `ContentBlock` union to include `InterruptResponseBlock`.

### 4. StopReason Extension (`src/types/messages.ts`)

Add `'interrupt'` to the `StopReason` type.

### 5. AgentResult Extension (`src/types/agent.ts`)

Add `interrupts` array to `AgentResult`:

```typescript
export class AgentResult {
  // ... existing fields
  readonly interrupts?: Interrupt[]
}
```

### 6. Hook Event Extensions (`src/hooks/events.ts`)

Add `interrupt()` method to `BeforeToolCallEvent` and `BeforeToolsEvent`:

```typescript
export class BeforeToolCallEvent extends HookableEvent {
  // ... existing fields
  
  interrupt(params: InterruptParams): unknown {
    // Returns response if resuming, throws InterruptError if not
  }
}

export class BeforeToolsEvent extends HookableEvent {
  // ... existing fields
  
  interrupt(params: InterruptParams): unknown {
    // Returns response if resuming, throws InterruptError if not
  }
}
```

### 7. ToolContext Extension (`src/tools/tool.ts`)

Add `interrupt()` method to `ToolContext`:

```typescript
export interface ToolContext {
  toolUse: ToolUse
  agent: LocalAgent
  
  interrupt(params: InterruptParams): unknown
}
```

### 8. Tool Implementations (`src/tools/function-tool.ts`, `src/tools/zod-tool.ts`)

- Pass interrupt-capable `ToolContext` to callbacks
- Catch `InterruptError` in stream() and convert to appropriate event

### 9. Agent Loop Modifications (`src/agent/agent.ts`)

- Add `_interruptState: InterruptState` property to Agent
- Handle `InterruptError` in tool execution and hook callbacks
- Implement resume logic for `interruptResponse` content blocks
- Return `AgentResult` with `stopReason: 'interrupt'` and `interrupts` array when interrupted

### 10. Snapshot Integration

Ensure `InterruptState` is included in agent snapshots for SessionManager compatibility.

---

## Acceptance Criteria

- [ ] `Interrupt` and `InterruptState` classes implemented
- [ ] `interrupt()` method works on `BeforeToolCallEvent`
- [ ] `interrupt()` method works on `BeforeToolsEvent`  
- [ ] `interrupt()` method works on `ToolContext` (in tool callbacks)
- [ ] Agent returns `stopReason: 'interrupt'` with `interrupts` array
- [ ] Agent can be resumed with `interruptResponse` content blocks
- [ ] Second call to `interrupt()` returns user response
- [ ] Interrupt state integrates with agent snapshots
- [ ] Unit tests for all interrupt classes and methods
- [ ] Unit tests for hook interrupt functionality
- [ ] Unit tests for tool interrupt functionality
- [ ] Single integration test for full interrupt/resume flow

---

## Documentation References

- Python SDK interrupt implementation: https://github.com/strands-agents/sdk-python/blob/main/src/strands/interrupt.py
- Python SDK interrupt types: https://github.com/strands-agents/sdk-python/blob/main/src/strands/types/interrupt.py
- Python SDK tool decorator (InterruptException handling): https://github.com/strands-agents/sdk-python/blob/main/src/strands/tools/decorator.py

## Related Issues

- #477 - Interrupts - Before Tool Call Event (superseded by this issue)
- #478 - Interrupts - Before Tools Event (superseded by this issue)

zastrowm:
## Description

Adds an interrupt system for human-in-the-loop workflows. Agents can pause execution within tool callbacks, `BeforeToolCallEvent` hooks, or `BeforeToolsEvent` hooks to collect user input before proceeding.

The `interrupt()` method halts the agent loop on first call, returning `stopReason: 'interrupt'` with an `interrupts` array. When the caller resumes with `InterruptResponseContent` objects, the same `interrupt()` call returns the user's response instead of halting.

Multiple hook callbacks on the same event can each raise their own interrupt. The registry collects all interrupts across callbacks before halting, and duplicate interrupt names are rejected. This matches the Python SDK's behavior.

Both assistant and tool result messages are appended only after tool execution completes, preventing dangling `toolUse` blocks without matching results. When an interrupt fires mid-batch, completed tool results are preserved so the agent skips the model call on resume and only executes remaining tools. Hook-level interrupts (from `BeforeToolCallEvent`/`BeforeToolsEvent`) also store pending execution state, so resume skips the model call just like tool-level interrupts.

### End-to-end usage

```typescript
import { Agent, tool } from '@strands-agents/sdk'
import { z } from 'zod'

const transferMoney = tool({
  name: 'transfer_money',
  description: 'Transfer money between accounts',
  inputSchema: z.object({ amount: z.number() }),
  callback: (input, context) => {
    if (input.amount > 1000) {
      const response = context.interrupt({
        name: 'confirm_transfer',
        reason: 'Confirm large transfer?',
      })
      if (!response.confirmed) return 'Transfer cancelled'
    }
    return 'Transfer completed'
  },
})

const agent = new Agent({ model, tools: [transferMoney] })

// First call — agent pauses at interrupt
const result = await agent.invoke('Transfer $5000')
// result.stopReason === 'interrupt'
// result.interrupts === [{ id: '...', name: 'confirm_transfer', reason: 'Confirm large transfer?' }]

// Resume with user response — skips model call, re-executes tool with response
const resumed = await agent.invoke(
  result.interrupts.map((i) => ({
    interruptResponse: { interruptId: i.id, response: { confirmed: true } },
  }))
)
// resumed.stopReason === 'endTurn'
```

### New types and exports

- `Interrupt` class — interrupt data returned in `AgentResult.interrupts`
- `InterruptParams`, `InterruptResponse`, `InterruptResponseContent` types
- `InvokeArgs` now includes `InterruptResponseContent[]` as a valid input type
- `ToolContext` now includes `interrupt<T>(params: InterruptParams): T`
- `InterruptError` and `InterruptState` are internal and not exported

## Related Issues

- strands-agents/sdk-typescript#94
- Closes strands-agents/sdk-typescript#479
- Closes strands-agents/sdk-typescript#477
- Closes strands-agents/sdk-typescript#478

## Documentation PR

strands-agents/docs#778

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

## Checklist

----


src/tools/tool.ts
```diff
@@ -21,6 +22,37 @@ export interface ToolContext {
...
+   *       if (!response.confirmed) {
+   *         return { success: false, reason: 'Transfer cancelled' }
+   *       }
+   *     }
+   *     return { success: true }
+   *   },
+   * })
+   * ```
+   */
+  interrupt<T = unknown>(params: InterruptParams): T
```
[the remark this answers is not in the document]
zastrowm: Nah, that's okay, ToolContext is ours

src/hooks/events.ts
```diff
@@ -196,15 +198,62 @@ export class BeforeToolCallEvent extends HookableEvent {
...
+   *       reason: 'Confirm file deletion?',
+   *     })
+   *     if (response !== 'approved') {
+   *       event.cancel = 'Deletion cancelled by user'
+   *     }
+   *   }
+   * })
+   * ```
+   */
+  interrupt<T = unknown>(params: InterruptParams): T {
```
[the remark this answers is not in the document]
zastrowm: Going to take a look at adding an interface; I'm reluctant to add a base class for this method alone and mixins (or protocols) aren't a thing in TS as much as Python

src/interrupt.ts
```diff
@@ -0,0 +1,329 @@
...
+  /**
+   * The assistant message containing tool use blocks, serialized as MessageData.
+   */
+  assistantMessageData: MessageData
+
+  /**
+   * Tool results that were completed before the interrupt.
+   * Maps toolUseId to serialized ToolResultBlock data.
+   */
+  completedToolResults: Record<string, ContentBlockData>
```
pgrayy: This should be `ToolResultBlockData`.

src/interrupt.ts
```diff
@@ -0,0 +1,329 @@
...
+}
+
+/**
+ * Tracks the state of interrupt events raised during agent execution.
+ *
+ * Interrupt state is cleared after resuming.
+ */
+export class InterruptState {
```
pgrayy: Nit: Should this `implements InterruptStateData`

src/interrupt.ts
```diff
@@ -0,0 +1,329 @@
...
+  getPendingExecution(): { assistantMessage: Message; completedToolResults: Map<string, ToolResultBlock> } | undefined {
+    if (!this.pendingToolExecution) {
+      return undefined
+    }
+
+    const assistantMessage = Message.fromMessageData(this.pendingToolExecution.assistantMessageData)
+
+    const completedToolResults = new Map<string, ToolResultBlock>()
+    for (const [toolUseId, resultData] of Object.entries(this.pendingToolExecution.completedToolResults)) {
+      const block = contentBlockFromData(resultData)
```
pgrayy: We have a toolResultContentFromData as well.

src/tools/tool.ts
```diff
@@ -21,6 +22,36 @@ export interface ToolContext {
...
```
pgrayy: Should the default be JSONValue instead of unknown?

strands-ts/src/multiagent/graph.ts
```diff
@@ -494,7 +494,9 @@ export class Graph implements MultiAgent {
     const blocks =
       typeof input === 'string'
         ? [new TextBlock(input)]
-        : input.map((b) => ('type' in b ? (b as ContentBlock) : contentBlockFromData(b)))
+        : (input as Exclude<typeof input, string>).map((b) =>
+            'type' in b ? (b as ContentBlock) : contentBlockFromData(b as ContentBlockData)
+          )
```
pgrayy: Is this necessary with the changes to the MultiAgentInput definition?

src/hooks/events.ts
```diff
@@ -78,6 +80,33 @@ export abstract class HookableEvent extends StreamEvent {
   }
 }
 
+/**
+ * Interface for events that support human-in-the-loop interrupts.
+ * Mirrors Python's `_Interruptible` protocol.
```
pgrayy: Nit: Generally speaking would recommend not commenting on ways in which certain features are implemented in Python.

src/hooks/events.ts
```diff
@@ -78,6 +80,33 @@ export abstract class HookableEvent extends StreamEvent {
...
+ */
+interface Interruptible {
+  interrupt<T = unknown>(params: InterruptParams): T
+}
```
pgrayy: Nit: Would recommend defining this in interrupt.ts as it could be implemented by ToolContext as well. Does not have to be exported top level.

src/tools/tool.ts
pgrayy: Note that this could break users mocking ToolContext for unit tests on custom function tools. I'm not so concerned about that though since we are rc right now.

src/hooks/events.ts
```diff
@@ -78,6 +80,33 @@ export abstract class HookableEvent extends StreamEvent {
...
+
+/**
+ * Shared interrupt logic for hook events.
+ * Accesses the agent's interrupt state to register or resume an interrupt.
+ */
+function _interruptFromAgent<T>(agent: LocalAgent, interruptId: string, params: InterruptParams): T {
```
pgrayy: Nit: Useful for ToolContext as well. With that said, it might be worth finding a shared place for this and the Interruptible interface.

strands-ts/src/agent/snapshot.ts
```diff
@@ -104,6 +105,11 @@ export function takeSnapshot(agent: LocalAgent, options: TakeSnapshotOptions): S
     data.systemPrompt = agent.systemPrompt !== undefined ? (systemPromptToData(agent.systemPrompt) as JSONValue) : null
   }
 
+  if (fields.has('interrupts')) {
+    const interruptState = (agent as unknown as { _interruptState?: InterruptState })._interruptState
+    data.interrupts = interruptState ? (interruptState.toJSON() as unknown as JSONValue) : null
```
pgrayy, line 114: I guess the idea is to expose public methods on Agent to perform snapshotting, in which case, it will have direct access to _interruptState, meaning the awkward casting here will eventually go away.

src/hooks/events.ts
```diff
@@ -78,6 +80,33 @@ export abstract class HookableEvent extends StreamEvent {
...
+  const interruptState = (agent as unknown as { _interruptState?: InterruptState })._interruptState
```
pgrayy: This awkward casting seems to be a sign that we need to find a way to expose internal state publicly? We can figure that out in a follow up though.

strands-ts/src/agent/agent.ts
```diff
@@ -473,6 +491,17 @@ export class Agent implements LocalAgent, InvokableAgent {
...
 
+    // Handle interrupt responses if present in input
+    const interruptResponses = this._extractInterruptResponses(args)
+    if (interruptResponses.length > 0) {
+      this._interruptState.resume(interruptResponses)
+    } else {
+      // If user sends a regular message (not interrupt responses), clear any pending state
+      // This allows the user to "abandon" an interrupted workflow and start fresh
+      this._interruptState.clearPendingToolExecution()
+      this._interruptState.deactivate()
```
pgrayy: This works here in TS because we don't add tool uses until we have the tool results correct? In Python, this would not work because we would have a unanswered tool uses in the messages array.

src/agent/agent.ts
pgrayy: What events fire on interrupt and resume is something we should make more clear in the [docs](https://strandsagents.com/docs/user-guide/concepts/interrupts/). In Python, we skip the AfterToolCallEvent if interrupting a tool. The motivation here is that we don't have a tool result to feed it. We do however still emit AfterInvocationEvent. We also emit the before events on resume including BeforeToolCallEvent. This means we emit a BeforeToolCallEvent without a pairing AfterToolCallEvent. Maybe we want to change the behavior for TS?

pgrayy: Maybe it would also be worth adding some sort of field to the Before events that indicates whether or not the agent is resuming from an interrupt. There might be certain things a users doesn't want done if just resuming from an interrupt. This could be considered for follow up though.

strands-ts/src/hooks/events.ts
pgrayy, line 1: We should consider emitting an InterruptEvent. In Python, we have a ToolInterruptEvent, but maybe we can generalize. This could be considered for follow up though.

strands-ts/src/interrupt.ts
```diff
@@ -0,0 +1,329 @@
...
+  /**
+   * Gets or creates an interrupt with the given ID.
+   * If the interrupt already exists, returns it (potentially with a response).
+   *
+   * @param id - Unique identifier for the interrupt
+   * @param name - User-defined name for the interrupt
+   * @param reason - Optional reason for the interrupt
+   * @returns The interrupt (may have a response if resuming)
+   */
+  getOrCreateInterrupt(id: string, name: string, reason?: JSONValue): Interrupt {
```
pgrayy: In Python, we allow a user to pass in a response from within their hook or tool definitions. This is to skip the interrupt if they already have a response ready. For example, they might raise an interrupt in BeforeToolCallEvent and save the response in state for future invocations ([details](Can users provide a preemptive response)). Not absolutely necessary but customers can get slightly cleaner code:

```typescript
// not supported
agent.addHook(BeforeToolCallEvent, (event) => {
  if (event.toolUse.name !== 'delete_files') return

  const cached = event.agent.appState.get('approval')
  if (cached) return

  const response = event.interrupt({ name: 'approval', reason: 'Confirm deletion?' })
  event.agent.appState.set('approval', response)

  if (response !== 'y') {
    event.cancel = 'User denied permission'
  }
})
```

```typescript
// supported
agent.addHook(BeforeToolCallEvent, (event) => {
  if (event.toolUse.name !== 'delete_files') return

  const cached = event.agent.appState.get('approval')
  const response = event.interrupt({ name: 'approval', reason: 'Confirm deletion?', response: cached })
  event.agent.appState.set('approval', response)

  if (response !== 'y') {
    event.cancel = 'User denied permission'
  }
})
```

Not sure if it is worth it anymore. Can follow up later. Definitely doesn't need to be considered for this PR.

strands-ts/src/__fixtures__/tool-helpers.ts
pgrayy, line 1: Since interrupts are an internal mechanism, the unit tests should offer sufficient coverage. Still, I think it would be good to at least integ test the happy paths. We do this in Python ([here](https://github.com/strands-agents/sdk-python/tree/main/tests_integ/interrupts)). Could be considered for follow up.

src/agent/agent.ts
```diff
@@ -636,6 +698,30 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+      this.messages.length > 0
+        ? this.messages[this.messages.length - 1]!
+        : new Message({ role: 'assistant', content: [] })
+
+    return new AgentResult({
+      stopReason: 'interrupt',
+      lastMessage,
+      traces: this._tracer.localTraces,
+      metrics: this._meter.metrics,
+      interrupts: this._interruptState.getInterruptsList(),
```
pgrayy: We should only be returning the interrupts not yet responded to. This is what we do in Python ([here](https://github.com/strands-agents/sdk-python/blob/main/src/strands/event_loop/event_loop.py#L538)). We can get into this situation in a few ways:
* Customer raises interrupt in BeforeToolsEvent and then again in BeforeToolCallEvent.
* Customer raises multiple interrupts on a single event (can only be processed one at a time).
* Customer doesn't respond to all the interrupts the first time around.

src/agent/agent.ts
pgrayy, line 745: It seems like we are allowing users to mix in other content block types with their interrupt responses (in Python we raise an exception). It looks like if interrupt response isn't the first content item, we could end up populating `inputMessages` which get appended to the messages array down below (through a second call to normalizeInput).

strands-ts/src/__fixtures__/tool-helpers.ts
zastrowm: Added an integ test

strands-ts/src/agent/snapshot.ts
zastrowm: Yes, that's the theory. Depending on where we do the snapshot creation, then this may or may not go away.  But assuming this code is moved, then the casting goes away

strands-ts/src/hooks/events.ts
zastrowm: Follow-up in https://github.com/strands-agents/sdk-typescript/issues/872

strands-ts/src/interrupt.ts
```diff
@@ -0,0 +1,328 @@
...
+  }
+
+  /**
+   * Serializes the interrupt to a JSON-compatible object.
+   */
+  toJSON(): { id: string; name: string; reason?: unknown; response?: unknown } {
+    return {
+      id: this.id,
+      name: this.name,
+      ...(this.reason !== undefined && { reason: this.reason }),
```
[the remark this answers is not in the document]
zastrowm: Changed to JsonValue

strands-ts/src/agent/agent.ts
zastrowm: Yep. We defers appending both messages until after tools complete ([agent.ts L773-774](strands-ts/src/agent/agent.ts)), so on interrupt `messages` never has dangling `toolUse` blocks. The pending state is stored separately in `InterruptState.pendingToolExecution` instead.

strands-ts/src/interrupt.ts
zastrowm: Added for parity

src/agent/agent.ts
zastrowm: Updated

src/agent/agent.ts
zastrowm: I'll add this as part of the docs

My thinking is that individual events that are interrupted (ToolCall) do not have the end item called, but group ones (ToolCalls & Invocation) do.

LMK if you disagree

src/hooks/events.ts
zastrowm: It's because it's a private member; I think it's fine as it's equivalent to "internal" but we can adjust in the future as needed

src/hooks/events.ts
zastrowm: Moved to common location

src/agent/agent.ts
zastrowm: Added an error for this based on another comment — `_extractInterruptResponses` now throws a `TypeError` if the input contains both interrupt responses and non-interrupt content.

src/hooks/events.ts
zastrowm: Moved

src/tools/tool.ts
zastrowm: I'm not even concerned about mocks TBH; I think we need to be free to add to stuff like this as otherwise it's going to have a poor devx for new properties added after shipping

strands-ts/src/agent/agent.ts
zastrowm: Looking at this further, I realized I don't like this - it's sort of a foot gun, non-obvious behavior. If we think this is a useful feature later we can add something, but for now going to match Python's behavior of throwing instead

strands-ts/src/index.ts
```diff
@@ -31,6 +31,11 @@ export {
   StructuredOutputError,
 } from './errors.js'
 
+// Interrupt system
+export { Interrupt } from './interrupt.js'
+export type { Interruptible } from './interrupt.js'
```
pgrayy: We made this private in Python ([src](https://github.com/strands-agents/sdk-python/blob/b340dc4d390f7effbcf75db9311fdd20cf6ae67b/src/strands/types/interrupt.py#L79)). Did you have a reason for exporting here in TS?

strands-ts/src/types/interrupt.ts
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * User-defined name for the interrupt.
+   * Must be unique within a single hook callback or tool execution.
+   */
+  name: string
+
+  /**
+   * User-provided reason for the interrupt.
+   */
+  reason?: JSONValue
```
pgrayy, line 24: Consideration: In Python, we allow users to pass in a preemptive response ([src](https://github.com/strands-agents/sdk-python/blob/main/src/strands/types/interrupt.py#L88)). This is useful for situations where you have session managed a trust response. Of course, you don't have to pass it through the interrupt method. You can setup an if condition around the call. We save the customer some code though by building in the condition to interrupt:
```typescript
// If no past response for user, interrupt. Otherwise, return session managed response.
const response = event.interrupt("myInterrupt", "needs approval", event.agent.appState["userResponse"])
```

This is an additive ability though so no need to worry about in this PR if we do in fact want it for TS.

strands-ts/src/types/interrupt.ts
```diff
@@ -0,0 +1,61 @@
...
+   * User's response to the interrupt.
+   */
+  response: JSONValue
+}
+
+/**
+ * Content block containing a user response to an interrupt.
+ * Used when invoking an agent to resume from an interrupted state.
+ */
+export interface InterruptResponseContent {
```
pgrayy: For consistency with other content block types (e.g., [TextBlock](https://github.com/strands-agents/sdk-typescript/blob/main/strands-ts/src/types/messages.ts#L178)), should we make this a class with a type field and `toJSON` and `fromJSON`. Might also help to simplify `isInterruptResponseContent`.

strands-ts/src/tools/tool.ts
```diff
@@ -21,6 +23,36 @@ export interface ToolContext {
...
+   *       if (!response.confirmed) {
+   *         return { success: false, reason: 'Transfer cancelled' }
+   *       }
+   *     }
+   *     return { success: true }
+   *   },
+   * })
+   * ```
+   */
+  interrupt<T = JSONValue>(params: InterruptParams): T
```
pgrayy: Can this just extend Interruptible now?

strands-ts/src/index.ts
zastrowm: No; Removing

strands-ts/src/tools/tool.ts
zastrowm: Yes

strands-ts/src/agent/agent.ts
```diff
@@ -1023,11 +1149,38 @@ export class Agent implements LocalAgent, InvokableAgent {
...
 
+    // Check if a hook raised an interrupt — if so, store pending state and propagate
+    const beforeToolsInterrupt = this._interruptState.getUnansweredInterrupt()
+    if (beforeToolsInterrupt) {
+      this._interruptState.setPendingToolExecution({
+        assistantMessageData: assistantMessage.toJSON(),
+        completedToolResults: {},
+      })
+      throw new InterruptError(beforeToolsInterrupt)
+    }
```
pgrayy: Is this necessary? Don't you handle the setting and throwing in the catch block above?

strands-ts/src/agent/agent.ts
```diff
@@ -1059,6 +1212,15 @@ export class Agent implements LocalAgent, InvokableAgent {
         toolResultBlocks.push(...cancelBlocks)
       } else {
         for (const toolUseBlock of toolUseBlocks) {
+          // Check if this tool was already completed in a previous run
+          const existingResult = toolResults.get(toolUseBlock.toolUseId)
+          if (existingResult) {
+            // Skip already completed tools
+            toolResultBlocks.push(existingResult)
+            yield new ToolResultEvent({ agent: this, result: existingResult })
```
pgrayy: Do we need to yield the result event again?

strands-ts/src/agent/agent.ts
```diff
@@ -1112,9 +1294,19 @@ export class Agent implements LocalAgent, InvokableAgent {
 
     // Retry loop for tool execution
     while (true) {
-      const beforeToolCallEvent = new BeforeToolCallEvent({ agent: this, toolUse, tool })
+      const beforeToolCallEvent = new BeforeToolCallEvent({
+        agent: this,
+        toolUse,
+        tool,
+      })
       yield beforeToolCallEvent
```
pgrayy, line 1657: Do we need a try/catch around this like with with beforeToolsEvent since you inject the interrupt error into the generator?

strands-ts/src/agent/agent.ts
```diff
@@ -1163,6 +1355,16 @@ export class Agent implements LocalAgent, InvokableAgent {
             input: toolUseBlock.input,
           },
           agent: this,
+          interrupt: <T = JSONValue>(params: InterruptParams): T => {
```
pgrayy, line 1719: Nit: You could use your helper function here right?

strands-ts/src/agent/snapshot.ts
pgrayy, line 114: If someone sets the snapshotting strategy to "message", I think we might end up missing interrupt state snapshotting if the user immediately shuts down the process after invocation ([src](https://github.com/strands-agents/sdk-typescript/blob/main/strands-ts/src/session/session-manager.ts#L205)). In Python, we always session persist on AfterInvocationEvent ([src](https://github.com/strands-agents/sdk-python/blob/main/src/strands/session/session_manager.py#L52)).

This concern might actually go beyond interrupts. We should probably always snapshot on after invocation event. Users can workaround this for now by setting a snapshotTrigger callback if their latest strategy is "message".

zastrowm: Good catch; created a bug to track in #956

strands-ts/src/types/interrupt.ts
zastrowm: Yep; Added response to InterruptParams. Works the same as Python — if provided, the interrupt returns immediately without halting.

strands-ts/src/types/interrupt.ts
zastrowm: Very nice catch; done

strands-ts/src/agent/agent.ts
zastrowm: ~~Yes; the catch stores pending execution state before re-throwing. Without the re-throw the interrupt would be swallowed. Added a clarifying comment.~~

strands-ts/src/agent/agent.ts
zastrowm: No; matching Python behavior, removing this

strands-ts/src/agent/agent.ts
zastrowm: Yes; another good catch; would have introduced a bug elsewhere otherwise :)

strands-ts/src/agent/agent.ts
zastrowm: No; it propagates up to the executeTools catch block which stores pending state. Added a test that verifies this: interrupts from BeforeToolCallEvent, asserts pendingToolExecution is stored, and confirms resume works.

strands-ts/src/agent/agent.ts
zastrowm: It's not needed; it should be handled via other mechanisms now

The other one related to beforeToolCallEvent can be/is removed too

pgrayy (approved):

PR #790, opened 2026-04-03
feat(context): track agent.messages token size

lizradway:
## Description

Adds visibility into LLM context window utilization by exposing the most recent `inputTokens` count through two new properties:

- **`AgentMetrics.latestContextSize`** — the most recent `inputTokens` value from model responses, enabling threshold-based decisions about context compression or externalization
- **`AgentResult.contextSize`** — convenience getter that delegates to `metrics.latestContextSize`

Zero overhead (reuses `inputTokens` already returned by every LLM call), zero latency impact (no additional API calls), and provider-agnostic (all providers normalize to the `Usage` interface).

### Usage

```typescript
const result = await agent.invoke('Do something with tools')

// Via AgentResult (convenience)
result.contextSize

// Via AgentMetrics
result.metrics?.latestContextSize
```

Port of [strands-agents/sdk-python#2009](https://github.com/strands-agents/sdk-python/pull/2009).

## Related Issues

https://github.com/strands-agents/sdk-typescript/issues/791

## Documentation PR

N/A

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`
- Added unit tests for `Meter.latestContextSize` tracking (undefined on fresh instance, updates after model calls, updates across cycles/invocations, handles missing usage)
- Added unit tests for `AgentMetrics.toJSON` with/without `latestContextSize`
- Added unit tests for `AgentResult.contextSize` delegation (with metrics, without context size, without metrics)

## Checklist

### DIFF
--- a/src/telemetry/__tests__/meter.test.ts
+++ b/src/telemetry/__tests__/meter.test.ts
@@ -272,6 +272,67 @@ describe('Meter', () => {
     })
   })
 
+  describe('latestContextSize', () => {
+    it('is undefined when no invocations have occurred', () => {
+      expect(meter.metrics.latestContextSize).toBeUndefined()
+    })
+
+    it('returns the most recent inputTokens after a model call', () => {
+      meter.updateCycle({
+        type: 'modelMetadataEvent',
+        usage: { inputTokens: 100, outputTokens: 50, totalTokens: 150 },
+      })
+
+      expect(meter.metrics.latestContextSize).toBe(100)
+    })
+
+    it('updates across multiple cycles', () => {
+      meter.updateCycle({
+        type: 'modelMetadataEvent',
+        usage: { inputTokens: 100, outputTokens: 50, totalTokens: 150 },
+      })
+      meter.updateCycle({
+        type: 'modelMetadataEvent',
+        usage: { inputTokens: 200, outputTokens: 80, totalTokens: 280 },
+      })
+
+      expect(meter.metrics.latestContextSize).toBe(200)
+    })
+
+    it('updates across multiple invocations', () => {
+      meter.startNewInvocation()
+      meter.startCycle()
+      meter.updateCycle({
+        type: 'modelMetadataEvent',
+        usage: { inputTokens: 100, outputTokens: 50, totalTokens: 150 },
+      })
+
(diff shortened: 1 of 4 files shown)


src/telemetry/__tests__/meter.test.ts
```diff
@@ -272,6 +272,67 @@ describe('Meter', () => {
...
+      })
+
+      meter.startNewInvocation()
+      meter.startCycle()
+      meter.updateCycle({
+        type: 'modelMetadataEvent',
+        usage: { inputTokens: 300, outputTokens: 100, totalTokens: 400 },
+      })
+
+      expect(meter.metrics.latestContextSize).toBe(300)
```
opieter-aws, line 317: Nit: it looks like the `latestContextSize` value is overridden in this case, but from the `startNewInvocation()` method it looks like it isn't reset. I assume we want it return `undefined` in case of no cycles on the latest invocation?

src/telemetry/meter.ts
```diff
@@ -474,6 +496,7 @@ export class Meter {
    */
   private _updateUsage(usage: Usage): void {
     accumulateUsage(this._accumulatedUsage, usage)
+    this._latestContextSize = usage.inputTokens
```
opieter-aws, line 498: Should this be `inputTokens + outputTokens` to get a more accurate number to use for current context size?

lizradway: `inputTokens` is the actual ground truth measured context size reported by the provider for that call.

`inputTokens + outputTokens` misses tool results that get appended between calls, and those output tokens already appear in the next call’s inputTokens anyway and would be double counted.

src/telemetry/__tests__/meter.test.ts
lizradway: The `latestContextSize` intentionally persists across invocations. Between invocations the conversation only grows (no messages are removed), so the last reported `inputTokens` remains a reasonable lower bound until the next model call updates it. Resetting to `undefined` would discard useful information; imo a stale estimate is more useful than no value. This matches the current Python SDK behavior as well, but would be easy to update either way in Python.

opieter-aws (approved):

PR #794, opened 2026-04-06
fix: migrate MultiagentPlugin to be an interface

JackYPCOnline:
## Description

Changes `MultiAgentPlugin` from an `abstract class` to an `interface`, aligning it with the existing `Plugin` pattern used for single-agent plugins.

### Motivation

`MultiAgentPlugin` was defined as an `abstract class` but had zero concrete methods — it was effectively an interface. This created an inconsistency with `Plugin` (which is an `interface`) and forced consumers to use `implements` anyway (since TypeScript only allows single inheritance with `extends`). For example, `SessionManager` already declares `implements Plugin, MultiAgentPlugin` — this worked due to structural typing but was semantically incorrect for an abstract class.

### Related PR
https://github.com/strands-agents/sdk-typescript/pull/654

## Related Issues

N/A

## Documentation PR

N/A — no user-facing API change (consumers already used `implements`).

## Type of Change

Refactor (non-breaking)

## Testing

- [x] I ran `npm run check`
- [x] All 1834 unit tests pass across 67 test files, no type errors

## Checklist

----

### DIFF
--- a/src/multiagent/plugins.ts
+++ b/src/multiagent/plugins.ts
@@ -1,7 +1,7 @@
 /**
  * Plugin interface and registry for extending multi-agent orchestrator functionality.
  *
- * This module defines the MultiAgentPlugin abstract class and MultiAgentPluginRegistry,
+ * This module defines the MultiAgentPlugin interface and MultiAgentPluginRegistry,
  * which provide a composable way to add behavior to multi-agent orchestrators (e.g. Swarm, Graph)
  * through hook registration and custom initialization.
  */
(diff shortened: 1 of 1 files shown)


src/multiagent/plugins.ts
```diff
@@ -1,27 +1,27 @@
...
- * Abstract base class for plugins that extend multi-agent orchestrator functionality.
+ * Interface for objects that implement multi-agent orchestrator plugin functionality.
  *
  * MultiAgentPlugins provide a composable way to add behavior to orchestrators
  * by registering hook callbacks in their `initMultiAgent` method.
  *
  * @example
  * ```typescript
- * class LoggingPlugin extends MultiAgentPlugin {
+ * class LoggingPlugin implements MultiAgentPlugin {
```
mkmeral: did we release this already? is this a backwards incompatible change?

mkmeral (approved):

PR #796, opened 2026-04-07
fix: prevent invocation lock leak when consumer breaks from stream

pgrayy:
## Description

When a consumer breaks out of `for-await-of` on `agent.stream()`, the runtime calls `.return()` on the generator. The `finally` block in `stream()` drains remaining events from the inner `_stream()` generator by yielding them. However, with no consumer to resume the generator, it suspends permanently at that `yield`, preventing scope exit. The `using _lock` disposable never runs, leaving `_isInvoking` stuck at `true`. Any subsequent call to `invoke()` or `stream()` throws `ConcurrentInvocationError`.

The fix tracks whether `finally` was entered via an exception (where the consumer may still be iterating and should receive drain events) or via `.return()` (consumer left). Drain events are only yielded in the error case; otherwise callbacks are invoked without yielding so the generator completes and the lock releases normally.

## Related Issues

None

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

Extended the existing "fires AfterInvocationEvent when consumer breaks from stream" test to also verify the agent is reinvokable after the break. Without the fix, this assertion fails with `ConcurrentInvocationError`.

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/src/agent/__tests__/agent.test.ts
+++ b/src/agent/__tests__/agent.test.ts
@@ -181,7 +181,7 @@ describe('Agent', () => {
     })
 
     describe('hook error cleanup', () => {
-      it('fires AfterInvocationEvent when consumer breaks from stream', async () => {
+      it('fires AfterInvocationEvent when consumer breaks from stream and allows reinvocation', async () => {
         const model = new MockMessageModel()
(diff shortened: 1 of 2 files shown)


notowen333 (approved):
discussed offline that we may follow up whether we should yield in the finally but this passes the integ test that discovered the bug in the first place so should be functionally correct

PR #800, opened 2026-04-08
feat: add swarm+session manager resume logic,unit tests, integration test

JackYPCOnline:
## Description

Add swarm resume logic, per-node save strategy for multi-agent session manager, `sessionManager` constructor arg, and JSDoc fixes.

### Session Manager changes:

- **New `'node'` save strategy (default):** Added `'node'` to `MultiAgentSaveLatestStrategy` which saves a snapshot after every node invocation completes via `AfterNodeCallEvent`. This enables resume from the last completed node after a crash or restart. The previous `'invocation'` strategy (saves only after the entire orchestrator invocation completes) is still available as an opt-in.
- **Breaking default change:** `multiAgentSaveLatestOn` now defaults to `'node'` instead of `'invocation'`. Existing users who relied on the implicit `'invocation'` default will now get per-node saves (higher I/O but more durable). They can opt back in with `multiAgentSaveLatestOn: 'invocation'`.

**We don't need to worry about `default` change, no customer use it yet, it was introduced in my PR 2 days ago, and it is not in top-level `index`**

- **Both strategies always save on `AfterMultiAgentInvocationEvent`** as a final save when the orchestrator finishes (ensures state is captured even in error/finally paths).
- Scoped `SaveLatestStrategy` JSDoc to agents (removed "or orchestrator" from `'invocation'` description, clarified `'message'` is agent-only)
- Split `multiAgentSaveLatestOn` JSDoc from single-line to multi-line, updated default to `'node'`

### Swarm changes:

- **`sessionManager` constructor arg:** Added `sessionManager` as a dedicated constructor parameter on `Swarm`, matching the `Agent` API. It is merged into the plugin registry internally. Throws if the same session manager is also passed in `plugins` to prevent duplicate registration.
- **`state.results.push` moved into `_streamNode`:** `state.results.push(nodeResult)` now happens inside `_streamNode` (before `AfterNodeCallEvent`) instead of in the `_stream` loop (after `_streamNode` returns). This ensures `state.results` is populated when the per-node snapshot fires on `AfterNodeCallEvent`.
- **Simplified `_findResumeNode(state)`:** Checks the last entry in `state.results` for a pending handoff. Since results are pushed in completion order before `AfterNodeCallEvent`, the last item is always the most recent completed node — no need to iterate `state.nodes` or compare `startTime` values.
- On resume, the swarm starts from the pending handoff target instead of always restarting from `this.start`. Steps count carries forward from the restored state.
- If the previous run completed normally (no pending handoff), the swarm starts fresh.
- Removed orphaned JSDoc comment (`Options for creating a Swarm instance.`) that was dangling above `SwarmNodeDefinition`.

## Type of Change

New feature

### Unit tests (`src/multiagent/__tests__/swarm.test.ts`) — 6 new tests:

- Resume from pending handoff (A→B stopped, resumes at B)
- Fresh start after completed run (no false resume)
- Mid-chain resume (A→B→C, stopped after B, resumes at C)
- Steps carry-forward across restarts
- Handoff context preservation on resume
- Fallback when resume target agent was removed from the swarm

### Unit tests (`src/session/__tests__/session-manager.test.ts`) — 2 updated tests:

- Registers `AfterNodeCallEvent` hook by default (node strategy)
- Registers `AfterMultiAgentInvocationEvent` hook when strategy is invocation
- Saves snapshot after node call when `multiAgentSaveLatestOn` is `'node'` (default)
- Saves snapshot after invocation when `multiAgentSaveLatestOn` is `'invocation'`

### Integration test (`test/integ/multiagent/session-manager-multiagent-integration.test.node.ts`) — 1 new test:

- Real Bedrock model + FileStorage, validates full stop/resume flow
- Asserts researcher executes only once (restored, not re-executed) and writer is the resumed node

- [x] I ran `npm run check`

## Checklist

----


test/integ/multiagent/session-manager.test.node.ts
pgrayy, line 1: I would recommend naming this file session-manager.test.node.ts. It is present in the multiagent integ test folder so we don't need the other components.

src/multiagent/swarm.ts
```diff
@@ -383,6 +386,51 @@ export class Swarm implements MultiAgent {
...
+   *
+   * If the last result has no `agentId`, the previous run completed normally
+   * and there is nothing to resume.
+   *
+   * @returns The handoff target node and its handoff context, or `undefined` for a fresh start
+   */
+  private _findResumeNode(state: MultiAgentState): { node: AgentNode; lastHandoff: HandoffResult } | undefined {
+    // Find the last completed node by checking per-node state.
+    // We use state.nodes rather than state.results because with the 'node' save strategy,
+    // AfterNodeCallEvent fires before state.results is populated.
```
pgrayy: I would recommend moving the state.results push then. state.results is pushed in order which means you just have to check the last item to get the restore point. This should simplify the logic and help you avoid having to look at start times.

pgrayy: Should just have to move `state.results.push(nodeResult)` to line 278 and 299.

src/multiagent/swarm.ts
```diff
@@ -200,10 +200,13 @@ export class Swarm implements MultiAgent {
       input,
     })
 
+    // Plugins (e.g. SessionManager) may restore state.results here via the hook
```
pgrayy: Are you planning on supporting session manager as a constructor arg? We should match Agent and accept it as a dedicated parameter.

JackYPCOnline: Good call out, will add it

src/multiagent/swarm.ts
```diff
@@ -293,6 +309,7 @@ export class Swarm implements MultiAgent {
 
       const result = next.value
       this._tracer.endNodeSpan(nodeSpan, { status: result.status, duration: result.duration, usage: result.usage })
+      state.results.push(result)
```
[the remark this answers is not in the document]
JackYPCOnline: I think we should adjust graph code as well

poshinchen (approved):

PR #809, opened 2026-04-13
feat: add graph+session manager integration + tests

JackYPCOnline:
## Description

Adds Graph resume support via SessionManager, bringing Graph to parity with Swarm's existing session management capabilities.

### Changes

**`src/multiagent/graph.ts`**
- Added `sessionManager` as a first-class constructor option on `GraphOptions` (matching Swarm's API), with duplicate-detection guard that throws if the same session manager appears in both `sessionManager` and `plugins`.
- Added `_findResumeTargets()` — on resume, identifies nodes that are ready to execute based on restored state. Any non-COMPLETED node whose dependencies are all COMPLETED (with edge handlers satisfied) becomes a resume target. Returns `undefined` for fresh starts (no restored state or all nodes already completed).
- Extracted `_allDependenciesSatisfied()` to share dependency-checking logic between `_findReady()` and `_findResumeTargets()`. This also fixes a subtle behavioral difference: the old `_findReady` evaluated edge handlers in parallel via `Promise.all`, while the new shared method evaluates sequentially with short-circuit, which is safer for handlers with side effects.
- Refactored `_findReady()` to use `_allDependenciesSatisfied()`, eliminating duplicated dependency-check logic.

**`src/multiagent/__tests__/graph.test.ts`**
- Added `resume with session manager` test suite covering:
  - Linear graph resume (A→B→C, A completes, maxSteps stops, resumes at B)
  - Parallel branch resume (A→B, A→C, C cancelled, resumes at C)
  - Fresh start when all nodes already completed
  - Conditional edge handlers respected on resume (handler returns false → node stays blocked)

**`test/integ/multiagent/session-manager.test.node.ts`**
- Added Graph resume integration tests alongside existing Swarm tests:
  - Parallel branches + sub-graph resume: tests `sessionManager` constructor arg, `MultiAgentNode` resume, AND-join across resume boundary, cross-boundary data flow
  - Conditional edge handler resume: tests that `_findResumeTargets` evaluates edge handlers (not just node statuses), with a `BeforeMultiAgentInvocationEvent` hook modifying `state.app` to flip the handler result between runs

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

### Unit tests
4 new tests in `src/multiagent/__tests__/graph.test.ts` covering resume scenarios with `MockSnapshotStorage`.

### Integration tests
2 new tests in `test/integ/multiagent/session-manager.test.node.ts` using real Bedrock models and `FileStorage` to verify end-to-end resume across process boundaries.

## Checklist

----

### DIFF
--- a/src/multiagent/graph.ts
+++ b/src/multiagent/graph.ts
@@ -548,11 +548,15 @@ export class Graph implements MultiAgent {
       }
     }
 
-    logger.debug(
-      `resume_targets=<${ready.map((n) => n.id).join(', ')}>, prior_steps=<${state.steps}> | resuming graph from restored state`
-    )
-    // Empty ready set means all nodes completed — start fresh
-    return ready.length > 0 ? ready : undefined
+    if (ready.length > 0) {
+      logger.debug(
+        `resume_targets=<${ready.map((n) => n.id).join(', ')}>, prior_steps=<${state.steps}> | resuming graph from restored state`
+      )
+      return ready
+    }
+
+    logger.debug('all nodes completed in restored state | starting fresh')
+    return undefined
   }
 
   /**
--- a/src/multiagent/graph.ts
+++ b/src/multiagent/graph.ts
@@ -3,10 +3,12 @@ import type { InvokableAgent } from '../types/agent.js'
 import type { MultiAgentInput } from './multiagent.js'
 import type { ContentBlock } from '../types/messages.js'
 import { TextBlock, contentBlockFromData } from '../types/messages.js'
+import { logger } from '../logging/logger.js'
 import { HookableEvent } from '../hooks/events.js'
 import { HookRegistryImplementation } from '../hooks/registry.js'
 import type { HookCallback, HookableEventConstructor, HookCleanup } from '../hooks/types.js'
 import type { MultiAgentPlugin } from './plugins.js'
+import type { SessionManager } from '../session/session-manager.js'
 import { MultiAgentPluginRegistry } from './plugins.js'
 import type { NodeDefinition } from './nodes.js'
 import { AgentNode, MultiAgentNode, Node } from './nodes.js'
@@ -53,6 +55,8 @@ export interface GraphOptions extends GraphConfig {
   edges: EdgeDefinition[]
   /** Explicit source node IDs. If omitted, auto-detected from nodes with no incoming edges. */
   sources?: string[]
+  /** Session manager for saving and restoring graph sessions. */
+  sessionManager?: SessionManager
   /** Plugins for event-driven extensibility. */
   plugins?: MultiAgentPlugin[]
   /** Custom trace attributes to include on all spans. */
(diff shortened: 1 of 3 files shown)


src/multiagent/graph.ts
```diff
@@ -120,8 +125,17 @@ export class Graph implements MultiAgent {
     this._sources = this._resolveSources(sources)
     this._validateSources()
 
+    this.sessionManager = sessionManager
+
+    if (sessionManager && plugins?.some((p) => p.name === sessionManager.name)) {
```
mkmeral, line 130: so this is essentially checking if there are multiple plugins with the same name, right? is there any reason we're checking it here and not in multiagent plugin registry? or is this just error sugar, kind of?

JackYPCOnline: this varibale `plugins` in Swarm/Graph is MultiagentPlugin type, I added session manager as constructor arg for consistency. User can pass in plugin or arg, just prevent double registery.

src/multiagent/graph.ts
```diff
@@ -498,6 +515,61 @@ export class Graph implements MultiAgent {
...
+   * - EXECUTING/FAILED/CANCELLED nodes from the previous run
+   * - Source nodes (no incoming edges) that are not COMPLETED
+   *
+   * Works for all node types including {@link AgentNode} and {@link MultiAgentNode}
+   * (subgraphs/swarms). A `MultiAgentNode` that didn't complete will be re-executed
+   * from scratch — its inner orchestrator manages its own state independently.
+   *
+   * @returns Array of ready nodes, or `undefined` if state was not restored (fresh start)
+   */
+  private async _findResumeTargets(state: MultiAgentState): Promise<Node[] | undefined> {
```
mkmeral, line 533: <details>
<summary>nit (possible followup? feature request?): graph topology update. consider a dev, that is updating their graph topology to better serve their customers. how can we handle that better?</summary>


### Topology mismatch on resume: node added after snapshot

If a node exists in the graph but not in the restored snapshot (e.g. user adds a node between save and resume), `loadStateFromJSON` clears `_nodes` and rebuilds only from snapshot data. The new node won't have a `NodeState` entry. `_findResumeTargets` still picks it up as a resume candidate (deps satisfied), but `_streamNode` does `state.node(node.id)!` → `undefined` → runtime crash on `nodeState.status = Status.EXECUTING`.

Swarm has the same latent bug — `loadStateFromJSON` wipes `_nodes` and rebuilds from snapshot only. If the model hands off to a node that was added after the snapshot, `_streamNode` crashes the same way. Harder to trigger in practice since the handoff target comes from model output, but still possible.

**Suggested fix:** Add an `ensureNode(id)` method on `MultiAgentState` that creates a PENDING `NodeState` if missing. Then reconcile in `_findResumeTargets` (and Swarm's `_stream` after restore):

```typescript
// MultiAgentState
ensureNode(id: string): NodeState {
  let ns = this._nodes.get(id)
  if (!ns) {
    ns = new NodeState()
    this._nodes.set(id, ns)
  }
  return ns
}
```


</details>

mkmeral (approved):

PR #810, opened 2026-04-13
feat(a2a): support custom ClientFactory in A2AAgent for authenticated requests

### LINKED ISSUE #802: [FEATURE] A2AAgent pass custom a2a client factory for authenticating requests
### Problem Statement

The `A2AAgent` currently accepts the following parameters:

https://github.com/strands-agents/sdk-typescript/blob/afb3912898c4484cef17005cbe425002b1bfe648/src/a2a/a2a-agent.ts#L23-L34

There seems to be no way to specify any parameters used for authentication with an A2A Server, such as a bearer token or signing a request with sigv4.

The Python SDK allows this by exposing httpx arguments in the `A2AAgentToolProvider`:

https://github.com/strands-agents/tools/blob/e172b1b0c32317618230dbd01866ae4c604b7094/src/strands_tools/a2a_client.py#L45-L54

And the equivalent `A2AAgent` class in python allows a custom a2a client factory to be passed:

https://github.com/strands-agents/sdk-python/blob/ca6f599d403a0d190d4f102f608e21034c79a439/src/strands/agent/a2a_agent.py#L31-L41

(though note that at the time of writing the `A2AAgent` in Python still fetches the agent card without any authentication https://github.com/strands-agents/sdk-python/issues/2006 - so currently `A2AAgentToolProvider` seems to be the simplest way connect A2A Agents on bedrock agentcore)

### Proposed Solution

Allow a custom A2A `ClientFactory` to be passed to `A2AAgent`. This in turn will allow a custom `fetchImpl` to be provided in order to authenticate requests:

```typescript
const sigv4Fetch = ...;

// Create a ClientFactory with SigV4 fetch for both card resolution and message transport
const a2aClientFactory = new ClientFactory(
  ClientFactoryOptions.createFrom(ClientFactoryOptions.default, {
    transports: [new JsonRpcTransportFactory({ fetchImpl: sigv4Fetch })],
    cardResolver: new DefaultAgentCardResolver({ fetchImpl: sigv4Fetch }),
  }),
);

new A2AAgent({ a2aClientFactory });
```

### Use Case

If invoking an A2AAgent on [Bedrock AgentCore](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-a2a.html), requests to the agent must be authenticated either by JWT or IAM.

### Alternatives Solutions

_No response_

### Additional Context

_No response_

cogwirrel:
## Motivation

The `A2AAgent` currently always creates a default `ClientFactory` internally, providing no way to customize how HTTP requests are made to the remote A2A server. This prevents users from authenticating requests with mechanisms like SigV4 signing or bearer tokens, which is required when invoking A2A agents on services like [Bedrock AgentCore](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-a2a.html).

The Python SDK already supports this by accepting a custom A2A client factory in its `A2AAgent` class.

Resolves: #802

## Public API Changes

`A2AAgentConfig` now accepts an optional `clientFactory` parameter:

```typescript
// Before: no way to authenticate requests
const agent = new A2AAgent({ url: "http://localhost:9000" })

// After: pass a custom ClientFactory for authenticated requests
const sigv4Fetch = ...
const a2aClientFactory = new ClientFactory(
  ClientFactoryOptions.createFrom(ClientFactoryOptions.default, {
    transports: [new JsonRpcTransportFactory({ fetchImpl: sigv4Fetch })],
    cardResolver: new DefaultAgentCardResolver({ fetchImpl: sigv4Fetch }),
  }),
)

const agent = new A2AAgent({
  url: "http://localhost:9000",
  clientFactory: a2aClientFactory,
})
```

When no `clientFactory` is provided, the default `ClientFactory` is used — fully backward compatible.

## Use Cases

- **SigV4 authentication**: Sign requests when invoking A2A agents on Bedrock AgentCore
- **Bearer token authentication**: Pass JWT or other bearer tokens for authenticated endpoints
- **Custom transport**: Provide any custom fetch implementation for specialized networking requirements

### DIFF
--- a/src/a2a/__tests__/a2a-agent.test.ts
+++ b/src/a2a/__tests__/a2a-agent.test.ts
@@ -190,6 +190,29 @@ describe('A2AAgent', () => {
       await agent.invoke('Hello')
       expect(mockGetAgentCard).toHaveBeenCalledOnce()
     })
+
+    it('uses custom clientFactory when provided', async () => {
+      const customSendMessageStream = vi.fn().mockReturnValue(mockStream(createMockTaskResponse()))
+      const customGetAgentCard = vi.fn().mockResolvedValue(mockAgentCard)
+      const customCreateFromUrl = vi.fn().mockResolvedValue({
+        sendMessageStream: customSendMessageStream,
+        getAgentCard: customGetAgentCard,
+      })
+      const customFactory = { createFromUrl: customCreateFromUrl }
+
+      const agent = new A2AAgent({
+        url: 'http://localhost:9000',
+        clientFactory: customFactory as never,
+      })
+
+      await agent.invoke('Hello')
+
+      expect(customCreateFromUrl).toHaveBeenCalledWith('http://localhost:9000', undefined)
+      expect(customGetAgentCard).toHaveBeenCalledOnce()
+      expect(customSendMessageStream).toHaveBeenCalledOnce()
+      // Default mock should not have been called
+      expect(mockSendMessageStream).not.toHaveBeenCalled()
+    })
   })
 
   describe('stream', () => {
--- a/src/a2a/a2a-agent.ts
+++ b/src/a2a/a2a-agent.ts
@@ -8,7 +8,7 @@
  */
 
 import type { AgentCard, Part } from '@a2a-js/sdk'
-import type { Client as A2AClientSdk } from '@a2a-js/sdk/client'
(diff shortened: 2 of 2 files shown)


poshinchen (approved):
This looks neat to me

mkmeral (approved):

Issue #551, opened 2026-02-19
Models - OpenAI - Responses API (stateful model support)


pgrayy:
### Summary

Extend the OpenAI model provider to support the [Responses API](https://developers.openai.com/api/docs/guides/migrate-to-responses/), which is OpenAI's successor to the Chat Completions API. The current `OpenAIModel` implementation uses the Chat Completions API exclusively. The Responses API introduces built-in tool use, streaming improvements, and a simplified conversation state model. We should determine whether to add Responses API support alongside the existing Chat Completions path or as a separate provider.

### Usage

> ⚠️ **Agent-generated example** - This is a preliminary API sketch and has not been finalized.

```typescript
import { Agent } from '@strands-agents/sdk'
import { OpenAIModel } from '@strands-agents/openai'

const agent = new Agent({
  model: new OpenAIModel({
    modelId: 'gpt-4o',
    apiType: 'responses', // opt into Responses API
  }),
})

const result = await agent.invoke('What is the weather in Seattle?')
```

### Documentation

N/A

### Reference

N/A

notowen333:
docs merged closing

PR #820, opened 2026-04-17
feat: add openai-responses model provider and stateful model support

notowen333:
### Motivation

OpenAI's Responses API manages conversation state server-side, sending only the latest message per turn and chaining via `previous_response_id`. This is fundamentally different from the Chat Completions API (which requires the full message history each turn) and enables built-in tools like `web_search` and `code_interpreter` that aren't available through Chat Completions. The SDK had no way to model this server-managed conversation pattern.

### Public API Changes

**Unified `OpenAIModel` with an `api` discriminator**

The Responses API is selected via `api: 'responses'`; omitting `api` (or passing `'chat'`) keeps the existing Chat Completions behavior. `api` is construction-only — changing it via `updateConfig` is ignored with a warning, since the field gates message-history management and `previous_response_id` chaining.

```typescript
import { Agent, OpenAIModel } from '@strands-agents/sdk'

// Responses API (default, changed)
const model = new OpenAIModel({ modelId: 'gpt-5.4' })

Before the chat completions mode was default. This aligns with python.

// Responses API — stateful by default
const model = new OpenAIModel({ api: 'responses', modelId: 'gpt-4o' })
const agent = new Agent({ model, systemPrompt: 'You are helpful.' })

// Multi-turn: server tracks context, local messages cleared automatically
await agent.invoke('My name is Alice.')
const result = await agent.invoke('What is my name?') // Server remembers "Alice"

// With built-in web search
const model = new OpenAIModel({
  api: 'responses',
  modelId: 'gpt-4o',
  params: { tools: [{ type: 'web_search' }] },
})

// Stateless mode (full history sent each turn, like Chat Completions)
const model = new OpenAIModel({ api: 'responses', modelId: 'gpt-4o', stateful: false })
```

`OpenAIModelOptions` is discriminated on `api` so `'responses'` narrows to expose `stateful`, and `'chat'` narrows to expose `frequencyPenalty` / `presencePenalty`. Per-API config types (`OpenAIChatConfig`, `OpenAIResponsesConfig`) are exported alongside `OpenAIModelConfig` and `OpenAIModelOptions`.

**Stateful model abstraction on `Model` base class**

```typescript
// Model providers can now declare server-side state management
class Model {
  get stateful(): boolean { return false } // default
}

// Agent exposes modelState for provider-specific persistence (e.g. response IDs)
agent.modelState // Record<string, JSONValue>
```

When `model.stateful` is `true`, the agent automatically uses `NullConversationManager` (throws if a `conversationManager` is also supplied), clears local messages after each invocation via `ModelPlugin`, and passes `modelState` through to the provider on each call. `modelState` is included in snapshots for session persistence.

**Provider-managed fields in `params` are ignored with a warning**

Fields the adapter owns cannot be overridden via `params` — a warning fires at config time, and the managed values win at request construction:

- Chat Completions: `model`, `messages`, `stream`, `stream_options`
- Responses API: `model`, `input`, `stream`, `store`

### Use Cases

- **Server-managed conversations**: Use the Responses API for multi-turn agents without managing message history client-side
- **Built-in tools**: Access OpenAI's `web_search`, `code_interpreter`, and other built-in tools that are only available through the Responses API
- **Session persistence**: `modelState` (containing `responseId`) is captured in snapshots, enabling session save/restore across process restarts

## Related Issues

#551

## Documentation PR

will follow up

## Type of Change

New feature

## Testing

- unit tests
- more exhaustive manual test suite
- integ tests

- [x] I ran `npm run check`

## Checklist

----


src/models/openai-responses.ts
pgrayy, line 1: This should be implemented as part of OpenAIModel with the OpenAIModelOptions api field set as "responses" by default ([here](https://github.com/strands-agents/sdk-typescript/blob/main/src/models/openai.ts#L154)). You can still split up the logic across multiple modules though (see [google](https://github.com/strands-agents/sdk-typescript/tree/main/src/models/google) for an example).

pgrayy (changes requested):

strands-ts/src/agent/agent.ts
```diff
@@ -188,6 +189,12 @@ export class Agent implements LocalAgent, InvokableAgent {
    * State is not passed to the model during inference.
    */
   public readonly appState: StateStore
+  /**
+   * Runtime state for the model provider. Used by stateful models to persist
+   * provider-specific data (e.g., response IDs for conversation chaining)
+   * across invocations.
+   */
+  public modelState: Record<string, JSONValue>
```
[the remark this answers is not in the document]
notowen333: This pattern is an exact analog to the Python implementation both the name and the data structure.

src/models/openai-responses.ts
```diff
@@ -0,0 +1,606 @@
...
+
+  /**
+   * Formats the full request for the Responses API.
+   */
+  private _formatRequest(messages: Message[], options?: StreamOptions): Record<string, unknown> {
+    const input = this._formatMessages(messages)
+
+    const stateful = this.stateful
+    const request: Record<string, unknown> = {
+      ...(this._config.params ?? {}),
```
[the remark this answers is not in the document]
notowen333: Managed keys in params already log a warning at config time and get overridden by spread order, so the behavior is defined — silent filtering would actually make debugging harder since the override just disappears with no signal.

src/models/openai-responses.ts
```diff
@@ -0,0 +1,606 @@
...
+              }
+            }
+            const details = resp.incomplete_details as { reason?: string } | null
+            if (details?.reason === 'max_output_tokens') {
+              stopReason = 'maxTokens'
+            }
+            break
+          }
+
+          case 'response.completed': {
```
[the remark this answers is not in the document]
notowen333: This is overly defensive. The code already handles stream interruption; when we get a response we can expect clear values

src/models/openai-responses.ts
```diff
@@ -0,0 +1,606 @@
...
+          }
+
+          case 'toolUseBlock': {
+            const toolBlock = block as { name: string; toolUseId: string; input: unknown }
+            toolCallItems.push({
+              type: 'function_call',
+              call_id: toolBlock.toolUseId,
+              name: toolBlock.name,
+              arguments: JSON.stringify(toolBlock.input),
+            } as unknown as ResponseInputItem)
```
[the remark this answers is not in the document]
notowen333: added some narrowing where possible (and documented where not possible)

strands-ts/src/agent/agent.ts
```diff
@@ -263,10 +283,13 @@ export class Agent implements LocalAgent, InvokableAgent {
     this._hooksRegistry = new HookRegistryImplementation()
 
     // Initialize plugin registry with all plugins to be initialized during initialize()
+    // ModelPlugin is registered last so that on AfterInvocationEvent (which uses reverse
+    // callback ordering), it runs first — clearing messages before SessionManager saves.
     this._pluginRegistry = new PluginRegistry([
```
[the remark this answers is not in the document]
notowen333: added a test verifying execution order

strands-ts/src/agent/snapshot.ts
```diff
@@ -154,6 +158,17 @@ export function loadSnapshot(agent: LocalAgent, snapshot: Snapshot): void {
       delete agent.systemPrompt
     }
   }
+
+  if ('modelState' in snapshot.data) {
+    const newState = snapshot.data.modelState
+    // Clear existing keys while preserving the object reference
+    for (const key of Object.keys(agent.modelState)) {
```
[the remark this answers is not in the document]
notowen333: added a comment

src/models/model.ts
```diff
@@ -415,3 +442,29 @@ export abstract class Model<T extends BaseModelConfig = BaseModelConfig> {
...
 }
+
+/**
+ * Built-in plugin that manages model-related lifecycle hooks.
+ *
+ * When the model is stateful (server-managed conversation state), this plugin
+ * clears the agent's local message history after each invocation since the
+ * server holds the authoritative conversation state.
+ */
+export class ModelPlugin implements Plugin {
```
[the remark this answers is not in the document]
notowen333: In TypeScript, export is required for any cross-file import — it doesn't signal "public API." What's public is what's re-exported from src/index.ts, and ModelPlugin isn't there.                                                                                                                                                              
                                                                                                                                                                                                    Removing export would break the import in agent.ts.

Also added a comment in `ModelPlugin` explaining

strands-ts/test/integ/models/openai/responses.test.ts
```diff
@@ -0,0 +1,295 @@
...
+
+import { collectIterator } from '$/sdk/__fixtures__/model-test-helpers.js'
+
+import { openaiResponses } from '../__fixtures__/model-providers.js'
+
+describe.skipIf(openaiResponses.skip)('OpenAIResponsesModel Integration Tests', () => {
+  describe('Configuration', () => {
+    it.concurrent('respects maxTokens configuration', async () => {
+      const provider = openaiResponses.createModel({
+        modelId: 'gpt-5.4-mini',
```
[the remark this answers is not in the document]
notowen333: We hard code model ids across unit tests and this is a real model

src/models/openai-responses.ts
notowen333: updated in revision!

strands-ts/src/models/openai/model.ts
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  /**
+   * Whether this model manages conversation state server-side.
+   *
+   * `true` only for `api: 'responses'` with `stateful !== false`. Chat Completions
+   * is always stateless.
+   */
+  override get stateful(): boolean {
+    return this._api === 'responses' && this._config.stateful !== false
```
pgrayy: Nit: stateful is a boolean so you can say `this._config.stateful`.

strands-ts/src/models/openai/index.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * OpenAI model provider.
+ *
+ * Supports both the Chat Completions API (default) and the Responses API
```
pgrayy: We want to make responses the default since it is the latest api. This is why currently we require api be passed in. We wanted this option to later default to responses.

strands-ts/src/models/openai/index.ts
```diff
@@ -0,0 +1,26 @@
...
+ * (with server-managed conversation state).
```
pgrayy: Nit: Some of this documentation seems to be suggesting that responses API will always operate with a server-managed conversation. That has to be enabled though. With that said, I wouldn't worry about explicitly calling it out in all places since it is just one of many features of the responses API.

strands-ts/src/models/openai/model.ts
```diff
@@ -0,0 +1,245 @@
+/**
+ * OpenAI model provider implementation.
+ *
+ * Supports both the Chat Completions API (stateless) and the Responses API
+ * (stateful, with server-managed conversation state). Selected via the `api`
```
pgrayy: Nit: Reiterating https://github.com/strands-agents/sdk-typescript/pull/820/changes#r3138958378. This is another example where I don't think we need to explicitly call out stateless vs stateful. Stateful mode is one of many features of the responses API and will be disabled by default.

strands-ts/src/models/openai/model.ts
```diff
@@ -0,0 +1,245 @@
...
+ * ```
+ */
+export class OpenAIModel extends Model<OpenAIModelConfig> {
+  private readonly _api: OpenAIApi
+  private _config: OpenAIModelConfig
+  private _client: OpenAI
+
+  constructor(options: OpenAIModelOptions) {
+    super()
+    const { apiKey, client, clientConfig, api = 'chat', ...modelConfig } = options
```
pgrayy: As noted above, we want responses as the default.

strands-ts/src/agent/snapshot.ts
```diff
@@ -104,6 +104,10 @@ export function takeSnapshot(agent: LocalAgent, options: TakeSnapshotOptions): S
     data.systemPrompt = agent.systemPrompt !== undefined ? (systemPromptToData(agent.systemPrompt) as JSONValue) : null
   }
 
+  if (fields.has('modelState')) {
+    data.modelState = JSON.parse(JSON.stringify(agent.modelState)) as JSONValue
```
pgrayy: Question: Is the stringify to parse necessary? Could we just say agent.modelState as JSONValue?

strands-ts/src/models/openai/responses-adapter.ts
```diff
@@ -0,0 +1,522 @@
...
+import type { ImageBlock, DocumentBlock } from '../../types/media.js'
+import { encodeBase64 } from '../../types/media.js'
+import { toMimeType } from '../../mime.js'
+import type { ModelStreamEvent } from '../streaming.js'
+import type { StreamOptions } from '../model.js'
+import { logger } from '../../logging/logger.js'
+import { formatImageDataUrl } from './formatting.js'
+import type { OpenAIResponsesConfig } from './types.js'
+
+export const DEFAULT_RESPONSES_MODEL_ID = 'gpt-4o'
```
pgrayy: Should this match the chat completions model? Should we do gpt-5.4?

strands-ts/src/models/openai/responses-adapter.ts
```diff
@@ -0,0 +1,522 @@
...
+            type: 'function_call_output',
+            call_id: resultBlock.toolUseId,
+            output,
+          }
+          toolResultItems.push(result)
+          break
+        }
+
+        case 'reasoningBlock': {
+          logger.warn('block_type=<reasoningBlock> | reasoning blocks cannot be re-submitted to responses api')
```
pgrayy: Is this correct? I assumed in Python we omit reason blocks simply because we didn't get around to formatting them. We can handle in a follow up but curious if it is actually the case openai responses doesn't accept them.

strands-ts/src/models/openai/responses-adapter.ts
```diff
@@ -0,0 +1,522 @@
...
+          parts.push('[JSON serialization error]')
+        }
+        break
+      }
+      case 'imageBlock':
+        parts.push('[image content]')
+        break
+      case 'documentBlock':
+        parts.push('[document content]')
+        break
```
pgrayy: We should be sending the real image and document data ([python reference](https://github.com/strands-agents/sdk-python/blob/main/src/strands/models/openai_responses.py#L662)).

strands-ts/src/plugins/model-plugin.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  constructor(model: Model) {
+    this._model = model
+  }
+
+  initAgent(agent: LocalAgent): void {
+    const model = this._model
+    agent.addHook(AfterInvocationEvent, () => {
+      if (model.stateful) {
+        agent.messages.length = 0
```
pgrayy, line 29: Consideration: Not sure we need to enforce this but I wonder if it should matter whether we clear the messages before or after session manager snapshotting. Python we don't worry about this because we session manage messages one at a time as MessageAddedEvents are emitted. TS we snapshot all state as a whole. Again, not sure it really matters though. Something we can follow up on if need be.

strands-ts/src/agent/agent.ts
```diff
@@ -263,10 +284,13 @@ export class Agent implements LocalAgent, InvokableAgent {
     this._hooksRegistry = new HookRegistryImplementation()
 
     // Initialize plugin registry with all plugins to be initialized during initialize()
+    // ModelPlugin is registered last so that on AfterInvocationEvent (which uses reverse
+    // callback ordering), it runs first — clearing messages before SessionManager saves.
```
pgrayy, line 316: Is session manager guaranteed to always run on AfterInvocationEvent? If not, it could end up saving messages on an earlier event. Consequently, when loading from session manager, we could end up with a non-empty messages array. We explicitly accounted for this in Python with the logic [here](https://github.com/strands-agents/sdk-python/blob/main/src/strands/session/repository_session_manager.py#L227). Might need something similar here in TS.

strands-ts/src/agent/snapshot.ts
notowen333: Switched to a shallow spread (\`{ ...agent.modelState }\`) in this revision. A plain \`as JSONValue\` would share nested refs — post-snapshot mutations to \`agent.modelState\` would leak into the snapshot.

strands-ts/src/models/openai/responses-adapter.ts
notowen333: Aligned with python

Reworded the warning to match the actual status: "reasoning content is not yet supported in multi-turn conversations with the responses api."

strands-ts/src/models/openai/responses-adapter.ts
notowen333: Done — \`formatToolResultOutput\` now encodes image/document bytes into \`input_image\`/\`input_file\` content parts and returns the array shape when any media is present. Text-only results still collapse to a string.

strands-ts/src/agent/agent.ts
notowen333: The cleanest way to achieve this would be to have a `model` property on `LocalAgent` Because this was brought up before and explicitly not chosen:

Added a guard in \`Agent.initialize()\` that clears \`messages\` post-\`InitializedEvent\` dispatch when \`model.stateful\` is true. Placing it there (rather than inside \`loadSnapshot\`) keeps it ordering-independent from any specific plugin.

strands-ts/src/plugins/model-plugin.ts
notowen333: no op for now

strands-ts/src/models/openai/model.ts
```diff
@@ -0,0 +1,260 @@
...
+ * ```
+ */
+export class OpenAIModel extends Model<OpenAIModelConfig> {
+  private readonly _api: OpenAIApi
+  private _config: OpenAIModelConfig
+  private _client: OpenAI
+
+  constructor(options: OpenAIModelOptions) {
+    super()
+    const { apiKey, client, clientConfig, api = 'responses', ...modelConfig } = options
```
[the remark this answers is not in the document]
notowen333: Intentional: this PR makes \`responses\` the default per an earlier review decision (https://github.com/strands-agents/sdk-typescript/pull/820#discussion_r3138951888). Called out in the PR description.

strands-ts/src/agent/snapshot.ts
```diff
@@ -104,6 +104,11 @@ export function takeSnapshot(agent: LocalAgent, options: TakeSnapshotOptions): S
     data.systemPrompt = agent.systemPrompt !== undefined ? (systemPromptToData(agent.systemPrompt) as JSONValue) : null
   }
 
+  if (fields.has('modelState')) {
+    // Shallow copy so later mutations to agent.modelState don't bleed into the snapshot.
+    data.modelState = { ...agent.modelState }
```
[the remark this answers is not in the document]
notowen333: Expanded the comment to note the assumption (primitives only today, upgrade to deep clone if a provider starts writing nested objects).

strands-ts/src/agent/agent.ts
```diff
@@ -331,6 +355,15 @@ export class Agent implements LocalAgent, InvokableAgent {
 
     await this._hooksRegistry.invokeCallbacks(new InitializedEvent({ agent: this }))
 
+    // For stateful models the server owns conversation history, so any messages
+    // a plugin (typically SessionManager) just restored must be discarded —
+    // otherwise the next invocation would re-send them alongside the server's
+    // state. If `model` ever moves onto the `LocalAgent` interface, the cleaner
+    // home for this guard is inside `loadSnapshot` itself.
+    if (this.model.stateful) {
```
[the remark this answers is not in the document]
notowen333: Good nudge. Moved the guard out of \`Agent.initialize()\` and into \`SessionManager._onAgentInitialized\` so it only fires when session restore is the source of messages — caller-seeded or plugin-seeded messages now survive. Included a debug log noting discards.

strands-ts/src/models/openai/responses-adapter.ts
```diff
@@ -0,0 +1,546 @@
...
+    case 'response.incomplete': {
+      const resp = event.response
+      if (resp.usage) {
+        state.finalUsage = {
+          inputTokens: resp.usage.input_tokens,
+          outputTokens: resp.usage.output_tokens,
+          totalTokens: resp.usage.total_tokens,
+        }
+      }
+      const details = resp.incomplete_details as { reason?: string } | null
```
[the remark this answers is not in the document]
notowen333: Done — cast removed, using the SDK-typed \`incomplete_details.reason\` directly.

package-lock.json
```diff
@@ -1,10 +1,12 @@
 {
   "name": "strands",
+  "version": "0.0.0",
```
pgrayy, line 3: Curious how this happened? Maybe we can remove these changes. Generally speaking, we should avoid posting changes to package-lock.json unless we explicitly add new dependencies to package.json (can be avoided using `npm ci` instead of `npm install`). This way we avoid constant merge conflicts across PRs.

strands-ts/src/types/agent.ts
```diff
@@ -151,6 +152,13 @@ export interface LocalAgent {
    */
   messages: Message[]
 
+  /**
+   * Runtime state for the model provider. Used by stateful models to persist
+   * provider-specific data (e.g., response IDs for server-side conversation chaining)
+   * across invocations.
+   */
+  modelState: Record<string, JSONValue>
```
pgrayy: * https://github.com/strands-agents/sdk-typescript/pull/784/changes#diff-257be9b4ec7364c9d818bcbfdcfe9bfef93a9d5626f3963a75f4cad912b6de2dR350
* https://github.com/strands-agents/sdk-typescript/pull/784/changes#diff-b72b80a5641bc634b4d3605cd565dab66e02ac63891c0c2f36851f2858304251R109

^ I think we should follow a pattern like this to avoid exposing modelState as a public attribute on LocalAgent (at least to start). I think this will also help to better indicate that modelState should only be modified in the model providers, not in hook callbacks. This would also match the expectations set in Python.

strands-ts/src/session/session-manager.ts
```diff
@@ -199,6 +199,18 @@ export class SessionManager implements Plugin, MultiAgentPlugin {
         `agent_id=<${event.agent.id}>, session_id=<${this._sessionId}> | agent had existing messages that were overwritten by session restore`
       )
     }
+
+    // Stateful models manage conversation history server-side, so any messages
+    // loaded from the snapshot would drift from the server's view on the next
+    // invocation. Duck-type the agent's `model` since `LocalAgent` does not
+    // expose it — `Agent` is the only implementor and always has one.
+    const statefulModel = (event.agent as { model?: { stateful?: boolean } }).model?.stateful === true
```
pgrayy: Nit: Could say `.stateful` instead of `.stateful === true` since it is a boolean field.

strands-ts/src/models/openai/types.ts
```diff
@@ -0,0 +1,142 @@
...
+   */
+  presencePenalty?: number
+}
+
+/**
+ * Configuration fields specific to the Responses API.
+ */
+export interface OpenAIResponsesConfig extends OpenAIBaseConfig {
+  /**
+   * When `true` (default for `api: 'responses'`), the server manages conversation
```
pgrayy: We should make this false (stateless) by default. This is what we do in Python.

strands-ts/src/models/openai/responses-adapter.ts
```diff
@@ -0,0 +1,545 @@
...
+    }
+  }
+
+  if (hasMedia) return parts
+
+  // Text-only: collapse to a single string to match the API's simpler shape.
+  const text = parts
+    .filter((p): p is { type: 'input_text'; text: string } => p.type === 'input_text')
+    .map((p) => p.text)
+    .join('\n')
```
pgrayy: Is there an advantage to doing this?

strands-ts/src/agent/agent.ts
```diff
@@ -188,6 +190,12 @@ export class Agent implements LocalAgent, InvokableAgent {
    * State is not passed to the model during inference.
    */
   public readonly appState: StateStore
+  /**
+   * Runtime state for the model provider. Used by stateful models to persist
+   * provider-specific data (e.g., response IDs for conversation chaining)
+   * across invocations.
+   */
+  public modelState: Record<string, JSONValue>
```
pgrayy: For consistency, I'm wondering if we should just make this a StateStore. What do you think?

strands-ts/src/models/openai/__tests__/responses.test.ts
pgrayy, line 1: Nit: Consider moving all openai unit tests under openai directory (similar to google tests). Then you could call this responses.tests.ts. Could be considered for follow up. Don't want to block on this.

strands-ts/test/integ/models/openai/responses.test.ts
pgrayy, line 1: Nit: Similar to the unit tests, would recommend creating an openai directory for further organization.

strands-ts/src/types/agent.ts
notowen333: discussed offline, prefer not to introduce cast pattern in a bunch of places/there are legitimate use cases for plugins

strands-ts/src/agent/agent.ts
notowen333: yes on board with this one!

package-lock.json
notowen333: the top level package had a version added to it: https://github.com/strands-agents/sdk-typescript/pull/875

strands-ts/src/session/session-manager.ts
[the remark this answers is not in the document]
notowen333: we won't be adding a subfield of model to LocalAgent

strands-ts/src/session/session-manager.ts
```diff
@@ -199,6 +199,18 @@ export class SessionManager implements Plugin, MultiAgentPlugin {
         `agent_id=<${event.agent.id}>, session_id=<${this._sessionId}> | agent had existing messages that were overwritten by session restore`
       )
     }
+
+    // Stateful models manage conversation history server-side, so any messages
+    // loaded from the snapshot would drift from the server's view on the next
+    // invocation. Duck-type the agent's `model` since `LocalAgent` does not
+    // expose it — `Agent` is the only implementor and always has one.
+    const statefulModel = (event.agent as { model?: { stateful?: boolean } }).model?.stateful
```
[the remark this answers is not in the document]
notowen333: this was already addressed we explicitly prefer the cast

pgrayy (approved):

Issue #823, opened 2026-04-20
[FEATURE] Default model settings tuning and centralization


notowen333:
### Problem Statement

Decide on clear model defaults (ID, max_tokens, etc) and centralize them in a single file.

Also emit a clear log re-iterating defaults when they are used.

### Use Case

Maintainer team can more easily keep track of and update default settings.

PR #824, opened 2026-04-20
fix(anthropic): update maxTokens default value, remove dead default and haiku fallback

opieter-aws:
## Description                                                                                                                                                                                          
                                                                                                                                                                                                      
The constructor in `AnthropicModel` set `maxTokens: 4096` as a default, which meant `this._config.maxTokens` was never `undefined`. This made the model-aware fallback in `_formatRequest`              
(`this._config.maxTokens ?? (modelId.includes('haiku-3') ? 4096 : 32768)`) dead code — the `??` branch never triggered. Every model silently defaulted to 4096 tokens regardless of capability.         
                                                                                                                                                                                                      
This PR:                                                                                                                                                                                                
- Makes `maxTokens` a required field in `AnthropicModelConfig`, aligning with the Python SDK where `max_tokens: Required[int]` has no default                                                           
- Removes the `maxTokens: 4096` constructor default                                                                                                                                                     
- Removes the dead haiku-3 fallback logic in `_formatRequest`                                                                                                                                           
- Updates all test call sites to pass `maxTokens` explicitly                                                                                                                                            
                                                                                                                                                                                                      
## Related Issues                                                                                                                                                                                       
                                                                                                                                                                                                      
https://github.com/strands-agents/sdk-typescript/issues/823                                                                                                                                                                                                 
                                                                                                                                                                                                      
## Documentation PR                                                                                                                                                                                     
                                                                                                                                                                                                      
N/A — API surface change; callers must now provide `maxTokens` explicitly.                                                                                                                              
                                                                                                                                                                                                      
## Type of Change                                                                                                                                                                                       
                                                                                                                                                                                                      
Breaking change                                                                                                                                                                                         
                                                                                                                                                                                                      
## Testing                                                                                                                                                                                              
                                                                                                                                                                                                      
- [x] I ran `npm run check`                                                                                                                                                                             
- Verified all existing tests pass after updating call sites to supply `maxTokens`                                                                                                                      
                                                                                                                                                                                                      
## Checklist                                                                                                                                                                                            
                                                                                                                                                                                                      
----                                                                                                                                                                                                    
                                                                                                                                                                                                      
### DIFF
--- a/strands-ts/src/models/anthropic.ts
+++ b/strands-ts/src/models/anthropic.ts
@@ -9,10 +9,17 @@ import { encodeBase64 } from '../types/media.js'
 import { logger } from '../logging/logger.js'
 
 const DEFAULT_ANTHROPIC_MODEL_ID = 'claude-sonnet-4-6'
+const DEFAULT_ANTHROPIC_MAX_TOKENS = 64_000
 const CONTEXT_WINDOW_OVERFLOW_ERRORS = ['prompt is too long', 'max_tokens exceeded', 'input too long']
 const TEXT_FILE_FORMATS = ['txt', 'md', 'markdown', 'csv', 'json', 'xml', 'html', 'yml', 'yaml', 'js', 'ts', 'py']
 
 export interface AnthropicModelConfig extends BaseModelConfig {
+  /**
+   * Maximum number of tokens the model can generate in a response.
+   *
+   * @defaultValue 64000 — subject to change between versions.
+   * Set this explicitly to avoid unexpected changes.
+   */
   maxTokens?: number
   stopSequences?: string[]
   params?: Record<string, unknown>
@@ -34,8 +41,14 @@ export class AnthropicModel extends Model<AnthropicModelConfig> {
 
     this._config = {
       modelId: DEFAULT_ANTHROPIC_MODEL_ID,
-      maxTokens: 4096,
+      maxTokens: DEFAULT_ANTHROPIC_MAX_TOKENS,
       ...modelConfig,
+    }
+
+    if (modelConfig.maxTokens === undefined) {
+      logger.warn(
+        `max_tokens=<${DEFAULT_ANTHROPIC_MAX_TOKENS}> | using default maxTokens, which is subject to change | set maxTokens explicitly to pin the value`
+      )
     }
 
     if (client) {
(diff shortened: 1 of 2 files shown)


strands-ts/src/models/anthropic.ts
```diff
@@ -211,12 +211,9 @@ export class AnthropicModel extends Model<AnthropicModelConfig> {
   private _formatRequest(messages: Message[], options?: StreamOptions): Anthropic.MessageStreamParams {
     if (!this._config.modelId) throw new Error('Model ID is required')
 
-    // Set max_tokens based on model: Haiku 3 supports 4096, others support up to 32k
-    const maxTokens = this._config.maxTokens ?? (this._config.modelId.includes('haiku-3') ? 4096 : 32768)
```
mkmeral, line 215: can we just default to 64k? https://platform.claude.com/docs/en/about-claude/models/overview

mkmeral: that is much simpler, and better devx

opieter-aws: I discussed with @zastrowm to remove the the default for now to pair with Python SDK. We can then add a default later, but we remove the risk of breaking change for TS 1.0. 

@zastrowm what are your thoughts?

zastrowm: Yeah; that's my summary; two-way-door to add a default later whereas updating the default later would be a breaking change.

If 64k is going to be enough for customer until the default changes, then I think it makes sense.  But we wanted to avoid the back & forth discussion of "How do we configure this for N models", which this accomplishes.

mkmeral: > If 64k is going to be enough for customer until the default changes, then I think it makes sense

I think 64k is a sensible default that covers most of the models as of today

Since we have said default model ids are subject to change over time, this param would be too. We should document this though

opieter-aws: Alright I updated to:
1. use 64k as default
2. Like the default model ID (https://github.com/strands-agents/sdk-python/pull/2164), log a warning

strands-ts/src/models/anthropic.ts
```diff
@@ -211,12 +225,9 @@ export class AnthropicModel extends Model<AnthropicModelConfig> {
...
-
     const request: Anthropic.MessageStreamParams = {
       model: this._config.modelId,
-      max_tokens: maxTokens,
+      max_tokens: this._config.maxTokens ?? DEFAULT_ANTHROPIC_MAX_TOKENS,
```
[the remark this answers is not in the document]
opieter-aws: The `??` is intentional. `maxTokens` is typed as `number | undefined` in `AnthropicModelConfig`, so TypeScript requires either a non-null assertion (`!`) or a fallback (`??`) here. We chose `??`  over `!` as it's defensive without being misleading.

strands-ts/src/models/anthropic.ts
```diff
@@ -211,12 +224,9 @@ export class AnthropicModel extends Model<AnthropicModelConfig> {
...
```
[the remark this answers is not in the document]
opieter-aws: The `??` is intentional. `maxTokens` is typed as `number | undefined` in `AnthropicModelConfig`, so TypeScript requires either a non-null assertion (!) or a fallback (??) here. I chose `??` over `!` as it's defensive without being misleading.

mehtarac (approved):

PR #853, opened 2026-04-21
feat: add countTokens() heuristic to Model base class

opieter-aws:
## Description

Adds `countTokens()` to the `Model` base class for estimating input token count before sending to the model, enabling proactive context management (e.g., triggering compression at a threshold).

Uses a character-based heuristic (`chars/4` for text, `chars/2` for JSON) as a universal fallback for all providers. Individual providers can override with native counting APIs in a follow-up PR.

Handles all content block types: text, toolUse, toolResult, reasoningContent, guardContent, citationsContent, system prompts (string and array), and tool specs. Non-serializable content (e.g., image bytes) is gracefully skipped.

### TypeScript port of
- Python PR: strands-agents/sdk-python#2031

### Follow-up
- Native provider overrides (Bedrock `CountTokens`, Anthropic `countTokens`, Google `countTokens`) tracked separately — branch `feat/token-estimation-all` on fork has the implementation ready.
- OpenAI Chat Completions has no native counting endpoint (only the Responses API does); documented in JSDoc.

## Type of Change

New feature

## Testing

- 12 new unit tests covering all content block types, system prompts, tool specs, edge cases
- All 2027 existing tests continue to pass
- `npm run build`, `lint`, `format:check`, `type-check` all pass

## Checklist
- [x] I have added tests that prove my fix is effective or my feature works
- [x] Existing tests continue to pass

### DIFF
--- a/strands-ts/src/models/model.ts
+++ b/strands-ts/src/models/model.ts
@@ -123,6 +123,22 @@ export interface StreamOptions {
 }
 
 /**
+ * Options for counting tokens in a set of messages.
+ */
+export interface CountTokensOptions {
+  /**
+   * System prompt to guide the model's behavior.
+   * Can be a simple string or an array of content blocks for advanced caching.
+   */
+  systemPrompt?: SystemPrompt
+
+  /**
+   * Array of tool specifications to include in the count.
+   */
+  toolSpecs?: ToolSpec[]
+}
+
+/**
  * Result interface for the streamAggregated method.
  * Contains the complete message, stop reason, and optional metadata.
  */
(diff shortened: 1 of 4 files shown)


strands-ts/src/models/model.ts
```diff
@@ -429,3 +447,92 @@ export abstract class Model<T extends BaseModelConfig = BaseModelConfig> {
...
+      total += heuristicJson(block.input)
+      break
+    case 'toolResultBlock':
+      for (const item of block.content) {
+        if (item.type === 'textBlock') {
+          total += heuristicText(item.text)
+        } else if (item.type === 'jsonBlock') {
+          total += heuristicJson(item.json)
+        }
+      }
```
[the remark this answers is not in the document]
opieter-aws: Intentionally skipped, consistent with the Python SDK implementation. `DocumentBlock` contains binary data (bytes/S3 refs). The text source variant is an edge case we can add later if needed.

opieter-aws:
/strands can you change the `estimateTokens` method to public?

strands-ts/src/models/model.ts
```diff
@@ -191,6 +207,24 @@ export abstract class Model<T extends BaseModelConfig = BaseModelConfig> {
...
+   *
+   * Subclasses should override this method to use native token counting APIs
+   * (e.g., Bedrock CountTokens, Anthropic countTokens, Gemini countTokens)
+   * for improved accuracy, falling back to `super.countTokens()` on API failure.
+   *
+   * @param messages - Array of conversation messages to count tokens for
+   * @param options - Optional options containing system prompt and tool specs
+   * @returns Total input token count
+   */
+  async countTokens(messages: Message[], options?: CountTokensOptions): Promise<number> {
```
lizradway, line 224: per our offline discussion, can this be non-message specific so that we can tokenize tool content, etc & make this generally accessible throughout the sdk.

lizradway: ignore me! didnt read

lizradway: will models be able to override the non-message specific tokenizer method (ie. `estimateContentBlockTokens`) as well? 

-----------------

im thinking out loud a little here, but if we are going to [use tokens as the unit of strands](https://github.com/strands-agents/docs/pull/777), then `count_tokens` is an influential unit of the sdk. 

this means that if we are not using the model override for non-message specific blocks, then `tiktoken`/heuristic is also required on the sdk. that to me just suggests that maybe we should make it clear that the user must install `tiktoken` on their own if they want better estimation (raising it as low level warning or maybe just making it VERY clear in the context management best practices docs page) and that this has far reaching impact.


more of a note for later and questions for myself to ponder than a comment.

opieter-aws: See my [draft PR](https://github.com/strands-agents/sdk-python/pull/2189), the entire `count_tokens` method is overridden. So there is no `contentBlock`-specific estimation in there, we leave the implementation details to the provider's API

lizradway (approved):

PR #854, opened 2026-04-21
feat: add concurrent tool execution

### LINKED ISSUE #817: [FEATURE] Add concurrent tool execution
### Problem

When the model returns multiple `tool_use` blocks in a single response (indicating it wants to call multiple tools simultaneously), the TypeScript SDK executes them **sequentially** via a `for` loop in `Agent.executeTools()`:

```typescript
// src/agent/agent.ts — executeTools method
for (const toolUseBlock of toolUseBlocks) {
    const toolResultBlock = yield* this.executeTool(toolUseBlock, toolRegistry)
    toolResultBlocks.push(toolResultBlock)
    yield new ToolResultEvent({ agent: this, result: toolResultBlock })
}
```

This negates the model's intent to parallelize independent operations. For example, if the model decides to create a character AND a location simultaneously (two independent API calls), they run one after the other instead of concurrently.

### Evidence

The model **does** return multiple `tool_use` blocks in a single response:

```
🧠 Agent Decision {
  stopReason: "toolUse",
  reasoning: "I'll create Bilu as a MAIN character and Tillu as a PRIMARY EXTERIOR location in parallel.",
  toolCalls: [
    { tool: "manage_characters", input: { action: "create", ... } },
    { tool: "manage_locations", input: { action: "create", ... } }
  ]
}
```

### Context

- The **Python SDK already supports this** via `ConcurrentToolExecutor` (which is the default executor)
- Claude's documentation states the model decides parallel tool use based on operation independence
- Bedrock Converse API correctly passes through multiple `tool_use` blocks
- Both `executeTools` and `executeTool` are `private`, so consumers cannot override them cleanly

### Use Case

We're building an AI copilot for a creative studio application (React/TypeScript, Bedrock Converse API). Users give natural language commands like:
- "Create a character called Cherry and a location called Cherry Farm" → 2 independent API calls
- "Generate images for character A and character B" → 2 independent generation jobs
- "List all characters and all locations" → 2 independent read calls

The SDK runs them sequentially, making the copilot feel sluggish — for generation jobs that take 5–10 seconds each, this turns a 10-second parallel wait into a 20-second sequential wait.

---

## Implementation Requirements

Based on discussion with the maintainer and analysis of the Python SDK's `ConcurrentToolExecutor`.

### 1. New `toolExecutor` Config Option

Add a `toolExecutor` property to `AgentConfig`:

```typescript
// In AgentConfig
toolExecutor?: 'sequential' | 'concurrent'
```

- **Default**: `'sequential'` (backward compatible with current behavior)
- **Naming**: `toolExecutor` (camelCase, TypeScript convention)
- **Type**: String union, not a class instance. A class-based extensibility pattern can be introduced later if needed.

```typescript
const agent = new Agent({
    model,
    tools,
    toolExecutor: 'concurrent',
})
```

### 2. Sequential Execution (Current Behavior — No Change)

When `toolExecutor` is `'sequential'` (or omitted), execution remains as-is: tools run one at a time in a `for` loop. This is the current `executeTools` behavior.

### 3. Concurrent Execution

When `toolExecutor` is `'concurrent'`, `executeTools` must run all tool calls concurrently while maintaining the streaming event contract.

#### Event Serialization Requirement

Tool executions run concurrently, but streamed `AgentStreamEvent` objects must still be yielded **one at a time** from the `executeTools` generator. Since `executeTool` is an async generator that yields events, and you cannot `yield` from multiple generators simultaneously, the implementation must bridge concurrent execution with the serial generator contract.

The Python SDK solves this with a queue-based pattern (`ConcurrentToolExecutor`), but any approach that preserves the serial yield contract is acceptable.

### 4. Hook Event Contract

All existing hook events must fire correctly in both modes:

| Event | Behavior |
|---|---|
| `BeforeToolsEvent` | Fires once before **all** tool executions begin (wraps the batch) |
| `BeforeToolCallEvent` | Fires once per tool, before that tool executes |
| `AfterToolCallEvent` | Fires once per tool, after that tool completes |
| `ToolResultEvent` | Fires once per tool with its result |
| `ToolStreamUpdateEvent` | Fires for each streaming event from a tool |
| `AfterToolsEvent` | Fires once after **all** tool executions complete (wraps the batch) |

In concurrent mode, the interleaving order of per-tool events across tools is non-deterministic, but events for a single tool must remain in order (Before → Stream → After → Result).

### 5. Per-Tool Retry

The `AfterToolCallEvent.retry` flag must work independently per tool in concurrent mode:

- Each tool task runs its own retry loop (matching the existing `executeTool` `while(true)` pattern)
- If Tool A's `AfterToolCallEvent.retry = true`, only Tool A re-executes; Tools B and C are unaffected
- This naturally falls out of running each tool in its own isolated task

### 6. Cancellation

Cancellation must work in concurrent mode following the same pattern as the Python SDK:

- **Before tools launch**: If `isCancelled` is true before launching concurrent tasks, create cancel results for all tools (same as current sequential behavior)
- **During concurrent execution**: In-progress tools run to completion unless they check `cancelSignal` internally. Tools that haven't started yielding results yet should check `isCancelled` and produce cancel results. This matches the cooperative cancellation model already documented in the SDK.
- **`BeforeToolsEvent.cancel`**: If the hook sets cancel, all tools get cancel results (same as current behavior)

### 7. Structured Output Compatibility

The structured output flow (`_extractStructuredOutput`) inspects tool results after `executeTools` completes. This must continue to work unchanged — the `toolResultMessage` returned by `executeTools` must contain all tool results regardless of execution mode.

### Files to Modify

- **`src/agent/agent.ts`** — `AgentConfig` type (add `toolExecutor`), constructor (store config), `executeTools` method (branch on mode)
- **`src/agent/__tests__/agent.test.ts`** — Tests for concurrent execution, event ordering, retry, cancellation
- **`src/index.ts`** — Export new types if any new public types are introduced

### Acceptance Criteria

- [ ] `AgentConfig` accepts `toolExecutor?: 'sequential' | 'concurrent'`
- [ ] Default behavior (`undefined` / `'sequential'`) is identical to current behavior (no regression)
- [ ] When `'concurrent'`, multiple tool calls from a single model response execute concurrently
- [ ] All hook events fire correctly in concurrent mode (Before/After pairs, per-tool events)
- [ ] Per-tool events remain ordered per tool (BeforeToolCall → ToolStreamUpdate* → AfterToolCall → ToolResult)
- [ ] `AfterToolCallEvent.retry` works independently per tool in concurrent mode
- [ ] Cancellation works correctly in concurrent mode (cooperative model)
- [ ] `BeforeToolsEvent.cancel` cancels all tools in concurrent mode
- [ ] Structured output extraction works correctly in concurrent mode
- [ ] Telemetry (tracing and metrics) works correctly for concurrent tool calls
- [ ] Unit tests cover sequential regression, concurrent execution, event ordering, retry, and cancellation
- [ ] All existing tests pass without modification (backward compatibility)
- [ ] 80%+ test coverage maintained
- [ ] TSDoc on all new exported types/functions

### Assumptions

- The string union (`'sequential' | 'concurrent'`) is sufficient for now. A class-based `ToolExecutor` interface can be introduced later for custom executors.
- No changes to CI/CD workflows are needed.
- `toolExecutor` is a static config per Agent instance — it does not change between invocations.

### Notes for Future Improvement

- **Default change**: The Python SDK defaults to concurrent. The TypeScript SDK may switch to concurrent as the default in a future release, pending evaluation at PR time.
- **Class-based API**: If users need custom execution strategies, a `ToolExecutor` interface pattern (matching Python) can be introduced as a non-breaking extension.
- Consider updating `AGENTS.md` and `README.md` to document the `toolExecutor` config option once implemented.

notowen333:
## Description

Adds an opt-in concurrent tool executor so that when a model emits multiple
`toolUse` blocks in a single assistant turn, the agent can run them in parallel
instead of serially. Today, two independent 5-second tool calls take 10 seconds;
with `toolExecutor: 'concurrent'` they take ~5.

### Motivation

Tool calls from a single assistant turn are independent by construction — the
model asked for N things before seeing any of their results. Running them
sequentially turns latency-bound tool batches (HTTP fetches, LLM-as-tool calls)
into an artificial bottleneck. The Python SDK already supports this via
`ConcurrentToolExecutor`; this brings TypeScript to parity.

### Public API Changes

New `AgentConfig.toolExecutor` field and exported `ToolExecutor` type:

```typescript
import { Agent, type ToolExecutor } from '@strands-agents/sdk'

// Default: historical behavior, one tool at a time.
const sequential = new Agent({
  tools: [fetchUser, fetchOrders],
  // toolExecutor: 'sequential',  // implicit
})

// Opt in to parallelism for independent tool calls.
const concurrent = new Agent({
  tools: [fetchUser, fetchOrders],
  toolExecutor: 'concurrent',
})
```

`ToolExecutor` is a string union (`'sequential' | 'concurrent'`). Class-based
executors can come later if needed; keeping the surface small avoids
over-committing.

Event contract is preserved in both modes:
- Per-tool event order (`BeforeToolCallEvent` → `ToolStreamUpdateEvent*` →
  `AfterToolCallEvent` → `ToolResultEvent`) is always preserved.
- In concurrent mode, events from different tools may interleave.
- `BeforeToolsEvent` / `AfterToolsEvent` continue to bracket the batch, and
  `AfterToolsEvent.message` always contains results in the original
  tool-use-block source order regardless of completion order.

Cancellation semantics differ slightly between modes — worth calling out:
- Pre-launch cancel (hook sets `BeforeToolsEvent.cancel`, or `agent.cancel()`
  fires before tools start) produces error results for every tool in the batch
  in both modes.
- Mid-flight `agent.cancel()` in sequential mode short-circuits not-yet-started
  tools. In concurrent mode all N tools have already launched, so prompt
  batch-wide cancellation requires each in-flight tool to cooperatively observe
  `agent.cancelSignal`. Tools that ignore the signal run to completion — same
  contract as sequential mode today, just more visible.

No behavior change for existing users: the default remains `'sequential'`.

### Implementation note: `Promise.race` merge vs. Python's queue

The Python SDK's `ConcurrentToolExecutor` fans each tool out onto its own
`asyncio.Task`, which pushes events onto a shared `asyncio.Queue` that the
agent loop drains. That pattern fits Python well because tasks and queues are
cheap and `asyncio` has first-class cancellation.

TypeScript doesn't have a direct analogue, and porting the queue shape would
mean either (a) introducing a bounded async queue utility and buffer-management
code, or (b) dropping back to a single-consumer pattern anyway once events
cross back into the agent generator. Neither earns its complexity here.

Instead, the concurrent executor merges the N per-tool async generators
directly: it keeps a `Map<index, Promise<Step>>` of in-flight `.next()` calls,
`Promise.race`s them, yields the winner's event, and replaces that entry with
its next `.next()`. A few properties fall out naturally:

- **Per-tool ordering is free.** Each generator is still iterated strictly
  serially — the next `.next()` for generator `i` is only scheduled after the
  previous one resolved and its value was yielded. No ordering primitive
  required.
- **Retry stays per-tool.** `executeTool`'s own `while(true)` retry loop is
  unchanged. A retry on generator A re-enters its own loop without touching
  generator B.
- **No buffering.** Events flow straight through the race; there's no queue to
  size, drain, or close.
- **Failure isolation.** Each `.next()` is wrapped into a non-rejecting
  discriminated union (`{ kind: 'next' | 'throw' }`) so a misbehaving generator
  can't reject the whole `Promise.race` and orphan its siblings.
- **Cleanup on consumer break.** A `finally` block calls `.return()` on every
  still-pending generator via `Promise.allSettled`, ensuring tracer spans and
  hook callbacks don't leak when the caller exits the `for await` early.

### Use Cases

- Agents with tool-heavy turns (e.g., fanout retrievals across multiple
  sources) where end-to-end latency scales with the slowest tool rather than
  the sum.
- LLM-as-tool patterns where sub-agents each make their own model call.
- Tools that are naturally I/O-bound and benefit from overlapping network
  waits.

#### Call out

Minor: normalized 'tool cancelled by hook' → 'Tool cancelled by hook' to match the sibling 'Tool execution cancelled' message. 

This is technically a breaking change but it is a minor edge case and normalizes our messages.

## Related Issues

Resolves #817

## Documentation PR

Will follow up

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

- unit testing

## Checklist

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
     "prettier": "^3.7.4"
   },
   "scripts": {
-    "prepare": "husky",
     "dev": "strands-dev",
+    "prepare": "husky && npm run build",
     "build": "npm run build -w strands-ts",
     "test": "npm run test -w strands-ts",
     "test:coverage": "npm run test:coverage -w strands-ts",
@@ -18,6 +18,7 @@
     "test:browser:install": "npm run test:browser:install -w strands-ts",
     "test:package": "npm run test:package -w strands-ts",
     "lint": "npm run lint -w strands-ts",
+    "format": "npm run format -w strands-ts",
     "format:check": "npm run format:check -w strands-ts",
     "type-check": "npm run type-check -w strands-ts",
     "check:browser-bundle": "npm run check:browser-bundle -w strands-ts"
--- a/strands-ts/package.json
+++ b/strands-ts/package.json
@@ -78,7 +78,7 @@
   "scripts": {
     "build": "tsc --project src/tsconfig.json",
     "prepack": "npm run build",
-    "check": "npm run build && npm run lint && npm run format && npm run type-check && npm run check:browser-bundle && npm run test:coverage && npm run test:package",
+    "check": "npm run lint && npm run format && npm run type-check && npm run check:browser-bundle && npm run test:coverage && npm run test:package",
     "check:browser-bundle": "esbuild src/index.ts --bundle --platform=browser --format=esm --packages=external --outfile=/dev/null",
     "clean": "rm -rf node_modules dist",
     "lock:refresh": "rm -rf node_modules && npm install --ignore-scripts --os=linux --os=darwin --os=win32 --cpu=x64 --cpu=arm64 --cpu=wasm32",
--- a/strands-ts/src/agent/__tests__/agent.concurrent.test.ts
+++ b/strands-ts/src/agent/__tests__/agent.concurrent.test.ts
@@ -9,54 +9,63 @@ import {
   ToolStreamUpdateEvent,
 } from '../../hooks/index.js'
 import { MockMessageModel } from '../../__fixtures__/mock-message-model.js'
-import { TextBlock, ToolResultBlock } from '../../types/messages.js'
+import { MockPlugin } from '../../__fixtures__/mock-plugin.js'
+import { Message, TextBlock, ToolResultBlock } from '../../types/messages.js'
 import { Tool, ToolStreamEvent, type ToolContext, type ToolStreamGenerator } from '../../tools/tool.js'
 import type { ToolSpec } from '../../tools/types.js'
 
 /**
- * A tool that sleeps for `delayMs` then returns, recording the start and end
- * timestamps on the provided tracker. Useful for asserting concurrency and
- * cooperative cancellation behavior.
+ * A tool whose `stream()` suspends until `release()` is called. Lets tests
+ * drive concurrency deterministically without wall-clock sleeps.
+ *
+ * `started` resolves as soon as the agent enters the tool's `stream()`, so
+ * tests can await "both tools in flight" without polling. The tool also
+ * honors `ctx.agent.cancelSignal`: aborting the signal resolves the gate and
+ * marks `observations.cancelled = true`.
  */
-class TimedTool extends Tool {
+class GatedTool extends Tool {
   name: string
   description: string
   toolSpec: ToolSpec
 
-  constructor(
-    name: string,
-    private readonly delayMs: number,
-    private readonly tracker: { name: string; start: number; end: number; cancelled: boolean }[],
-    private readonly honorCancelSignal = true
-  ) {
+  readonly started: Promise<void>
+  readonly observations = { started: false, cancelled: false, completed: false }
+
+  private _signalStarted!: () => void
+  private readonly _releaser: Promise<void>
+  private _release!: () => void
+
+  constructor(name: string) {
     super()
     this.name = name
-    this.description = `Timed tool ${name}`
+    this.description = `Gated tool ${name}`
     this.toolSpec = { name, description: this.description, inputSchema: { type: 'object', properties: {} } }
+    this.started = new Promise<void>((resolve) => (this._signalStarted = resolve))
+    this._releaser = new Promise<void>((resolve) => (this._release = resolve))
+  }
+
+  release(): void {
+    this._release()
   }
 
   // eslint-disable-next-line require-yield
   async *stream(ctx: ToolContext): ToolStreamGenerator {
-    const entry = { name: this.name, start: Date.now(), end: 0, cancelled: false }
-    this.tracker.push(entry)
-    if (this.honorCancelSignal) {
-      // Cooperatively race sleep against the cancel signal.
-      await new Promise<void>((resolve) => {
-        const timer = globalThis.setTimeout(() => resolve(), this.delayMs)
-        ctx.agent.cancelSignal.addEventListener(
-          'abort',
-          () => {
-            globalThis.clearTimeout(timer)
-            entry.cancelled = true
-            resolve()
-          },
-          { once: true }
-        )
-      })
-    } else {
-      await new Promise<void>((resolve) => globalThis.setTimeout(resolve, this.delayMs))
-    }
-    entry.end = Date.now()
+    this.observations.started = true
+    this._signalStarted()
+
+    await new Promise<void>((resolve) => {
+      void this._releaser.then(resolve)
+      ctx.agent.cancelSignal.addEventListener(
+        'abort',
+        () => {
+          this.observations.cancelled = true
+          resolve()
+        },
+        { once: true }
+      )
+    })
+
+    this.observations.completed = true
     return new ToolResultBlock({
       toolUseId: ctx.toolUse.toolUseId,
       status: 'success',
@@ -66,35 +75,58 @@ class TimedTool extends Tool {
 }
 
 /**
- * A tool that yields the specified number of `ToolStreamEvent`s before returning.
- * Used to exercise cross-tool interleaving of stream updates.
+ * A streaming tool whose `emit(data)` yields a `ToolStreamEvent` and resolves
+ * only after the agent has fully dispatched it; `complete()` terminates the
+ * stream. Tests can drive exact interleaving between tools without timers.
  */
-class StreamingTool extends Tool {
+class GatedStreamingTool extends Tool {
   name: string
   description: string
   toolSpec: ToolSpec
 
-  constructor(
-    name: string,
-    private readonly steps: number,
-    private readonly stepDelayMs: number
-  ) {
+  private readonly _queue: { cmd: { type: 'emit'; data: unknown } | { type: 'complete' }; ack: () => void }[] = []
+  private _notify: (() => void) | null = null
+
+  constructor(name: string) {
     super()
     this.name = name
-    this.description = `Streaming tool ${name}`
+    this.description = `Gated streaming tool ${name}`
     this.toolSpec = { name, description: this.description, inputSchema: { type: 'object', properties: {} } }
   }
 
-  async *stream(ctx: ToolContext): ToolStreamGenerator {
-    for (let i = 0; i < this.steps; i++) {
-      await new Promise<void>((resolve) => globalThis.setTimeout(resolve, this.stepDelayMs))
-      yield new ToolStreamEvent({ data: { tool: this.name, step: i } })
-    }
-    return new ToolResultBlock({
-      toolUseId: ctx.toolUse.toolUseId,
-      status: 'success',
-      content: [new TextBlock(`${this.name} streamed ${this.steps}`)],
+  async emit(data: unknown): Promise<void> {
+    return this._send({ type: 'emit', data })
+  }
+
+  async complete(): Promise<void> {
+    return this._send({ type: 'complete' })
+  }
+
+  private _send(cmd: { type: 'emit'; data: unknown } | { type: 'complete' }): Promise<void> {
+    return new Promise<void>((ack) => {
+      this._queue.push({ cmd, ack })
+      this._notify?.()
+      this._notify = null
     })
+  }
+
+  async *stream(ctx: ToolContext): ToolStreamGenerator {
+    while (true) {
+      while (this._queue.length === 0) {
+        await new Promise<void>((resolve) => (this._notify = resolve))
+      }
+      const { cmd, ack } = this._queue.shift()!
+      if (cmd.type === 'complete') {
+        ack()
+        return new ToolResultBlock({
+          toolUseId: ctx.toolUse.toolUseId,
+          status: 'success',
+          content: [new TextBlock(`${this.name} done`)],
+        })
+      }
+      yield new ToolStreamEvent({ data: cmd.data })
+      ack()
+    }
   }
 }
 
@@ -109,78 +141,103 @@ function twoToolTurn(): MockMessageModel {
 
 describe('Agent concurrent tool execution', () => {
   it('runs multiple tools in parallel', async () => {
-    const tracker: { name: string; start: number; end: number; cancelled: boolean }[] = []
-    const tools = [new TimedTool('toolA', 80, tracker), new TimedTool('toolB', 80, tracker)]
+    const toolA = new GatedTool('toolA')
+    const toolB = new GatedTool('toolB')
     const agent = new Agent({
       model: twoToolTurn(),
-      tools,
+      tools: [toolA, toolB],
       toolExecutor: 'concurrent',
       printer: false,
     })
 
-    await agent.invoke('Go')
-
-    expect(tracker).toHaveLength(2)
-    const [first, second] = tracker.sort((a, b) => a.start - b.start)
-    // B started before A finished — proves the tools ran concurrently.
-    expect(second!.start).toBeLessThan(first!.end)
+    const invocation = agent.invoke('Go')
+    // Both tools reach their stream() before either is released — proves
+    // concurrency without relying on wall-clock overlap.
+    await Promise.all([toolA.started, toolB.started])
+    expect(toolA.observations.completed).toBe(false)
+    expect(toolB.observations.completed).toBe(false)
+    toolA.release()
+    toolB.release()
+    await invocation
+    expect(toolA.observations.completed).toBe(true)
+    expect(toolB.observations.completed).toBe(true)
   })
 
   it('runs tools sequentially under default executor', async () => {
-    const tracker: { name: string; start: number; end: number; cancelled: boolean }[] = []
-    const tools = [new TimedTool('toolA', 40, tracker), new TimedTool('toolB', 40, tracker)]
+    const toolA = new GatedTool('toolA')
+    const toolB = new GatedTool('toolB')
     const agent = new Agent({
       model: twoToolTurn(),
-      tools,
+      tools: [toolA, toolB],
       // default (sequential)
       printer: false,
     })
 
-    await agent.invoke('Go')
-    const [first, second] = tracker.sort((a, b) => a.start - b.start)
-    // B did not start until A finished
-    expect(second!.start).toBeGreaterThanOrEqual(first!.end)
+    const invocation = agent.invoke('Go')
+    await toolA.started
+    // B has not started — sequential executor is still blocked on A.
+    expect(toolB.observations.started).toBe(false)
+    toolA.release()
+    await toolB.started
+    toolB.release()
+    await invocation
   })
 
   it('preserves per-tool event ordering while interleaving across tools', async () => {
-    const tools = [new StreamingTool('toolA', 3, 10), new StreamingTool('toolB', 3, 10)]
+    const toolA = new GatedStreamingTool('toolA')
+    const toolB = new GatedStreamingTool('toolB')
+    const plugin = new MockPlugin()
     const agent = new Agent({
       model: twoToolTurn(),
-      tools,
+      tools: [toolA, toolB],
       toolExecutor: 'concurrent',
       printer: false,
+      plugins: [plugin],
     })
 
-    const events: { kind: string; toolUseId?: string; tool?: string }[] = []
-    agent.addHook(BeforeToolCallEvent, (e) => void events.push({ kind: 'before', toolUseId: e.toolUse.toolUseId }))
-    agent.addHook(AfterToolCallEvent, (e) => void events.push({ kind: 'after', toolUseId: e.toolUse.toolUseId }))
-    agent.addHook(ToolResultEvent, (e) => void events.push({ kind: 'result', toolUseId: e.result.toolUseId }))
-    agent.addHook(ToolStreamUpdateEvent, (e) => {
-      const data = e.event.data as { tool: string } | undefined
-      events.push(data?.tool !== undefined ? { kind: 'stream', tool: data.tool } : { kind: 'stream' })
-    })
+    const invocation = agent.invoke('Go')
+    // Drive explicit A,B,A,B,A,B interleaving.
+    await toolA.emit({ tool: 'toolA', step: 0 })
+    await toolB.emit({ tool: 'toolB', step: 0 })
+    await toolA.emit({ tool: 'toolA', step: 1 })
+    await toolB.emit({ tool: 'toolB', step: 1 })
+    await toolA.emit({ tool: 'toolA', step: 2 })
+    await toolB.emit({ tool: 'toolB', step: 2 })
+    await toolA.complete()
+    await toolB.complete()
+    await invocation
 
-    await agent.invoke('Go')
+    // Reduce MockPlugin's invocations to the per-tool lifecycle events we care about.
+    type Entry = { kind: string; toolUseId?: string; tool?: string }
+    const events: Entry[] = plugin.invocations
+      .map((e): Entry | null => {
+        if (e instanceof BeforeToolCallEvent) return { kind: 'before', toolUseId: e.toolUse.toolUseId }
+        if (e instanceof AfterToolCallEvent) return { kind: 'after', toolUseId: e.toolUse.toolUseId }
+        if (e instanceof ToolResultEvent) return { kind: 'result', toolUseId: e.result.toolUseId }
+        if (e instanceof ToolStreamUpdateEvent) {
+          const data = e.event.data as { tool?: string } | undefined
+          return data?.tool !== undefined ? { kind: 'stream', tool: data.tool } : { kind: 'stream' }
+        }
+        return null
+      })
+      .filter((e): e is Entry => e !== null)
 
-    // Extract per-tool subsequence and validate its shape.
+    // Per-tool subsequence shape: [before, stream*, after, result].
     for (const toolUseId of ['a', 'b']) {
       const subseq = events.filter(
         (e) => e.toolUseId === toolUseId || (e.kind === 'stream' && e.tool === (toolUseId === 'a' ? 'toolA' : 'toolB'))
       )
       const kinds = subseq.map((e) => e.kind)
-      // First event for a tool is its BeforeToolCallEvent
       expect(kinds[0]).toBe('before')
-      // Last two are AfterToolCallEvent then ToolResultEvent
       expect(kinds.slice(-2)).toEqual(['after', 'result'])
-      // Middle events (if any) are all stream updates
       for (const k of kinds.slice(1, -2)) {
         expect(k).toBe('stream')
       }
     }
 
-    // Cross-tool interleaving: collapse consecutive same-tool events into runs.
-    // Strictly sequential execution produces 2 runs (e.g. [A,A,A,B,B,B]);
-    // anything > 2 means the stream alternated between tools at least once.
+    // Cross-tool interleaving: collapse consecutive same-tool stream events
+    // into runs. Strictly sequential execution produces 2 runs (A,A,A,B,B,B);
+    // anything > 2 means the stream alternated at least once.
     const streamTools = events.filter((e) => e.kind === 'stream').map((e) => e.tool)
     const runs = streamTools.reduce<(string | undefined)[]>((acc, t) => {
       if (acc.length === 0 || acc[acc.length - 1] !== t) acc.push(t)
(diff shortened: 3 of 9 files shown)


strands-ts/src/agent/__tests__/agent.concurrent.test.ts
```diff
@@ -0,0 +1,493 @@
...
+import { TextBlock, ToolResultBlock } from '../../types/messages.js'
+import { Tool, ToolStreamEvent, type ToolContext, type ToolStreamGenerator } from '../../tools/tool.js'
+import type { ToolSpec } from '../../tools/types.js'
+
+/**
+ * A tool that sleeps for `delayMs` then returns, recording the start and end
+ * timestamps on the provided tracker. Useful for asserting concurrency and
+ * cooperative cancellation behavior.
+ */
+class TimedTool extends Tool {
```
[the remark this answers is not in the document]
notowen333: TimedTool is tied to this file's use case: tracker-pushing, cancelSignal race,          
  wall-clock capture. Outside of concurrency/cancellation tests specifically for the             
  tool-execution loop, nobody has a reason to reach for it

strands-ts/src/agent/__tests__/agent.hook.test.ts
```diff
@@ -545,7 +545,7 @@ describe('Agent Hooks Integration', () => {
         new ToolResultBlock({
           toolUseId: 'tool-1',
           status: 'error',
-          content: [new TextBlock('tool cancelled by hook')],
+          content: [new TextBlock('Tool cancelled by hook')],
```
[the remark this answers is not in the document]
notowen333: 1. The other cancel message was already capitalized ('Tool execution cancelled'). Before this  
  PR, the two sibling messages disagreed — any user pattern-matching literally would already have
   to handle both cases. Normalizing them makes the surface less surprising, not more.           
  2. Scope is extremely narrow. The lowercase message only appeared when a hook set
  BeforeToolCallEvent.cancel = true (the boolean form). Users who set a string get passthrough,  
  unaffected.
  
  will add to pr description

strands-ts/src/agent/__tests__/agent.concurrent.test.ts
```diff
@@ -0,0 +1,471 @@
...
+          () => {
+            globalThis.clearTimeout(timer)
+            entry.cancelled = true
+            resolve()
+          },
+          { once: true }
+        )
+      })
+    } else {
+      await new Promise<void>((resolve) => globalThis.setTimeout(resolve, this.delayMs))
```
zastrowm: How long do these take to run?  I'd rather not use explicit sleeps in favor of tests that control the lifecycle exactly via promises or some other mechanism

strands-ts/src/agent/__tests__/agent.concurrent.test.ts
```diff
@@ -0,0 +1,471 @@
...
+    const tools = [new StreamingTool('toolA', 3, 10), new StreamingTool('toolB', 3, 10)]
+    const agent = new Agent({
+      model: twoToolTurn(),
+      tools,
+      toolExecutor: 'concurrent',
+      printer: false,
+    })
+
+    const events: { kind: string; toolUseId?: string; tool?: string }[] = []
+    agent.addHook(BeforeToolCallEvent, (e) => void events.push({ kind: 'before', toolUseId: e.toolUse.toolUseId }))
```
zastrowm: Can we use `MockPlugin` instead and just verify the `invocations` events?

strands-ts/src/agent/__tests__/agent.concurrent.test.ts
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    await agent.invoke('Go')
+
+    const [a, b] = results.sort((x, y) => x.toolUseId.localeCompare(y.toolUseId))
+    expect(a!.status).toBe('error')
+    expect(b!.status).toBe('success')
+  })
+
+  it('handles an unknown tool in a batch without affecting siblings', async () => {
```
zastrowm: What is this doing/verifying?

strands-ts/src/agent/__tests__/agent.concurrent.test.ts
```diff
@@ -0,0 +1,471 @@
...
+          { type: 'toolUseBlock', name: 'toolA', toolUseId: 'a', input: {} },
+          { type: 'toolUseBlock', name: 'unknownTool', toolUseId: 'b', input: {} },
+        ])
+        .addTurn({ type: 'textBlock', text: 'Done' }),
+      tools,
+      toolExecutor: 'concurrent',
+      printer: false,
+    })
+
+    let afterMessage: import('../../types/messages.js').Message | undefined
```
zastrowm: Why not make this top level?

strands-ts/src/agent/__tests__/agent.hook.test.ts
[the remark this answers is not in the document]
zastrowm: It makes me chuckle that we're justifying ourselves to the bot ;)

strands-ts/src/agent/agent.ts
```diff
@@ -162,6 +178,11 @@ export type AgentConfig = {
    * Optional unique identifier for the agent. Defaults to "agent".
    */
   id?: string
+  /**
+   * Strategy for executing tool calls from a single assistant turn.
+   * Defaults to `'sequential'`. See {@link ToolExecutor} for details.
+   */
+  toolExecutor?: ToolExecutor
```
zastrowm: Thoughts on changing this to parallel before 1.0?

strands-ts/src/agent/agent.ts
```diff
@@ -1087,6 +1130,152 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+    const beforeToolsEvent = new BeforeToolsEvent({ agent: this, message: assistantMessage })
+    yield beforeToolsEvent
+
+    const toolUseBlocks = assistantMessage.content.filter(
+      (block): block is ToolUseBlock => block.type === 'toolUseBlock'
+    )
+
+    if (toolUseBlocks.length === 0) {
+      throw new Error('Model indicated toolUse but no tool use blocks found in message')
+    }
```
zastrowm: This should be done before the call and the toolUseBlocks are passed in instead.  E.g. factor out the common code between the two executors

strands-ts/src/agent/agent.ts
```diff
@@ -1087,6 +1130,152 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+
+    try {
+      while (pendingNext.size > 0) {
+        const winner = await Promise.race(pendingNext.values())
+        const { i } = winner
+        const block = gens[i]!.block
+
+        if (winner.kind === 'throw') {
+          pendingNext.delete(i)
+          const err = normalizeError(winner.error)
```
zastrowm, line 1205: Reading this line makes me chuckle

> Normalize the error of the winner!

strands-ts/src/agent/agent.ts
```diff
@@ -1087,6 +1130,152 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+            status: 'error',
+            content: [new TextBlock(err.message)],
+            error: err,
+          })
+          resultsByToolUseId.set(block.toolUseId, result)
+          yield new ToolResultEvent({ agent: this, result })
+          continue
+        }
+
+        if (winner.r.done) {
```
zastrowm: the `r` here stands out compared to the rest; I presume `r` is result?

package.json
```diff
@@ -7,8 +7,8 @@
     "prettier": "^3.7.4"
   },
   "scripts": {
-    "prepare": "husky",
     "dev": "strands-dev",
+    "prepare": "husky && npm run build",
```
pgrayy, line 12: I think this does make sense to add. We used to run build on prepare. We should update the strands-ts/package.json though to remove build from the `check` command ([here](https://github.com/strands-agents/sdk-typescript/blob/main/strands-ts/package.json#L81)). It'll now be redundant since prepare is called on every `npm install/ci`.

strands-ts/src/agent/agent.ts
```diff
@@ -80,6 +80,22 @@ import { CancelledError } from '../errors.js'
...
+ * - `'concurrent'` — runs all tool calls from a single turn in parallel. Per-tool event
+ *   order (`BeforeToolCallEvent` → `ToolStreamUpdateEvent*` → `AfterToolCallEvent` →
+ *   `ToolResultEvent`) is preserved, while cross-tool events may interleave.
+ *
+ * Cancellation works identically in both modes: {@link Agent.cancel} flips
+ * {@link Agent.cancelSignal} and tools must observe it cooperatively to stop early.
+ * In concurrent mode, prompt batch-wide cancellation requires every in-flight tool
+ * to honor the signal.
+ */
+export type ToolExecutor = 'sequential' | 'concurrent'
```
pgrayy: We should call this something else. Maybe `ToolExecutorStrategy`. The reason is that we may want to expose a ToolExecutor interface for people to derive their own custom executor. This is what we planned for in Python (the interface is currently private).

strands-ts/src/agent/agent.ts
```diff
@@ -80,6 +80,22 @@ import { CancelledError } from '../errors.js'
  */
 export type ToolList = (Tool | McpClient | Agent | ToolList)[]
 
+/**
+ * Strategy for executing tool calls that the model emits in a single assistant turn.
+ *
+ * - `'sequential'` (default) — runs tool calls one at a time, matching the historical
```
pgrayy: Nit: The "historical" call out seems out of place. Generally speaking, I think it good to always speak in the present in docstrings and avoid referencing past behaviors.

strands-ts/src/agent/agent.ts
pgrayy: I recommend updating the type name to ToolExecutorStrategy above, but I think we can still call this param toolExecutor. It could be of type ToolExecutorStrategy to start but if we support custom executor implementations in the future, we can switch to a union (similar to model).

strands-ts/src/agent/agent.ts
pgrayy: Additional perspective: https://github.com/strands-agents/docs/blob/main/designs/0005-state-machine.md#orchestrators

strands-ts/src/agent/agent.ts
```diff
@@ -1024,6 +1049,25 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+  }
+
+  /**
+   * Executes tools sequentially and streams all tool events.
+   */
+  private async *_executeToolsSequential(
+    assistantMessage: Message,
+    toolRegistry: ToolRegistry
   ): AsyncGenerator<AgentStreamEvent, Message, undefined> {
     const beforeToolsEvent = new BeforeToolsEvent({ agent: this, message: assistantMessage })
```
pgrayy: Are we able to move some of the "before tools" and "after tools" logic into `executeTools`. Looks like some of this logic is getting duplicated between _executeToolsSequential and _executeToolsConcurrent.

strands-ts/src/agent/agent.ts
```diff
@@ -1087,6 +1130,152 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+        (r): Step => ({ i, kind: 'next', r }),
+        (error: unknown): Step => ({ i, kind: 'throw', error })
+      )
+
+    const pendingNext = new Map<number, Promise<Step>>(gens.map((_, i) => [i, step(i)]))
+    const resultsByToolUseId = new Map<string, ToolResultBlock>()
+
+    try {
+      while (pendingNext.size > 0) {
+        const winner = await Promise.race(pendingNext.values())
```
pgrayy, line 1199: Note: When multiple generators have already-resolved promises, `Promise.race` picks by map insertion order, so a chatty low-index tool can starve higher-indexed tools' events. Compare that to the queue approach taken in Python which has more fair FIFO ordering. Not sure though if this will really be a problem in practice. May be something to consider though with agents-as-tools which do have many events to stream back.

strands-ts/src/agent/__tests__/agent.concurrent.test.ts
notowen333: 400ms. Will swap away from sleeps

strands-ts/src/agent/__tests__/agent.concurrent.test.ts
notowen333: renamed to "handles a hallucinated tool name in..."

The scenario is concurrent execution of tools in the tool registry and not.

strands-ts/src/agent/agent.ts
notowen333: Do you mean defaulting to concurrent? Or renaming concurrent to parallel?

notowen333: @zastrowm means to make concurrent the default strategy. I'm going to push that off to a future PR

strands-ts/src/agent/agent.ts
notowen333: Updated i -> idx and r -> res for readability :)

package.json
notowen333: sure I can remove that!

strands-ts/src/agent/agent.ts
notowen333: On board with this rename

package.json
```diff
@@ -7,8 +7,8 @@
     "prettier": "^3.7.4"
   },
   "scripts": {
-    "prepare": "husky",
     "dev": "strands-dev",
```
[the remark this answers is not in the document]
notowen333: these are small changes

strands-ts/src/agent/agent.ts
notowen333: resolved via @zastrowm comment

strands-ts/src/agent/agent.ts
notowen333: That's a great point. I think we'd want to start by trying a round robin the starting index so a different tool is first each time.

I think that could be considered in a separate PR. There are also some other approaches but that would be the lowest hanging fruit

strands-ts/src/agent/agent.ts
```diff
@@ -1087,6 +1130,152 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+    }
+    if (this.isCancelled) {
+      const cancelBlocks = this._cancelAllAsResults(toolUseBlocks, 'Tool execution cancelled')
+      for (const result of cancelBlocks) {
+        yield new ToolResultEvent({ agent: this, result })
+      }
+      toolResultMessage = new Message({ role: 'user', content: cancelBlocks })
+      yield new AfterToolsEvent({ agent: this, message: toolResultMessage })
+      return toolResultMessage
+    }
```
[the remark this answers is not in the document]
notowen333: resolved in another comment

strands-ts/src/agent/agent.ts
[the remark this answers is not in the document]
notowen333: it has this label

strands-ts/src/index.ts
```diff
@@ -13,7 +13,7 @@ export { StateStore } from './state-store.js'
 
 // Agent types
 export { AgentResult } from './types/agent.js'
-export type { AgentConfig, ToolList } from './agent/agent.js'
+export type { AgentConfig, ToolList, ToolExecutorStrategy } from './agent/agent.js'
```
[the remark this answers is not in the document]
notowen333: Stale — `MultiAgentSaveLatestStrategy` is still exported at `src/index.ts:236`. Rebase picked up #873.

strands-ts/package.json
```diff
@@ -77,7 +77,7 @@
   },
   "scripts": {
     "build": "tsc --project src/tsconfig.json",
-    "check": "npm run build && npm run lint && npm run format && npm run type-check && npm run check:browser-bundle && npm run test:coverage && npm run test:package",
+    "check": "npm run lint && npm run format && npm run type-check && npm run check:browser-bundle && npm run test:coverage && npm run test:package",
```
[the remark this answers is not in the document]
notowen333: Stale — rebased onto latest main, `prepack` is present and `build` is back in the `check` script.

package.json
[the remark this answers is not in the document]
notowen333: Keeping these as-is — they are small repo-hygiene changes and splitting them would churn more than it clarifies.

strands-ts/src/agent/agent.ts
```diff
@@ -80,6 +80,21 @@ import { CancelledError } from '../errors.js'
...
+ * - `'concurrent'` — runs all tool calls from a single turn in parallel. Per-tool event
+ *   order (`BeforeToolCallEvent` → `ToolStreamUpdateEvent*` → `AfterToolCallEvent` →
+ *   `ToolResultEvent`) is preserved, while cross-tool events may interleave.
+ *
+ * Cancellation works identically in both modes: {@link Agent.cancel} flips
+ * {@link Agent.cancelSignal} and tools must observe it cooperatively to stop early.
+ * In concurrent mode, prompt batch-wide cancellation requires every in-flight tool
+ * to honor the signal.
+ */
+export type ToolExecutorStrategy = 'sequential' | 'concurrent'
```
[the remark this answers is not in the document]
notowen333: Renamed to `ToolExecutorStrategy` per @pgrayy's feedback (https://github.com/strands-agents/sdk-typescript/pull/854#discussion_r3132330590) — reserves the `ToolExecutor` name for a future class-based interface, matching the direction described in the Python SDK's planned orchestrator work.

strands-ts/src/agent/agent.ts
```diff
@@ -1028,59 +1056,199 @@ export class Agent implements LocalAgent, InvokableAgent {
...
     yield beforeToolsEvent
 
+    const toolUseBlocks = assistantMessage.content.filter(
+      (block): block is ToolUseBlock => block.type === 'toolUseBlock'
+    )
+    if (toolUseBlocks.length === 0) {
+      // Preserve BeforeToolsEvent/AfterToolsEvent bracket symmetry even on
+      // this invariant-violation branch.
+      yield new AfterToolsEvent({ agent: this, message: new Message({ role: 'user', content: [] }) })
+      throw new Error('Model indicated toolUse but no tool use blocks found in message')
```
pgrayy, line 1066: Maybe could check empty tool use blocks before we emit the before tools event. Could consider for a follow up PR though. Doesn't have to be addressed here.

strands-ts/src/agent/agent.ts
```diff
@@ -1028,59 +1056,199 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+          continue
+        }
+
+        const toolResultBlock = yield* this.executeTool(toolUseBlock, toolRegistry)
+        toolResultBlocks.push(toolResultBlock)
+        yield new ToolResultEvent({ agent: this, result: toolResultBlock })
+      }
+    } finally {
+      toolResultMessage = new Message({ role: 'user', content: toolResultBlocks })
+      yield new AfterToolsEvent({ agent: this, message: toolResultMessage })
```
pgrayy, line 1138: Nit: Both _executeToolsSequential and _executeToolsConcurrent yield AfterToolsEvent. This could instead be done centrally in executeTools. I think that would be nice since this is also where we yield BeforeToolsEvent.

strands-ts/src/agent/__tests__/agent.tracer.test.node.ts
```diff
@@ -440,6 +440,50 @@ describe('Agent tracer integration', () => {
...
+        .addTurn([
+          new ToolUseBlock({ name: 'tool1', toolUseId: 'id-1', input: {} }),
+          new ToolUseBlock({ name: 'tool2', toolUseId: 'id-2', input: {} }),
+        ])
+        .addTurn({ type: 'textBlock', text: 'Done' })
+
+      // Tools sleep briefly so the concurrent executor has time to launch both
+      // before either resolves. The assertions below check call order, not
+      // wall-clock timing.
+      const sleep = (ms: number) => new Promise<void>((r) => globalThis.setTimeout(r, ms))
```
pgrayy, line 455: Generally I would recommend avoiding sleeps in unit tests since they are should be executed as fast as possible. As an alternative approach, you might be able to use cooperative signaling. Here is an example in Python that uses asyncio.Event. Not sure what the equivalents would be in TS.

An advantage here also is that you can make the tests even more deterministic.

pgrayy: Looks like you have an approach worked out for strands-ts/src/agent/__tests__/agent.concurrent.test.ts.

strands-ts/src/agent/__tests__/agent.concurrent.test.ts
```diff
@@ -0,0 +1,564 @@
...
+    // AfterToolsEvent.message should have entries for both tools:
+    // toolA completed normally, toolB gets a fallback "interrupted" result.
+    expect(afterToolsMessage).toBeDefined()
+    const blocks = afterToolsMessage!.content as ToolResultBlock[]
+    expect(blocks).toHaveLength(2)
+    expect(blocks.map((b) => b.toolUseId)).toEqual(['a', 'b'])
+    expect(blocks.find((b) => b.toolUseId === 'a')!.status).toBe('success')
+    expect(blocks.find((b) => b.toolUseId === 'b')!.status).toBe('error')
+  })
+})
```
pgrayy, line 564: Considerations: I would recommend adding a test for asserting mid-tool stream events are yielded properly through agent.stream when running in concurrent mode. I would also recommend testing BeforeToolCallEvent.cancel.

pgrayy (approved):

strands-ts/src/agent/agent.ts
notowen333: Will address this and below comments in a follow up

Issue #447, opened 2026-01-26
[FEATURE] Explore OTEL browser support


lizradway:
### Problem Statement

# Browser Support for OpenTelemetry Configuration

## Problem
The current telemetry configuration in `src/telemetry/config.ts` uses Node.js-specific implementations (`NodeTracerProvider` and `AsyncHooksContextManager`) that are incompatible with browser environments.

## Proposed Solution
Create a separate browser-compatible implementation using:
- `WebTracerProvider` from `@opentelemetry/sdk-trace-web`
- `ZoneContextManager` from `@opentelemetry/context-zone` for context propagation

## Implementation Approach
1. Create `src/telemetry/config.browser.ts` with browser-specific implementation
2. Keep existing `src/telemetry/config.ts` for Node.js environments
3. Use conditional exports in `package.json` to load the appropriate version:
   ```json
   "exports": {
     "./telemetry": {
       "browser": "./dist/telemetry/config.browser.js",
       "default": "./dist/telemetry/config.js"
     }
   }
   ```

## Considerations
- Browser environments typically use different exporters (e.g., OTLP over fetch/XHR instead of http)
- Zone.js dependency for `ZoneContextManager` may not be suitable for all projects
- May need to handle differences in resource detection between Node and browser
- Console exporter behavior differs between environments

## Related
- Current implementation: `src/telemetry/config.ts`
- Documentation note already mentions Node.js-only compatibility

PR #867, opened 2026-04-22
chore: upgrade to otel js sdk v2

lizradway:
## Description

PR #831 upgraded the OTel JS SDK from v1 to v2 and replaced `provider.register()` with `trace.setGlobalTracerProvider()`. This broke context propagation because `register()` also sets the `AsyncLocalStorageContextManager` and W3C propagators — without them, `context.active()` always returns a noop root context.

This is because `.register` was removed from `BasicTraceProvider` in OTEL JS SDK v2: https://github.com/open-telemetry/opentelemetry-js/pull/5503

This caused:
- `trace.getSpan(context.active())` returning `undefined` inside tool callbacks
- MCP trace context injection (`injectTraceContext`) writing no `traceparent` into `_meta`
- `tracer.startActiveSpan()` not propagating parent context to child spans

**Fix:** Alongside `trace.setGlobalTracerProvider()`, also register the `AsyncLocalStorageContextManager` and `CompositePropagator` (W3C TraceContext + Baggage) using the same dynamic require pattern already used for `NodeTracerProvider`. In browser environments (where these aren't available), only the tracer provider is registered — matching the existing behavior.

Also fixes the integration test which had the same `setGlobalTracerProvider` bug, causing a flaky context propagation test.

## Related Issues

Fixes context propagation regression introduced by #831

## Documentation PR

N/A

## Type of Change

Bug fix

## Testing

- [X] I ran `npm run check`
- Tested locally in various smoke tests w langfuse

## Checklist

----

### DIFF
--- a/strands-ts/src/telemetry/config.ts
+++ b/strands-ts/src/telemetry/config.ts
@@ -15,7 +15,13 @@
  */
 
 import { context as otelContext, metrics as otelMetrics, propagation, trace } from '@opentelemetry/api'
-import type { ContextManager, Meter as OtelMeter, TextMapPropagator, TracerProvider, Tracer as OtelTracer } from '@opentelemetry/api'
+import type {
+  ContextManager,
+  Meter as OtelMeter,
+  TextMapPropagator,
+  TracerProvider,
+  Tracer as OtelTracer,
+} from '@opentelemetry/api'
 import { resourceFromAttributes, envDetector, type Resource } from '@opentelemetry/resources'
 import {
   BasicTracerProvider,
--- a/strands-ts/src/telemetry/config.ts
+++ b/strands-ts/src/telemetry/config.ts
@@ -15,8 +15,7 @@
  */
 
 import { context as otelContext, metrics as otelMetrics, propagation, trace } from '@opentelemetry/api'
-import type { ContextManager, TextMapPropagator } from '@opentelemetry/api'
-import type { Meter as OtelMeter, Tracer as OtelTracer } from '@opentelemetry/api'
+import type { ContextManager, Meter as OtelMeter, TextMapPropagator, TracerProvider, Tracer as OtelTracer } from '@opentelemetry/api'
 import { resourceFromAttributes, envDetector, type Resource } from '@opentelemetry/resources'
 import {
   BasicTracerProvider,
@@ -120,7 +119,7 @@ export interface TracerConfig {
    * Custom TracerProvider instance. If not provided, NodeTracerProvider is
    * used when available, otherwise BasicTracerProvider.
    */
-  provider?: BasicTracerProvider
+  provider?: TracerProvider
 
   /**
    * Exporter configuration.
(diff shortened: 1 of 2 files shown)


afarntrog (dismissed):

afarntrog (approved):

strands-ts/src/telemetry/config.ts
```diff
@@ -145,30 +164,34 @@ let _provider: BasicTracerProvider | null = null
  * telemetry.setupTracer({ exporters: { otlp: true } })
  * ```
  */
-export function setupTracer(config: TracerConfig = {}): BasicTracerProvider {
-  if (_provider) {
+export function setupTracer(config?: Omit<TracerConfig, 'provider'>): BasicTracerProvider
+export function setupTracer(config: TracerConfig): TracerProvider
+export function setupTracer(config: TracerConfig = {}): TracerProvider {
```
poshinchen, line 169: Are these valid?

strands-ts/src/telemetry/config.ts
```diff
@@ -145,30 +164,34 @@ let _provider: BasicTracerProvider | null = null
...
 
   if (typeof globalThis.process?.once === 'function') {
     globalThis.process.once('beforeExit', () => {
-      if (_provider) {
-        _provider.forceFlush().catch((err: unknown) => {
-          logger.warn(`error=<${err}> | failed to flush tracer provider on exit`)
-        })
-      }
+      _provider?.forceFlush()?.catch((err: unknown) => {
+        logger.warn(`error=<${err}> | failed to flush tracer provider on exit`)
```
poshinchen, line 193: We removed force_flush in python, do we want to do that in ts?

strands-ts/src/telemetry/config.ts
lizradway: yes, although i hate them. ultimately the long term path should be to migrate to separate Node Tracer Provider and Web Tracer Providers as we had originally planned in https://github.com/strands-agents/sdk-typescript/issues/447


it is set up like this now, becaues we opted for the default `BasicTracerProvider` to quickly unblock wasm/browser support in https://github.com/strands-agents/sdk-typescript/pull/622

lizradway: 1. no custom tracer provider -> use BasicTracerProvider class
2. custom tracer provider -> use Tracer Provider interface
3. not exposed

lizradway: I think this messiness to me begs that maybe we should wait to upgrade to v2 until we support `WebTracerProvider`, but lmk ur thoughts

poshinchen: LGTM

poshinchen (approved):

strands-ts/src/telemetry/config.ts
lizradway: this forceFlush is in the beforeExit hook to ensure buffered spans get exported before the process exits rather than after every span end as in python

lizradway: do we want users to force_flush themselves? is that what they are currently doing in python?

Issue #846, opened 2026-04-21
[FEATURE] Interventions Primitive


lizradway:
### Problem Statement

Strands agents have multiple independent control layers (authorization, steering, guardrails) but no shared interface between them. Each is a standalone plugin with its own action vocabulary, no ordering guarantees, no short-circuiting, and no unified audit trail. Plugins communicate by mutating the event object — the framework can't distinguish a hard deny from a soft guide, can't skip expensive handlers when an earlier one already blocked, and can't log decisions in a single stream.

### Proposed Solution

Implement the intervention primitive in the TypeScript SDK:

- `InterventionAction` type — `Proceed | Deny | Guide | Interrupt | Transform`
- `InterventionHandler` base class — default no-op methods per lifecycle event (`beforeToolCall`, `afterToolCall`, `beforeModelCall`, `afterModelCall`), framework detects which are overridden
- `InterventionRegistry` — wires handlers into the hook system, dispatches in registration order, applies conflict resolution (Deny > Interrupt > Transform > Guide > Proceed)
- `Agent({ interventions: [...] })` — first-class API parameter
- Unified audit trail at `agent.interventions.auditLog`

```typescript
const agent = new Agent({
    tools: [queryDatabase, sendEmail],
    interventions: [cedar, guardrails, steering],
})
```

### Use Case

A developer building an agent with Cedar authorization, content guardrails, and LLM steering wants them to compose — Cedar denies in sub-ms and steering never runs, all decisions are logged in one place, and Deny/Guide/Interrupt/Transform are handled correctly by the framework.

### Alternative Solutions

- Keep separate plugins — each re-implements short-circuiting, audit, conflict resolution independently. See [Appendix C of the design doc](https://github.com/strands-agents/docs/designs/0007-intervention-primitive.md) for why this doesn't scale.
- Userland wrapper — we built an `InterventionPipeline` that works but every team ends up writing the same thing.

### Additional Context

Design doc: [0007-intervention-primitive.md](https://github.com/strands-agents/docs/designs/0007-intervention-primitive.md)

PR #883, opened 2026-04-23
feat: add interventions primitive

lizradway:
## Description

Adds the intervention primitive — a composable control layer for agents that enables authorization (Cedar, OPA), guardrails (Datadog AI Guard, content filtering), steering (LLM-based guidance), and operational controls (Galileo Agent Control) to share a common interface with ordered evaluation, short-circuiting, and typed actions.

### Example usage

```typescript
import { Agent, InterventionHandler, BeforeToolCallEvent, deny, proceed } from '@strands-agents/sdk'

class CedarAuth extends InterventionHandler {
  readonly name = 'cedar-auth'

  override beforeToolCall(event: BeforeToolCallEvent) {
    const userId = event.agent.appState.get('user_id')
    if (!this.isAuthorized(userId, event.toolUse.name)) {
      return deny(`User ${userId} not authorized for ${event.toolUse.name}`)
    }
    return proceed()
  }
}

class ToneSteeringHandler extends InterventionHandler {
  readonly name = 'tone-steering'

  override afterModelCall(event: AfterModelCallEvent) {
    const text = event.stopData?.message.content
      .filter((b) => b.type === 'textBlock')
      .map((b) => b.text)
      .join('')
    if (text && this.isTooAgressive(text)) {
      return guide('Use a more professional tone.')
    }
    return proceed()
  }
}

const agent = new Agent({
  tools: [queryDatabase, sendEmail],
  interventions: [new CedarAuth(), new ToneSteeringHandler()],
})

const result = await agent.invoke('Send an email to the client')
```

Cedar denies in sub-ms, steering never runs for denied tool calls (short-circuiting).

---

### Public Exports (from `@strands-agents/sdk`)

```typescript
// Class — consumers extend this
export { InterventionHandler }

// Factory helpers — construct action return values
export { proceed, deny, guide, interrupt, transform }

// Type — error policy
export type { OnError }
```

### `AgentConfig.interventions`

```typescript
interventions?: InterventionHandler[]
```

### `InterventionHandler`

```typescript
abstract class InterventionHandler {
  readonly onError: OnError = 'throw'

  initAgent(agent: LocalAgent): void | Promise<void>

  beforeInvocation(event): Proceed | Deny | Guide | Transform
  beforeToolCall(event): Proceed | Deny | Guide | Interrupt | Transform
  afterToolCall(event): Proceed | Transform
  beforeModelCall(event): Proceed | Deny | Guide | Transform
  afterModelCall(event): Proceed | Guide | Transform
}
```

All lifecycle methods default to `proceed()`. Override only the ones you need — the framework detects overrides and only registers hooks for those.

### Action Types

| Action | Factory | Description |
|--------|---------|-------------|
| Proceed | `proceed(reason?)` | Allow the operation to continue |
| Deny | `deny(reason)` | Block the operation (sets event.cancel) |
| Guide | `guide(feedback, reason?)` | Steer with feedback |
| Interrupt | `interrupt(prompt, reason?)` | Pause for human approval (beforeToolCall only) |
| Transform | `transform(apply, reason?)` | Modify event content in-place |

### Action-to-Event Compatibility Matrix

| Action    | beforeInvocation | beforeToolCall | beforeModelCall | afterToolCall | afterModelCall |
|-----------|------------------|----------------|-----------------|---------------|----------------|
| Proceed   | —                | —              | —               | —             | —              |
| Deny      | cancel           | cancel         | cancel          | —             | —              |
| Guide     | cancel+          | cancel+        | inject          | —             | inject + retry |
| Interrupt | —                | interrupt      | —               | —             | —              |
| Transform | apply            | apply          | apply           | apply         | apply          |

- **—** = no-op (warns at runtime)
- **cancel** = sets event.cancel, short-circuits remaining handlers
- **cancel+** = sets event.cancel with accumulated feedback from all guiding handlers
- **interrupt** = calls event.interrupt() for native pause/resume
- **inject** = appends feedback as a user message so the model sees it on this call
- **inject + retry** = appends feedback and retries so the model sees guidance
- **apply** = calls action.apply(event) for in-place mutation

### Hook Ordering

- Before* events: interventions run **last** (`SDK_LAST`) — closest to execution
- After* events: interventions run **first** (`SDK_FIRST`) — closest to execution

Flow: `plugins → intervention → [operation] → intervention → plugins`

### `onError` Policy

| Value | Behavior |
|-------|----------|
| `'throw'` | Rethrow (default). Invocation fails. |
| `'proceed'` | Skip handler, continue to next. |
| `'deny'` | Apply Deny (fail-closed). |

### `initAgent(agent)`

Called once during agent initialization. Use to register context-gathering hooks (e.g. steering context providers):

```typescript
override initAgent(agent: LocalAgent): void {
  agent.addHook(MessageAddedEvent, (event) => this.updateContext(event))
}
```

---

## What's NOT included (internal / planned)

- **InterventionRegistry** — not publicly exported. Internal dispatch mechanism.
- **Audit log** — removed for now. Will be added when consumption patterns are clear.
- **`agent.interventions` getter** — removed. Registry is fully internal.

## Related Issues

- strands-agents/sdk-typescript#846
- strands-agents/sdk-typescript#889
- strands-agents/sdk-typescript#906

## Documentation PR

N/A

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`
- 32 unit tests covering all action types, short-circuiting, guide accumulation, dispatch ordering, override detection, all three onError modes, async handlers, duplicate name rejection, conflict resolution, initAgent, native interrupt, and full Agent integration (deny prevents tool execution, hook ordering verified)

## Checklist

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -447,8 +447,6 @@ export class Agent implements LocalAgent, InvokableAgent {
       })
     )
 
-    await this._interventionRegistry?.initialize(this)
-
     await this._pluginRegistry.initialize(this)
 
     await this._hooksRegistry.invokeCallbacks(new InitializedEvent({ agent: this }))
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -36,6 +36,8 @@ import { ToolRegistry } from '../registry/tool-registry.js'
 import { StateStore } from '../state-store.js'
 import { AgentPrinter, getDefaultAppender, type Printer } from './printer.js'
 import type { Plugin } from '../plugins/plugin.js'
+import type { InterventionHandler } from '../interventions/handler.js'
+import { InterventionRegistry } from '../interventions/registry.js'
 import { PluginRegistry } from '../plugins/registry.js'
 import { SlidingWindowConversationManager } from '../conversation-manager/sliding-window-conversation-manager.js'
 import { NullConversationManager } from '../conversation-manager/null-conversation-manager.js'
@@ -181,6 +183,10 @@ export type AgentConfig = {
    */
   retryStrategy?: RetryStrategy | RetryStrategy[] | null
   /**
+   * Intervention handlers evaluated in registration order at each lifecycle point.
+   */
+  interventions?: InterventionHandler[]
+  /**
    * Zod schema for structured output validation.
    */
   structuredOutputSchema?: z.ZodSchema
@@ -280,6 +286,7 @@ export class Agent implements LocalAgent, InvokableAgent {
 
   private readonly _hooksRegistry: HookRegistryImplementation
   private readonly _pluginRegistry: PluginRegistry
+  private readonly _interventionRegistry?: InterventionRegistry
   private _toolRegistry: ToolRegistry
   private _mcpClients: McpClient[]
   private _initialized: boolean
@@ -336,6 +343,10 @@ export class Agent implements LocalAgent, InvokableAgent {
 
     // Initialize hooks registry
     this._hooksRegistry = new HookRegistryImplementation()
+
+    if (config?.interventions && config.interventions.length > 0) {
+      this._interventionRegistry = new InterventionRegistry(config.interventions, this._hooksRegistry)
+    }
 
     // `undefined` (omitted) → install the default; `null`/`[]` → explicit opt-out.
     const retryStrategies: RetryStrategy[] =
@@ -435,6 +446,8 @@ export class Agent implements LocalAgent, InvokableAgent {
         }
       })
     )
+
+    await this._interventionRegistry?.initialize(this)
 
     await this._pluginRegistry.initialize(this)
 
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -286,7 +286,7 @@ export class Agent implements LocalAgent, InvokableAgent {
 
   private readonly _hooksRegistry: HookRegistryImplementation
   private readonly _pluginRegistry: PluginRegistry
-  private readonly _interventionRegistry?: InterventionRegistry
+  private readonly _interventionRegistry: InterventionRegistry
   private _toolRegistry: ToolRegistry
   private _mcpClients: McpClient[]
   private _initialized: boolean
@@ -344,9 +344,7 @@ export class Agent implements LocalAgent, InvokableAgent {
     // Initialize hooks registry
     this._hooksRegistry = new HookRegistryImplementation()
 
-    if (config?.interventions && config.interventions.length > 0) {
-      this._interventionRegistry = new InterventionRegistry(config.interventions, this._hooksRegistry)
-    }
+    this._interventionRegistry = new InterventionRegistry(config?.interventions ?? [], this._hooksRegistry)
 
     // `undefined` (omitted) → install the default; `null`/`[]` → explicit opt-out.
     const retryStrategies: RetryStrategy[] =
--- a/strands-ts/src/index.ts
+++ b/strands-ts/src/index.ts
@@ -225,7 +225,7 @@ export type { Plugin } from './plugins/index.js'
 
 // Intervention system
 export { InterventionHandler } from './interventions/index.js'
-export { proceed, deny, guide, interrupt, transform } from './interventions/index.js'
+export { interventions } from './interventions/index.js'
 export type { OnError } from './interventions/index.js'
 
 // Retry
--- a/strands-ts/src/index.ts
+++ b/strands-ts/src/index.ts
@@ -223,6 +223,11 @@ export type {
 // Plugin system
 export type { Plugin } from './plugins/index.js'
 
+// Intervention system
+export { InterventionHandler } from './interventions/index.js'
+export { proceed, deny, guide, interrupt, transform } from './interventions/index.js'
+export type { OnError } from './interventions/index.js'
+
 // Retry
 export {
   type BackoffContext,
--- a/strands-ts/src/index.ts
+++ b/strands-ts/src/index.ts
@@ -225,7 +225,7 @@ export type { Plugin } from './plugins/index.js'
 
 // Intervention system
 export { InterventionHandler } from './interventions/index.js'
-export { interventions } from './interventions/index.js'
+export { InterventionActions } from './interventions/index.js'
 export type { OnError } from './interventions/index.js'
 
 // Retry
--- a/strands-ts/src/index.ts
+++ b/strands-ts/src/index.ts
@@ -224,8 +224,7 @@ export type {
 export type { Plugin } from './plugins/index.js'
 
 // Intervention system
-export { InterventionHandler } from './interventions/index.js'
-export { InterventionActions } from './interventions/index.js'
+export { InterventionHandler, InterventionActions } from './interventions/index.js'
 export type { OnError } from './interventions/index.js'
 
 // Retry
--- a/strands-ts/src/interventions/__tests__/registry.test.ts
+++ b/strands-ts/src/interventions/__tests__/registry.test.ts
@@ -644,7 +644,7 @@ describe('InterventionRegistry', () => {
       new InterventionRegistry([new InterruptOnInvocation()], hookRegistry)
 
       await hookRegistry.invokeCallbacks(makeBeforeInvocationEvent())
-      expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('has no effect on this event type'))
+      expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('has no effect'))
 
       warnSpy.mockRestore()
     })
--- a/strands-ts/src/interventions/__tests__/registry.test.ts
+++ b/strands-ts/src/interventions/__tests__/registry.test.ts
@@ -485,41 +485,6 @@ describe('InterventionRegistry', () => {
     })
   })
 
-  describe('initAgent', () => {
-    it('is called during initialize()', async () => {
-      const initFn = vi.fn()
-
-      class InitHandler extends InterventionHandler {
-        readonly name = 'init-handler'
-        override initAgent(): void {
-          initFn()
-        }
-      }
-
-      const registry = new InterventionRegistry([new InitHandler()], hookRegistry)
-      expect(initFn).not.toHaveBeenCalled()
-
-      await registry.initialize({} as never)
-      expect(initFn).toHaveBeenCalledOnce()
-    })
-
-    it('only runs once even if called multiple times', async () => {
-      const initFn = vi.fn()
-
-      class InitHandler extends InterventionHandler {
-        readonly name = 'init-handler'
-        override initAgent(): void {
-          initFn()
-        }
-      }
-
-      const registry = new InterventionRegistry([new InitHandler()], hookRegistry)
-      await registry.initialize({} as never)
-      await registry.initialize({} as never)
-      expect(initFn).toHaveBeenCalledOnce()
-    })
-  })
-
   describe('agent integration', () => {
     it('deny on beforeToolCall prevents tool execution', async () => {
       const { MockMessageModel } = await import('../../__fixtures__/mock-message-model.js')
@@ -662,19 +627,26 @@ describe('InterventionRegistry', () => {
       await expect(hookRegistry.invokeCallbacks(makeBeforeToolCallEvent())).rejects.toThrow('apply boom')
     })
 
-    it('initAgent throwing leaves registry uninitialized for retry', async () => {
-      class FailingInit extends InterventionHandler {
-        readonly name = 'failing-init'
-        override initAgent(): void {
-          throw new Error('init failed')
+    it('warns when action has no effect on event type', async () => {
+      const { logger } = await import('../../logging/logger.js')
+      const warnSpy = vi.spyOn(logger, 'warn')
+
+      // Force an interrupt return on beforeInvocation (which doesn't support it)
+      // via cast to test the runtime warning path
+      class InterruptOnInvocation extends InterventionHandler {
+        readonly name = 'interrupt-invocation'
+        override beforeInvocation() {
+          // Force an interrupt return via any cast to test the runtime warning
+          return { type: 'interrupt', prompt: 'test' } as never
         }
       }
 
-      const registry = new InterventionRegistry([new FailingInit()], hookRegistry)
+      new InterventionRegistry([new InterruptOnInvocation()], hookRegistry)
 
-      await expect(registry.initialize({} as never)).rejects.toThrow('init failed')
-      // Can retry — not marked as initialized
-      await expect(registry.initialize({} as never)).rejects.toThrow('init failed')
+      await hookRegistry.invokeCallbacks(makeBeforeInvocationEvent())
+      expect(warnSpy).toHaveBeenCalledWith(expect.stringContaining('has no effect on this event type'))
+
+      warnSpy.mockRestore()
     })
   })
 })
--- a/strands-ts/src/interventions/__tests__/registry.test.ts
+++ b/strands-ts/src/interventions/__tests__/registry.test.ts
@@ -1,0 +1,594 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { InterventionRegistry } from '../registry.js'
+import { InterventionHandler } from '../handler.js'
+import { HookRegistryImplementation } from '../../hooks/registry.js'
+import { Agent } from '../../agent/agent.js'
+import {
+  BeforeInvocationEvent,
+  BeforeToolCallEvent,
+  BeforeModelCallEvent,
+  AfterModelCallEvent,
+} from '../../hooks/events.js'
+import { Message, TextBlock } from '../../types/messages.js'
+import { deny } from '../actions.js'
+import type { InterventionAction, Guide, Transform, Proceed } from '../actions.js'
+
+class DenyHandler extends InterventionHandler {
+  readonly name = 'deny-handler'
+
+  override beforeToolCall(): InterventionAction {
+    return { type: 'deny', reason: 'not authorized' }
+  }
+}
+
+class GuideHandler extends InterventionHandler {
+  readonly name = 'guide-handler'
+
+  override beforeToolCall(): InterventionAction {
+    return { type: 'guide', feedback: 'add more context' }
+  }
+}
+
+class InterruptHandler extends InterventionHandler {
+  readonly name = 'interrupt-handler'
+
+  override beforeToolCall(): InterventionAction {
+    return { type: 'interrupt', prompt: 'approve this action?' }
+  }
+}
+
+class ProceedHandler extends InterventionHandler {
+  readonly name = 'proceed-handler'
+
+  override beforeToolCall(): InterventionAction {
+    return { type: 'proceed', reason: 'all good' }
+  }
+}
+
+class ThrowingHandler extends InterventionHandler {
+  readonly name = 'throwing-handler'
+  override readonly onError = 'throw' as const
+
+  override beforeToolCall(): InterventionAction {
+    throw new Error('handler crashed')
+  }
+}
+
+class ThrowingProceedHandler extends InterventionHandler {
+  readonly name = 'throwing-proceed'
+  override readonly onError = 'proceed' as const
+
+  override beforeToolCall(): InterventionAction {
+    throw new Error('handler crashed')
+  }
+}
+
+class ThrowingDenyHandler extends InterventionHandler {
+  readonly name = 'throwing-deny'
+  override readonly onError = 'deny' as const
+
+  override beforeToolCall(): InterventionAction {
+    throw new Error('handler crashed')
+  }
+}
+
+class AsyncDenyHandler extends InterventionHandler {
+  readonly name = 'async-deny'
+
+  override async beforeToolCall(): Promise<InterventionAction> {
+    return { type: 'deny', reason: 'async denial' }
+  }
+}
+
+class ModelGuideHandler extends InterventionHandler {
+  readonly name = 'model-guide'
+
+  override afterModelCall(): Proceed | Guide | Transform {
+    return { type: 'guide', feedback: 'be more specific' }
+  }
+}
+
+describe('InterventionRegistry', () => {
+  let hookRegistry: HookRegistryImplementation
+  let agent: Agent
+  const toolUse = { name: 'testTool', toolUseId: 'id-1', input: {} }
+
+  beforeEach(() => {
+    hookRegistry = new HookRegistryImplementation()
+    agent = new Agent()
+  })
+
+  function makeBeforeInvocationEvent() {
+    return new BeforeInvocationEvent({ agent, invocationState: {} })
+  }
+
+  function makeBeforeToolCallEvent() {
+    return new BeforeToolCallEvent({ agent, toolUse, tool: undefined, invocationState: {} })
+  }
+
+  function makeBeforeModelCallEvent() {
+    return new BeforeModelCallEvent({ agent, model: {} as never, invocationState: {} })
+  }
+
+  function makeAfterModelCallEvent() {
+    return new AfterModelCallEvent({
+      agent,
+      model: {} as never,
+      invocationState: {},
+      attemptCount: 0,
+      stopData: {
+        message: new Message({ role: 'assistant', content: [new TextBlock('response')] }),
+        stopReason: 'endTurn',
+      },
+    })
+  }
+
+  describe('constructor', () => {
+    it('rejects duplicate handler names', () => {
+      expect(() => new InterventionRegistry([new DenyHandler(), new DenyHandler()], hookRegistry)).toThrow(
+        "Duplicate intervention handler name: 'deny-handler'"
+      )
+    })
+
+    it('accepts handlers with unique names', () => {
+      // No throw means success
+      new InterventionRegistry([new DenyHandler(), new GuideHandler()], hookRegistry)
+    })
+  })
+
+  describe('hook registration', () => {
+    it('only registers hooks for overridden methods', async () => {
+      new InterventionRegistry([new DenyHandler()], hookRegistry)
+
+      const beforeToolEvent = makeBeforeToolCallEvent()
+      await hookRegistry.invokeCallbacks(beforeToolEvent)
+      expect(beforeToolEvent.cancel).toBe('DENIED: not authorized')
+
+      // afterModelCall should not be registered — no handler overrides it
+      const afterModelEvent = makeAfterModelCallEvent()
+      await hookRegistry.invokeCallbacks(afterModelEvent)
+      expect(afterModelEvent.retry).toBeUndefined()
+    })
+  })
+
+  describe('dispatch ordering', () => {
+    it('calls handlers in registration order', async () => {
+      const callOrder: string[] = []
+
+      class First extends InterventionHandler {
+        readonly name = 'first'
+        override beforeToolCall(): InterventionAction {
+          callOrder.push('first')
+          return { type: 'proceed' }
+        }
+      }
+      class Second extends InterventionHandler {
+        readonly name = 'second'
+        override beforeToolCall(): InterventionAction {
+          callOrder.push('second')
+          return { type: 'proceed' }
+        }
+      }
+
+      new InterventionRegistry([new First(), new Second()], hookRegistry)
+
+      await hookRegistry.invokeCallbacks(makeBeforeToolCallEvent())
+      expect(callOrder).toEqual(['first', 'second'])
+    })
+
+    it('skips handlers that do not override the method', async () => {
+      const callOrder: string[] = []
+
+      class ToolHandler extends InterventionHandler {
+        readonly name = 'tool'
+        override beforeToolCall(): InterventionAction {
+          callOrder.push('tool')
+          return { type: 'proceed' }
+        }
+      }
+      class ModelHandler extends InterventionHandler {
+        readonly name = 'model'
+        override afterModelCall(): Proceed | Guide | Transform {
+          callOrder.push('model')
+          return { type: 'proceed' }
+        }
+      }
+
+      new InterventionRegistry([new ToolHandler(), new ModelHandler()], hookRegistry)
+
+      await hookRegistry.invokeCallbacks(makeBeforeToolCallEvent())
+      expect(callOrder).toEqual(['tool'])
+    })
+  })
+
+  describe('deny', () => {
+    it('sets cancel on BeforeToolCallEvent', async () => {
+      new InterventionRegistry([new DenyHandler()], hookRegistry)
+
+      const event = makeBeforeToolCallEvent()
+      await hookRegistry.invokeCallbacks(event)
+
+      expect(event.cancel).toBe('DENIED: not authorized')
+    })
+
+    it('short-circuits — later handlers do not run', async () => {
+      const laterCalled = vi.fn()
+
+      class LaterHandler extends InterventionHandler {
+        readonly name = 'later'
+        override beforeToolCall(): InterventionAction {
+          laterCalled()
+          return { type: 'proceed' }
+        }
+      }
+
+      new InterventionRegistry([new DenyHandler(), new LaterHandler()], hookRegistry)
+
+      await hookRegistry.invokeCallbacks(makeBeforeToolCallEvent())
+      expect(laterCalled).not.toHaveBeenCalled()
+    })
+
+    it('sets cancel on BeforeInvocationEvent', async () => {
+      class InvocationDeny extends InterventionHandler {
+        readonly name = 'invocation-deny'
+        override beforeInvocation() {
+          return deny('unauthorized user')
+        }
+      }
+
+      new InterventionRegistry([new InvocationDeny()], hookRegistry)
+
+      const event = makeBeforeInvocationEvent()
+      await hookRegistry.invokeCallbacks(event)
+      expect(event.cancel).toBe('DENIED: unauthorized user')
+    })
+
+    it('sets cancel on BeforeModelCallEvent', async () => {
+      class ModelDeny extends InterventionHandler {
+        readonly name = 'model-deny'
+        override beforeModelCall() {
+          return deny('prompt injection detected')
+        }
+      }
+
+      new InterventionRegistry([new ModelDeny()], hookRegistry)
+
+      const event = makeBeforeModelCallEvent()
+      await hookRegistry.invokeCallbacks(event)
+      expect(event.cancel).toBe('DENIED: prompt injection detected')
+    })
+  })
+
+  describe('guide', () => {
+    it('sets cancel with guidance on BeforeToolCallEvent', async () => {
+      new InterventionRegistry([new GuideHandler()], hookRegistry)
+
+      const event = makeBeforeToolCallEvent()
+      await hookRegistry.invokeCallbacks(event)
+      expect(event.cancel).toBe('GUIDANCE: [guide-handler] add more context')
+    })
+
+    it('accumulates feedback from multiple handlers', async () => {
+      class SecondGuide extends InterventionHandler {
+        readonly name = 'second-guide'
+        override beforeToolCall(): InterventionAction {
+          return { type: 'guide', feedback: 'also check permissions' }
+        }
+      }
+
+      new InterventionRegistry([new GuideHandler(), new SecondGuide()], hookRegistry)
+
+      const event = makeBeforeToolCallEvent()
+      await hookRegistry.invokeCallbacks(event)
+      expect(event.cancel).toBe('GUIDANCE: [guide-handler] add more context\n[second-guide] also check permissions')
+    })
+
+    it('sets retry=true and injects guidance message on AfterModelCallEvent', async () => {
+      new InterventionRegistry([new ModelGuideHandler()], hookRegistry)
+
+      const event = makeAfterModelCallEvent()
+      const messageCountBefore = event.agent.messages.length
+      await hookRegistry.invokeCallbacks(event)
+
+      expect(event.retry).toBe(true)
+      expect(event.agent.messages).toHaveLength(messageCountBefore + 1)
+      const guidanceMessage = event.agent.messages[event.agent.messages.length - 1]!
+      expect(guidanceMessage.role).toBe('user')
+      expect(guidanceMessage.content[0]).toMatchObject({ type: 'textBlock', text: '[model-guide] be more specific' })
+    })
+  })
+
+  describe('interrupt', () => {
+    it('calls event.interrupt() on BeforeToolCallEvent', async () => {
+      new InterventionRegistry([new InterruptHandler()], hookRegistry)
+
+      const event = makeBeforeToolCallEvent()
+      await expect(hookRegistry.invokeCallbacks(event)).rejects.toThrow('Interrupt raised')
+    })
+
+    it('short-circuits — later handlers do not run', async () => {
+      const laterCalled = vi.fn()
+
+      class LaterHandler extends InterventionHandler {
+        readonly name = 'later'
+        override beforeToolCall(): InterventionAction {
+          laterCalled()
+          return { type: 'proceed' }
+        }
+      }
+
+      new InterventionRegistry([new InterruptHandler(), new LaterHandler()], hookRegistry)
+
+      await expect(hookRegistry.invokeCallbacks(makeBeforeToolCallEvent())).rejects.toThrow()
+      expect(laterCalled).not.toHaveBeenCalled()
+    })
+  })
+
+  describe('transform', () => {
+    it('calls the apply function with the event', async () => {
+      const applyFn = vi.fn()
+
+      class TransformHandler extends InterventionHandler {
+        readonly name = 'transform-handler'
+        override beforeToolCall(): InterventionAction {
+          return { type: 'transform', apply: applyFn, reason: 'sanitized input' }
+        }
+      }
+
+      new InterventionRegistry([new TransformHandler()], hookRegistry)
+
+      const event = makeBeforeToolCallEvent()
+      await hookRegistry.invokeCallbacks(event)
+      expect(applyFn).toHaveBeenCalledWith(event)
+    })
+
+    it('later handlers see the transformed state', async () => {
+      const observed: string[] = []
+
+      class Transformer extends InterventionHandler {
+        readonly name = 'transformer'
+        override beforeToolCall(): InterventionAction {
+          return {
+            type: 'transform',
+            apply: (e) => {
+              ;(e as BeforeToolCallEvent).cancel = 'transformed'
+            },
+          }
+        }
+      }
+
+      class Observer extends InterventionHandler {
+        readonly name = 'observer'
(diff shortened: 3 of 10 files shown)


strands-ts/src/agent/agent.ts
```diff
@@ -222,6 +228,7 @@ export class Agent implements LocalAgent, InvokableAgent {
 
   private readonly _hooksRegistry: HookRegistryImplementation
   private readonly _pluginRegistry: PluginRegistry
+  private readonly _interventionRegistry?: InterventionRegistry
```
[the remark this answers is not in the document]
lizradway: A public getter (agent.interventions.auditLog) will be added when the API is finalized for export with an intervention added.

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  get handlers(): readonly InterventionHandler[] {
+    return this._handlers
+  }
+
+  get auditLog(): readonly AuditRecord[] {
+    return this._auditLog
+  }
+
+  /**
```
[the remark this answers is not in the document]
lizradway: fixed audit log to return a shallow copy.

strands-ts/src/agent/agent.ts
```diff
@@ -262,6 +269,13 @@ export class Agent implements LocalAgent, InvokableAgent {
     // Initialize hooks registry
     this._hooksRegistry = new HookRegistryImplementation()
```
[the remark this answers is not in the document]
lizradway: interventions should always run before plugins.

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  /** Map a Guide action onto the mutable fields of the event. Returns false if unmapped. */
+  private _applyGuide(event: HookableEvent, action: Guide): boolean {
+    if (event instanceof BeforeToolCallEvent) {
+      event.cancel = `GUIDANCE: ${action.feedback}`
+      return true
+    } else if (event instanceof AfterModelCallEvent) {
+      event.retry = true
+      event.agent.messages.push(new Message({ role: 'user', content: [new TextBlock(action.feedback)] }))
```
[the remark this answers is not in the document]
lizradway: this is what steering does today

strands-ts/src/agent/agent.ts
```diff
@@ -377,6 +393,14 @@ export class Agent implements LocalAgent, InvokableAgent {
     return this._toolRegistry
   }
 
+  /**
+   * The intervention registry, if interventions were configured.
+   * Access the audit trail via `agent.interventions.auditLog`.
+   */
+  get interventions(): InterventionRegistry | undefined {
```
mkmeral: I'd expect `agent.interventions` to return interventions, not a registry. See tool registry and tools above

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,239 @@
...
+  get handlers(): readonly InterventionHandler[] {
+    return this._handlers
+  }
+
+  get auditLog(): readonly AuditRecord[] {
+    return [...this._auditLog]
+  }
+
+  /** Clear the audit log. Use this for long-running agents to prevent unbounded memory growth. */
+  clearAuditLog(): void {
```
mkmeral: are these logs saved anywhere?

mkmeral: what is customer experience with audit logs? why do we keep them in memory?

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,239 @@
...
+    this._auditLog.length = 0
+  }
+
+  /**
+   * Wire handlers into the hook system. Only registers callbacks for
+   * lifecycle methods where at least one handler overrides the default.
+   */
+  register(hookRegistry: HookRegistry): void {
```
mkmeral: why is this public? what is the use case? 

Also `registry.register` method in my mind takes hooks, and registers them to registry. This method signature is weird imho (like reverse)

strands-ts/src/interventions/registry.ts
mkmeral: do the customers need to handle this themselves?

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  /**
+   * Iterate handlers in registration order and resolve the winning action.
+   *
+   * - Deny / Interrupt short-circuit immediately (remaining handlers are skipped).
+   * - Guide feedback strings accumulate across handlers and are applied at the end.
+   * - If a handler throws, behavior depends on {@link InterventionHandler.onError}:
+   *   `'throw'` (default) rethrows, `'deny'` fails closed, `'proceed'` skips.
+   *
+   * Transform (modify content in-place) is planned but blocked on mutable event
```
mkmeral: > Transform (modify content in-place) is planned but blocked on mutable event * content fields

what does this mean? what is the issue?

agent-of-mkmeral (commented):
**TL;DR**: Solid primitive that survived adversarial testing (27 tests, 0 bugs found). Three non-blocking design concerns worth discussing before merge.

<details>
<summary>📝 API Bar Raising Review</summary>

### Tenet Alignment
| Tenet | Assessment |
|-------|------------|
| **Simple at any scale** | ✅ Clean base class pattern — override what you need, defaults handle the rest |
| **Extensible by design** | ✅ Per-handler `onError` policy, composable handler stacking, `clearAuditLog()` for long-running agents |
| **Composability** | ✅ Builds on hooks as prescribed by the "Hooks as Low-Level Primitives" decision record |
| **Obvious path is happy path** | ⚠️ `afterModelCall` returns full `InterventionAction` union but only `Proceed`/`Guide` are effective — runtime warning but no compile-time narrowing |
| **Accessible to humans and agents** | ✅ Good TSDoc, clear naming, example in PR description |
| **Embrace common standards** | ✅ Follows Cedar auth + guardrails patterns common in the space |

### Decision Record Alignment
- **"Hooks as Low-Level Primitives"** — ✅ InterventionHandler is exactly the high-level abstraction built on hooks that the decision prescribes
- **"Prefer Flat Namespaces"** — ✅ Everything re-exported from `src/index.ts`
- **"Pay for Play"** — ✅ Agents without `interventions:` config are completely unaffected
- **"Provide Both Low-Level and High-Level APIs"** — ✅ Low-level hooks remain for edge cases; InterventionHandler is the guided path

</details>

<details>
<summary>🔬 Adversarial Testing Report (27 tests, all passing)</summary>

| # | Category | Result | Description |
|---|----------|--------|-------------|
| 1 | Edge Case | ✅ | Empty handler list — no-op, no crash |
| 2 | Edge Case | ✅ | Empty string handler name accepted, duplicate empty strings rejected |
| 3 | Edge Case | ✅ | Special characters in handler names |
| 4 | Override Detection | ✅ | Grandchild class inheriting overrides |
| 5 | Override Detection | ✅ | Re-overriding with same return value still detected |
| 6 | Guide Accumulation | ✅ | Guide + Proceed accumulates correctly |
| 7 | Conflict Resolution | ✅ | Guide then Deny — Deny wins and short-circuits |
| 8 | Guide Accumulation | ✅ | Three guides accumulate all feedback |
| 9 | Action Applicability | ✅ | Deny on afterModelCall is no-op with warning |
| 10 | Action Applicability | ✅ | Interrupt on afterModelCall is no-op with warning |
| 11 | Async Errors | ✅ | async throw with onError=throw rethrows |
| 12 | Async Errors | ✅ | async throw with onError=deny denies |
| 13 | Async Errors | ✅ | async throw with onError=proceed continues |
| 14 | Non-Error Throws | ✅ | Throwing a string handled by onError=deny |
| 15 | Non-Error Throws | ✅ | Throwing null handled by onError=proceed |
| 16 | Non-Error Throws | ✅ | Throwing undefined handled by onError=throw |
| 17 | Message Injection | ✅ | Multiple afterModelCall guides inject ONE accumulated message |
| 18 | Double Registration | ✅ | register() twice doubles callbacks (design concern #2) |
| 19 | Event Mutation | ✅ | Handler mutating toolUse.input visible to later handlers |
| 20 | Cancel Values | ✅ | Deny with empty reason string sets cancel |
| 21 | Cancel Values | ✅ | Guide with empty feedback string |
| 22 | Audit Log | ✅ | clearAuditLog + fresh dispatches |
| 23 | Audit Log | ✅ | Timestamps are monotonically increasing |
| 24 | Agent Integration | ✅ | agent.interventions is undefined when not configured |
| 25 | Agent Integration | ✅ | agent.interventions is defined when configured |
| 26 | Agent Integration | ✅ | Empty array → interventions is undefined |
| 27 | Guide Accumulation | ✅ | afterModelCall guide accumulates into single message |

</details>

<details>
<summary>⚠️ Design Concerns (Not Bugs — Worth Discussing)</summary>

### 1. `afterModelCall` Return Type Is Too Permissive
**Severity**: Medium (API design)

`afterModelCall` returns `InterventionAction` (the full union: Proceed | Deny | Guide | Interrupt), but only `Proceed` and `Guide` are effective at runtime. Deny and Interrupt silently warn + no-op.

Per tenet #4 ("The obvious path is the happy path"), the type system should guide developers away from ineffective patterns. The earlier v3 approach with `BeforeAction`/`AfterAction` was strictly better from a "pit of success" perspective.

**Suggestion**: Narrow `afterModelCall` return type to `Proceed | Guide | Promise<Proceed | Guide>`. Non-breaking since it only restricts what handlers can return.

### 2. `register()` Has No Idempotency Guard
**Severity**: Low (robustness)

Calling `registry.register(hookRegistry)` twice doubles all callbacks — adversarial test #18 proves this. The Agent constructor only calls it once, but the method is public.

**Suggestion**: Add a `_registered` boolean flag and either throw or no-op on second call.

### 3. No Retry Budget for `afterModelCall` Guide Loop
**Severity**: Medium (production safety)

When a guide handler on `afterModelCall` sets `event.retry = true`, the agent retries. If the model consistently produces output that triggers guidance, this creates an unbounded retry loop — burning tokens and accumulating audit records.

**Suggestion**: Consider a configurable `maxRetries` on `InterventionRegistry` defaulting to something safe (e.g., 3). Can be a follow-up.

</details>

<details>
<summary>✅ What's Done Well</summary>

1. **Architecture**: Clean separation (actions.ts / handler.ts / registry.ts) with clear responsibilities
2. **Hook Integration**: Data-driven EVENT_MAP makes adding lifecycle methods trivial
3. **Error Handling**: Three-mode `onError` policy covers all deployment scenarios
4. **Audit Trail**: Copy-on-read getter, `clearAuditLog()`, timestamps — production-ready
5. **Test Coverage**: 26 author tests + 27 adversarial tests = 53 total, all passing
6. **Iterative Improvement**: 8+ review rounds that progressively improved the design

</details>

---

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,239 @@
...
+        action = this._handleError(handler, method, error)
+        if (!action) continue
+      }
+
+      this._log(handler.name, method, action)
+
+      if (action.type === 'guide') {
+        guides.push(`[${handler.name}] ${action.feedback}`)
+      } else if (apply(action, handler.name)) {
+        return
```
mkmeral: can we add more logs, maybe debug? How would a customer debug these cases? Do we have enough logs? (feel free to push back, but i feel like we are lacking a bit)

strands-ts/src/agent/agent.ts
lizradway: that makes sense

strands-ts/src/interventions/registry.ts
lizradway: just got unblocked yesterday in https://github.com/strands-agents/sdk-typescript/pull/907 and can update accordingly

strands-ts/src/agent/agent.ts
lizradway: looked at prior art and decided to update `agent.interventions` as a clean public facade over the internal registry matching `agent.tools` / `agent.toolRegistry` pattern without leaking internal machinery
```
agent.interventions?.handlers    // the handlers you configured
agent.interventions?.auditLog    // the audit trail
agent.interventions?.clearAuditLog()  // clear for long-running agents
```

thoughts?

strands-ts/src/agent/agent.ts
pgrayy, line 1: Need to look through the design but wanted to go ahead and share the work I already completed for steering ([draft](https://github.com/strands-agents/sdk-typescript/pull/943)). One difference from Python that I want to explicitly call out is that context callbacks are no longer configured in a unique manner. Instead, they are configured using the plugin interface, which offers a consistent DX. Am I still enabled to take this approach using the interventions interfaces?

strands-ts/src/interventions/registry.ts
lizradway: they're only saved to memory right now. i think that these could be exported as spans to OTEL or any third party storage as a follow up once common storage interface (https://github.com/strands-agents/sdk-python/issues/2171) lands in sandboxing.

strands-ts/src/interventions/registry.ts
lizradway: yeah fair, removed

strands-ts/src/interventions/registry.ts
lizradway: sure, added debug logs for each handler that runs.

strands-ts/src/interventions/registry.ts
mkmeral: No these shouldn't land in sandbox or agent storage. either one is ephemeral. consider agentcore deployed agents with code interpreter as sandbox. both of those will be gone, and both of those are scoped down to sessions. 

A developer who wants audits would likely either export to their 3P service or to logs. I think we should default to logs, but allow them to consume audits some other way. 

I don't like that devs would need to handle clearing the queue right now to avoid unbounded growth, devs shouldnt need to worry about this.

So question: what are the use cases for audit logs? where should it go by default? 


asked my agent to do a bit of a research fyi: https://gist.github.com/agent-of-mkmeral/3f57715d20aa04a4ac3c5c86a79c7996

strands-ts/src/agent/agent.ts
mkmeral: On this note, did you review the design @pgrayy ? I'd say it's related to middlewares that was in the state proposal. Do we cover all the use cases needed there? is it different?

strands-ts/src/interventions/handler.ts
```diff
@@ -0,0 +1,69 @@
...
+} from '../hooks/events.js'
+import type { LocalAgent } from '../types/agent.js'
+import type { InterventionAction } from './actions.js'
+
+/**
+ * What to do when a handler throws during evaluation.
+ *
+ * - `'throw'` — rethrow the error (default, safest: a broken policy check blocks execution)
+ * - `'proceed'` — log the error and continue as if the handler returned Proceed
+ * - `'deny'` — log the error and treat it as a Deny (fail-closed)
```
opieter-aws, line 17: What is the practical difference between a `throw` and a `deny`? Does `throw` support the Guide action?

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,313 @@
...
+      }
+    }
+  }
+
+  private _initialized = false
+
+  /** Initialize all handlers. Safe to call multiple times — only runs once. */
+  async initialize(agent: LocalAgent): Promise<void> {
+    if (this._initialized) return
+    this._initialized = true
```
opieter-aws: Should we set this after line 80 after the actual initialization?

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,313 @@
...
+            return
+          }
+        } catch (error) {
+          this._handleError(handler, method, error)
+        }
+      }
+    }
+
+    // Guide feedback accumulates across handlers, applied after all have run.
+    if (guides.length > 0) {
```
opieter-aws, line 232: If the function short-circuits because of a `deny`, this won't be reached. Is this intentional? If so, do we still want to log the `guide` actions that were never evaluated into the audit logs?

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,313 @@
...
+   */
+  private _handleError(handler: InterventionHandler, method: string, error: unknown): InterventionAction | undefined {
+    const errorMsg = error instanceof Error ? error.message : String(error)
+
+    if (handler.onError === 'throw') {
+      throw error
+    } else if (handler.onError === 'deny') {
+      return { type: 'deny', reason: `Handler threw: ${errorMsg}` }
+    } else {
+      this._log(handler.name, method, { type: 'proceed', reason: `Handler threw: ${errorMsg}` })
```
opieter-aws: It's confusing that `proceed` is both a type of  `OnError` and `InterventionAction`. Based on the log line here, it's ambiguous for a reader if the handler returned a `Proceed`, or the handler threw but the `onError` behaviour let it proceed. Does it make sense to either log an `Error` type, or make this more obvious from the `reason` prefix?

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,313 @@
...
+        guides.push(`[${handler.name}] ${action.feedback}`)
+      } else {
+        try {
+          if (apply(action, handler.name)) {
+            logger.debug(`handler=<${handler.name}>, event=<${method}> | short-circuited`)
+            return
+          }
+        } catch (error) {
+          this._handleError(handler, method, error)
+        }
```
opieter-aws, line 226: If the `apply` throws, and `onError` is set to `Deny`, it won't be applied nor logged, correct?

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,313 @@
...
+            event.cancel = `GUIDANCE: ${action.feedback}`
+            return false
+          case 'transform':
+            action.apply(event)
+            return false
+          case 'proceed':
+            return false
+          default:
+            logger.warn(
+              `handler=<${handlerName}>, event=<${method}> | ${(action as InterventionAction).type} has no effect on this event type`
```
opieter-aws: nit: is `action as InterventionAction` necessary here given `action` is already typed?

opieter-aws: ditto for warning below

strands-ts/src/interventions/registry.ts
lizradway: no this would not be reached. i don't think so? i would want the audit log to show what has actually been executed. guide wouldn't even be reached to accumulate feedback if a prior intervention hits a deny.

strands-ts/src/interventions/registry.ts
lizradway: yeah that is a fair call out. i can reframe the error types to have different namings than the interventions -- i honestly do not even know if that is necessary or if i was over-engineering it. any thoughts on that?

strands-ts/src/interventions/registry.ts
lizradway: correct -- it will be treated as an explicit InterventionHandler `deny`

strands-ts/src/interventions/registry.ts
lizradway: makes sense, updated to switch by `action.type` instead

strands-ts/src/interventions/registry.ts
opieter-aws: I do think that different namings would make it easier to distinguish the code here, and read the audit logs

strands-ts/src/interventions/registry.ts
opieter-aws: Not sure if I understand you correctly. When `onError` is `deny`, `_handleError` returns `{ type: 'deny', reason: '...' }` but nothing calls `apply()` on it or logs it. The deny is silently dropped and execution continues to the next handler.

strands-ts/src/interventions/registry.ts
opieter-aws: That's also my understanding, that this would not be reached. But the action is being logged in line 127 already. So we'd need to update the logic here to achieve no logging for non-executed `guide` actions.

lizradway: oh nice! good catch will update thanks

strands-ts/src/interventions/registry.ts
lizradway: yea def agree. do you think the error actions are necessary to begin with? not sure if that is over engineering?

strands-ts/src/interventions/registry.ts
lizradway: ah i see, updating.

strands-ts/src/interventions/registry.ts
lizradway: moved so that this can be reinitiated if init throws

strands-ts/src/interventions/handler.ts
lizradway: - throw actually bubbles up through to the agent loop whereas deny rejects the too/model call and is logged in audit call as actual `Deny` action
- nope

strands-ts/src/interventions/registry.ts
opieter-aws: True, we can also just use `type: Error` here to simplify?

lizradway: i like that. adding separate log path for errors to not disguise them as default  deny/proceed types.

agent-of-mkmeral:
## DevX Discussion: Is class-based `InterventionHandler` the best surface for defining interventions?

The architecture here is solid — typed decisions, short-circuiting, unified audit, first-class `Agent({ interventions })` parameter. The question is whether the **class-only interface** is the optimal DevX for the TypeScript ecosystem, or whether we should also offer functional sugar for simple cases.

<details>
<summary><strong>6 DevX Approaches Compared</strong></summary>

### Approach 1: Class-Based InterventionHandler (This PR — Current)

```typescript
class CedarAuth extends InterventionHandler {
  readonly name = 'cedar-auth'
  
  override beforeToolCall(event: BeforeToolCallEvent): InterventionAction {
    if (!this.isAuthorized(event)) {
      return { type: 'deny', reason: 'Not authorized' }
    }
    return { type: 'proceed' }
  }
}

const agent = new Agent({
  interventions: [new CedarAuth(), new ToneSteering()],
})
```

**Pros:** Encapsulates state, `initAgent()` lifecycle, type-safe methods, framework detects overrides, cross-SDK parity with Python  
**Cons:** High boilerplate for simple checks, Java-ish in TS ecosystem, single inheritance limitation

---

### Approach 2: Functional/Callback Pattern

```typescript
const cedarAuth = intervention('cedar-auth', {
  beforeToolCall: (event) => {
    if (!isAuthorized(event)) return deny('Not authorized')
    return proceed()
  },
  onError: 'deny',
})

const agent = new Agent({
  interventions: [cedarAuth, toneSteering],
})
```

**Pros:** Zero boilerplate, idiomatic TS/JS, trivially testable, composable with higher-order functions  
**Cons:** Harder to hold complex state, no `initAgent` without extra ceremony, less IDE self-documenting

---

### Approach 3: Middleware Pattern (#944)

```typescript
const cedarMiddleware: Middleware = {
  wrap(step) {
    return {
      async *stream(options) {
        if (step.type === 'toolCall' && !isAuthorized(options)) {
          throw new DenyError('Not authorized')
        }
        yield* step.stream(options)
      }
    }
  }
}

const agent = new Agent({
  middleware: [cedarMiddleware],
})
```

**Pros:** Most familiar to web devs (Express/Koa/Next.js), direct execution control, unified concept for all cross-cutting concerns  
**Cons:** Requires state machine refactor, no built-in conflict resolution, implicit short-circuiting, unfamiliar for policy devs (Cedar/OPA think allow/deny not wrap)

---

### Approach 4: Hook-Based Plugin Extension (Status Quo)

```typescript
class CedarPlugin implements Plugin {
  name = 'cedar-auth'
  initAgent(agent) {
    agent.hooks.addCallback(BeforeToolCallEvent, (event) => {
      if (!this.isAuthorized(event)) event.cancelTool = 'Not authorized'
    })
  }
}
```

**Pros:** No new concept, works today  
**Cons:** Can't distinguish deny/guide/interrupt, no short-circuiting, no audit trail, ordering undefined — **these are exactly the problems that justify the new primitive**

---

### Approach 5: Declarative Config Pattern

```typescript
const agent = new Agent({
  interventions: {
    authorization: { engine: 'cedar', policies: './policies' },
    guardrails: { engine: 'bedrock', guardrailId: 'abc123' },
    steering: { engine: 'llm', systemPrompt: 'Ensure cheerful tone' },
  },
})
```

**Pros:** Zero code, config-driven, easiest for non-expert devs  
**Cons:** Not extensible for custom logic, limited to known engines — only works for vended handlers

---

### Approach 6: Hybrid — Functional + Class Escape Hatch

```typescript
// Simple case: functional helper
const blockPII = beforeToolCall('pii-filter', (event) => {
  if (hasPII(event.toolUse.input)) return deny('PII detected')
  return proceed()
})

// Complex case: class (full power)
class CedarAuth extends InterventionHandler {
  readonly name = 'cedar-auth'
  private policySet: PolicySet
  async initAgent(agent) { this.policySet = await loadPolicies() }
  override beforeToolCall(event) { /* ... */ }
}

// Vended case: factory
const guardrails = bedrockGuardrails({ guardrailId: 'abc123' })

const agent = new Agent({
  interventions: [blockPII, new CedarAuth(), guardrails],
})
```

**Pros:** Right tool for the job (simple→function, complex→class, vended→factory), lowest ceremony for common cases, full power when needed  
**Cons:** Multiple ways to do same thing (documentation burden)

</details>

<details>
<summary><strong>Evaluation Matrix</strong></summary>

| Criterion | Class (This PR) | Functional | Middleware | Hooks | Declarative | Hybrid |
|-----------|:-:|:-:|:-:|:-:|:-:|:-:|
| Simple case boilerplate | ⚠️ High | ✅ Low | ⚠️ Medium | ✅ Low | ✅ Zero | ✅ Low |
| Complex case (state, init) | ✅ Natural | ⚠️ Awkward | ⚠️ Awkward | ✅ Natural | ❌ Impossible | ✅ Natural |
| Type safety | ✅ Strong | ✅ Strong | ⚠️ Medium | ❌ Weak | ⚠️ Schema | ✅ Strong |
| Conflict resolution | ✅ Built-in | ✅ Built-in | ❌ Manual | ❌ None | ✅ Built-in | ✅ Built-in |
| Short-circuiting | ✅ Built-in | ✅ Built-in | ⚠️ Implicit | ❌ None | ✅ Built-in | ✅ Built-in |
| Audit trail | ✅ Built-in | ✅ Built-in | ❌ Manual | ❌ None | ✅ Built-in | ✅ Built-in |
| Familiarity (TS devs) | ⚠️ Java-ish | ✅ Idiomatic | ✅ Very familiar | ✅ Familiar | ✅ Easy | ✅ Good |
| Testability | ⚠️ Medium | ✅ Trivial | ⚠️ Medium | ⚠️ Medium | ⚠️ Integration | ✅ Mix |
| Cross-SDK parity | ✅ Easy | ⚠️ Different | ⚠️ Different | ✅ Already | ⚠️ Different | ⚠️ Complex |

</details>

<details>
<summary><strong>Key Tensions & Open Questions</strong></summary>

### 1. TS idiom vs cross-SDK parity
Functional patterns are more idiomatic in TypeScript. Classes are more natural in Python. This PR maintains parity but the class ceremony feels Java-ish in the JS ecosystem.

### 2. Concept proliferation
After this ships, developers have **4 interception patterns**: hooks (low-level), plugins (lifecycle extensions), interventions (typed decisions), and potentially middleware (#944 from the state machine design). The design doc's "Why Not Just Hooks?" section justifies interventions well, but the "when do I use which?" question becomes real.

### 3. Simple case tax
A 5-line authorization check becomes a 12-line class. For a TypeScript developer writing `if (!authorized) return deny(...)`, the class ceremony feels like paying a Java tax.

### 4. Middleware overlap
#944 (by `pgrayy`) proposes middleware for similar use cases. The comment below where `pgrayy` asks "Am I still enabled to take this approach using the interventions interfaces?" reveals the tension — two parallel proposals for the same space. Should middleware be the execution wrapper and interventions be the decision layer? Or should one subsume the other?

### 5. Steering: intervention handler or plugin?
PR #943 implements steering as a Plugin. The design doc says it should be an InterventionHandler. Which is the canonical path?

</details>

<details>
<summary><strong>Recommendation</strong></summary>

**The architecture in this PR is correct** — merge the class-based `InterventionHandler` as the foundational layer. Then consider adding functional sugar as a non-breaking follow-up:

```typescript
// Potential follow-up: intervention() factory for simple cases
import { intervention, deny, proceed } from '@strands-agents/sdk'

const blockPII = intervention('pii-filter', {
  beforeToolCall: (event) => {
    if (hasPII(event.toolUse.input)) return deny('PII detected')
    return proceed()
  },
})
```

This is additive — doesn't change the PR's implementation, just wraps it for ergonomics. The `intervention()` factory would internally create an `InterventionHandler` subclass.

**Decisions needed:**
1. Ship class-only now, add functional sugar later? Or design the full DevX surface upfront?
2. Interventions vs Middleware (#944) — coexist (different layers) or converge?
3. Steering canonical path: intervention handler (design doc) or plugin (PR #943)?

</details>

---

strands-ts/src/interventions/handler.ts
```diff
@@ -0,0 +1,69 @@
...
+ *   override beforeToolCall(event: BeforeToolCallEvent): InterventionAction {
+ *     if (!this.isAuthorized(event)) {
+ *       return { type: 'deny', reason: 'User not authorized for this tool' }
+ *     }
+ *     return { type: 'proceed' }
+ *   }
+ * }
+ * ```
+ */
+export abstract class InterventionHandler {
```
notowen333, line 42: Could/should this be a `Plugin`? It appears to already satisfy the interface

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,331 @@
...
+    return [...this._auditLog]
+  }
+
+  /** Clear the audit log. Use this for long-running agents to prevent unbounded memory growth. */
+  clearAuditLog(): void {
+    this._auditLog.length = 0
+  }
+
+  /**
+   * Iterate handlers in registration order and resolve the winning action.
```
notowen333, line 180: I have the same idea for Plugins...

Should we give an interface to declare a specific ordering instead of relying on registration order? Registration order creates more possibility for surprise IMO

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,331 @@
...
+   *
+   * - Deny / Interrupt short-circuit immediately (remaining handlers are skipped).
+   * - Guide feedback strings accumulate across handlers and are applied at the end.
+   * - Transform is applied in-place so later handlers see the mutation.
+   * - If a handler throws, behavior depends on {@link InterventionHandler.onError}:
+   *   `'throw'` (default) rethrows, `'deny'` fails closed, `'proceed'` skips.
+   */
+  private async _dispatch(event: HookableEvent, method: LifecycleMethod): Promise<void> {
```
notowen333: nit: rename to _evaluate()?

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,331 @@
...
+        // Safe: _registerHooks() only wires each method to its matching event type,
+        // so the event is always the correct type for the method being called.
+        action = await handler[method](event as never)
+      } catch (error) {
+        action = this._handleError(handler, method, error)
+        if (!action) continue
+        fromError = true
+      }
+
+      logger.debug(`handler=<${handler.name}>, event=<${method}> | returned ${action.type}`)
```
notowen333, line 209: Maybe could make sense to allow the instantiator to configure logging? I think this handler could get noisy

strands-ts/src/agent/agent.ts
zastrowm, line 1: PR comment: can the PR description be updated with all of the new types/apis that are being added?  That helps api bar-raising and clarifies the overall implementation

strands-ts/src/interventions/handler.ts
```diff
@@ -0,0 +1,75 @@
...
+  abstract readonly name: string
+
+  /** What to do when this handler throws. Defaults to 'throw'. */
+  readonly onError: OnError = 'throw'
+
+  /** Called once during agent initialization. Use to register context-gathering hooks. */
+  initAgent(_agent: LocalAgent): void | Promise<void> {}
```
mkmeral: what's the use case here? how would devs use it? can we achieve it another way?

I know we have this pattern in the sdk, but i don't like agent subclassess using agent object. it couples them a lot, and creates footguns (when you pass same handler to different objects)

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,286 @@
...
+    }
+    this._handlers = handlers
+    this._registerHooks(hookRegistry)
+  }
+
+  private _registerHooks(hookRegistry: HookRegistry): void {
+    // Before*: interventions run last (closest to execution)
+    for (const [method, eventType] of BEFORE_EVENTS) {
+      if (this._handlers.some((h) => h[method] !== InterventionHandler.prototype[method])) {
+        hookRegistry.addCallback(eventType, (event) => this._dispatch(event, method), { order: HookOrder.SDK_LAST })
```
mkmeral: are we sure interventions will be first and last? why don't we have `HookOrder.INTERVENTIONS` ?

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,286 @@
...
+    for (const [method, eventType] of AFTER_EVENTS) {
+      if (this._handlers.some((h) => h[method] !== InterventionHandler.prototype[method])) {
+        hookRegistry.addCallback(eventType, (event) => this._dispatch(event, method), { order: HookOrder.SDK_FIRST })
+      }
+    }
+  }
+
+  private _initialized = false
+
+  /** Initialize all handlers. Safe to call multiple times — only runs once. */
```
mkmeral: nit: race conditions? can i call it before first call finished? i think i can

strands-ts/src/interventions/registry.ts
lizradway: They shouldn't be absolute first. Users should be able to set custom logic to go first if they want/we should be able to expand new first options as well if we decide.

im really against hardcoding specific orders as default into the sdk. we can always add this if we need later on, but it scales really poorly and makes understanding sdk order very confusing. currently we only need interventions to run before sdk, not at any other specific point.

strands-ts/src/interventions/registry.ts
lizradway: i suppose a user could call this outside the agent loop. can add defensive check.

strands-ts/src/interventions/registry.ts
mkmeral: sure but my question is not about the user, it's about the SDK. You are saying this is the first event that we will send from the SDK. are we sure that will be the case? do we want to box ourselves in right now? why not just do:

```python
HookOrder.INTERVENTIONS = HookOrder.SDK_FIRST + 10
```

this would not box us (ie we can add more hooks), it also gives customers ability to add hooks before/after specific events (i.e. interrupts), in case we decide to change them later on

lizradway: we can always add this (`HookOrder.INTERVENTIONS = HookOrder.SDK_FIRST + 10`) later if we add alot of internal SDK ordering that makes this necessary.
  
  but imo i do not want to bake in such rigid fragile ordering into the sdk unless absolutely necessary

mkmeral: > we can always add this (HookOrder.INTERVENTIONS = HookOrder.SDK_FIRST + 10) later

that changes when interventions get invoked though. Someone who depends on interventions being first (and they want to run their own hook, right after) would set `HookOrder.SDK_FIRST + 1` today. And that would get broken when we want to add it. 

I am just not sure why we are explicitly making this first or last (when we don't need to)

strands-ts/src/agent/agent.ts
```diff
@@ -334,6 +341,10 @@ export class Agent implements LocalAgent, InvokableAgent {
     // Initialize hooks registry
     this._hooksRegistry = new HookRegistryImplementation()
 
+    if (config?.interventions && config.interventions.length > 0) {
+      this._interventionRegistry = new InterventionRegistry(config.interventions, this._hooksRegistry)
```
zastrowm: Any reason not to initialize with zero and avoid the conditionals?

strands-ts/src/interventions/handler.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+export abstract class InterventionHandler {
+  abstract readonly name: string
+
+  /** What to do when this handler throws. Defaults to 'throw'. */
+  readonly onError: OnError = 'throw'
+
+  beforeInvocation(
+    _event: BeforeInvocationEvent
+  ): (Proceed | Deny | Guide | Transform) | Promise<Proceed | Deny | Guide | Transform> {
```
zastrowm: Can we extract a utility type to turn this into:

```
SyncOrAsync<Proceed | Deny | Guide | Transform>
```

But with a good name

strands-ts/src/interventions/registry.ts
zastrowm: Agreed, let's add a named constant for interventions

lizradway: any preference between `INTERVENTIONS_BEFORE: 90, INTERVENTIONS_AFTER: -90` on hook order or flipping the order internally in interventions on after events: `{ order: -HookOrder.INTERVENTIONS }`?

I think flipping the order internally is a little misleading to the user's own set orders, but maintaining duplicate interventions values is annoying.

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,274 @@
...
+  /**
+   * Iterate handlers in registration order and resolve the winning action.
+   *
+   * - Deny / Interrupt short-circuit immediately (remaining handlers are skipped).
+   * - Guide feedback strings accumulate across handlers and are applied at the end.
+   * - Transform is applied in-place so later handlers see the mutation.
+   * - If a handler throws, behavior depends on {@link InterventionHandler.onError}:
+   *   `'throw'` (default) rethrows, `'deny'` fails closed, `'proceed'` skips.
+   */
+  private async _dispatch(event: HookableEvent, method: LifecycleMethod): Promise<void> {
```
zastrowm: I think this would be more readable if we ended up doing explicit hook subscription.

IMHO:

ubscribe to all possible events - those that don't have handlers are no-ops. Each hook subscription is effectively something like this:

```ts
agent.addCallback(BeforeToolCallEvent, event => this.onBeforeToolCall)
...
private onBeforeToolCall() {
   this._dispatch(..., (action, handlerName) => {
        const actionType = action.type
        switch (actionType) {
          case 'deny':
            event.cancel = `DENIED: ${action.reason}`
            return true
          case 'interrupt':
            event.interrupt({ name: handlerName, reason: action.prompt })
            return true
          case 'guide':
            event.cancel = `GUIDANCE: ${action.feedback}`
            return false
          case 'transform':
            action.apply(event)
            return false
          case 'proceed':
            return false
          default:
            logger.warn(`handler=<${handlerName}>, event=<${method}> | ${actionType} has no effect on this event type`)
            return false
        }
      }
```

which gets rid of the switch statement before and makes it's more explicit what each handler is doing in the entirety

strands-ts/src/index.ts
```diff
@@ -218,6 +218,11 @@ export type {
 // Plugin system
 export type { Plugin } from './plugins/index.js'
 
+// Intervention system
+export { InterventionHandler } from './interventions/index.js'
+export { proceed, deny, guide, interrupt, transform } from './interventions/index.js'
```
zastrowm: What about namespacing these?

```
export const interventions = {
  proceed,
  deny,
  guide,
  interrupt,
  transform,
}
```

Top-level only for interventions seems to elevate it a lot; thoughts?


Or something like

```ts
return InterventionHandler.proceed()
```
?

strands-ts/src/agent/agent.ts
lizradway: then intervention registry always exists, which i guess makes sense to have an empty array for it if uninitialized. updating :)

strands-ts/src/interventions/handler.ts
lizradway: thoughts on `Awaitable`? `MaybeAsync`?

zastrowm: Agent says Awaitable:

> 1. Awaitable<T> ⭐ recommended
type Awaitable<T> = T | Promise<T>
This is the most widely used name in the ecosystem. Vue, Vite, TypeScript-ESLint, and Vitest all use Awaitable<T>. It reads naturally: "the return type is an awaitable action" — meaning you can await it but don't have to. Short, idiomatic, and immediately understood by most TS developers.

which works for me

strands-ts/src/interventions/__tests__/registry.test.ts
```diff
@@ -0,0 +1,652 @@
...
+
+      // afterModelCall should not be registered — no handler overrides it
+      const afterModelEvent = makeAfterModelCallEvent()
+      await hookRegistry.invokeCallbacks(afterModelEvent)
+      expect(afterModelEvent.retry).toBeUndefined()
+    })
+  })
+
+  describe('dispatch ordering', () => {
+    it('calls handlers in registration order', async () => {
```
mkmeral, line 155: here's a question. in the future, would we have plugins that provide interrupts, and in that case when would they run? 

(not blocking)

strands-ts/src/interventions/actions.ts
```diff
@@ -0,0 +1,142 @@
...
+ * ```typescript
+ * override afterModelCall(event: AfterModelCallEvent): InterventionAction {
+ *   if (this.isTooVague(event.stopData?.message)) {
+ *     return { type: 'guide', feedback: 'Be more specific in your response.' }
+ *   }
+ *   return { type: 'proceed' }
+ * }
+ * ```
+ */
+export type Guide = { type: 'guide'; feedback: string; reason?: string }
```
mkmeral, line 66: nit: feedback/reason confused me regarding what would actually be shown to the LLM, we should add it to docs + update jsdoc

mkmeral (dismissed):

strands-ts/src/interventions/__tests__/registry.test.ts
lizradway: depends if it is before/after hook.


`plugins -> interventions -> model/tool -> interventions -> plugins` is the way I like to think about it.

so it wouldn't interfere with the intervention either way.

mkmeral (approved):

opieter-aws (approved):

strands-ts/src/interventions/registry.ts
```diff
@@ -0,0 +1,250 @@
...
+      if (seen.has(h.name)) {
+        throw new Error(`Duplicate intervention handler name: '${h.name}'`)
+      }
+      seen.add(h.name)
+    }
+    this._handlers = handlers
+    this._registerHooks(hookRegistry)
+  }
+
+  private _registerHooks(hookRegistry: HookRegistry): void {
```
notowen333, line 42: comment explaining the `HookOrder` mechanics here would be helpful

notowen333 (approved):

strands-ts/src/interventions/registry.ts
lizradway: can add in follow up where i add HITL :)

PR #886, opened 2026-04-23
feat: override countTokens with native token counting for supported providers

opieter-aws:
## Description                                                                                                                                                                                          
                                                                                                                                                                                                        
Override `countTokens()` in provider subclasses that expose native token counting APIs, improving accuracy over the character-based heuristic added in #853. The method signature is unchanged — native 
counting is transparent to callers.                                                                                                                                                                     

Each provider reuses its existing request formatting logic, calls the native counting API, and falls back to `super.countTokens()` with a warning log on any failure:                                   

- **Bedrock** — `CountTokensCommand` via AWS SDK (same format as Converse API)                                                                                                                          
- **Anthropic** — `client.messages.countTokens()` (same format as Messages API)                                                                                                                         
- **Google Gemini** — `client.models.countTokens()` for message contents; system instructions and tools are estimated via the base class heuristic because the Gemini API (non-Vertex backend) does not support `systemInstruction`/`tools` in `CountTokensConfig`                                                                                                                                              

Fields irrelevant to token counting are stripped from the request before calling the counting API.                                     
                                                                                                                                                                                                        
### TypeScript port of                                                                                                                                                                                  
                                                                                                                                                                                                        
- Python PR: https://github.com/strands-agents/sdk-python/pull/2189                                                                                                                                     
                                                                                                                                                                                                        
## Related Issues                                                                                                                                                                                       
                                                                                                                                                                                                        
https://github.com/strands-agents/sdk-python/issues/2179
                                                                                                                                                                                                        
## Documentation PR                                                                                                                                                                                     
                                                                                                                                                                                                        
Documentation will be updated alongside proactive compression work
                                                                                                                                                                                                        
## Type of Change                                                                                                                                                                                       
                                                                                                                                                                                                        
New feature
                                                                                                                                                                                                        
## Testing                                                                                                                                                                                              
                                                                                                                                                                                                        
- [x] I ran `npm run check`                                                                                                                                                                             
- 18 new unit tests (6 per provider) covering: success, system prompt, tool specs, inference config stripping, API error fallback, generic exception fallback                                           
- 6 new integration tests (2 per provider) covering: messages-only counting, tools+system prompt returns more tokens than without                                                                       
- All 402 existing model unit tests continue to pass                                                                                                                                                    

## Checklist                                                                                                                                                                                            

### DIFF
--- a/strands-ts/src/errors.ts
+++ b/strands-ts/src/errors.ts
@@ -173,6 +173,19 @@ export class SessionError extends Error {
 }
 
 /**
+ * Thrown when a model provider's native token counting API fails.
+ *
+ * This error is used as internal control flow within provider `countTokens()` overrides.
+ * When caught, the provider falls back to the base class heuristic estimation.
+ */
+export class ProviderTokenCountError extends ModelError {
+  constructor(message: string, options?: { cause?: unknown }) {
+    super(message, options)
+    this.name = 'ProviderTokenCountError'
+  }
+}
+
+/**
  * Thrown when a tool fails validation during registration.
  */
 export class ToolValidationError extends Error {
--- a/strands-ts/src/index.ts
+++ b/strands-ts/src/index.ts
@@ -27,6 +27,7 @@ export {
   JsonValidationError,
   ConcurrentInvocationError,
   ModelThrottledError,
+  ProviderTokenCountError,
   ToolValidationError,
   StructuredOutputError,
 } from './errors.js'
--- a/strands-ts/src/models/__tests__/anthropic.test.ts
+++ b/strands-ts/src/models/__tests__/anthropic.test.ts
@@ -22,6 +22,7 @@ function createMockClient(streamGenerator: () => AsyncGenerator<unknown>): Anthr
   return {
     messages: {
       stream: vi.fn(() => streamGenerator()),
+      countTokens: vi.fn(),
     },
   } as unknown as Anthropic
 }
@@ -32,6 +33,7 @@ vi.mock('@anthropic-ai/sdk', () => {
     return {
       messages: {
         stream: vi.fn(),
+        countTokens: vi.fn(),
       },
     }
   })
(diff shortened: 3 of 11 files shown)


strands-ts/src/models/bedrock.ts
```diff
@@ -460,6 +467,43 @@ export class BedrockModel extends Model<BedrockModelConfig> {
...
+
+      const response = await this._client.send(
+        new CountTokensCommand({
+          modelId: this._config.modelId,
+          input: { converse: converseInput },
+        })
+      )
+
+      if (response.inputTokens == null) {
+        throw new Error('Bedrock CountTokens returned undefined for inputTokens')
```
JackYPCOnline: Worth adding dedicated Exception for countToken usage : https://github.com/strands-agents/sdk-typescript/blob/main/strands-py/strands/types/exceptions.py

opieter-aws: The throw here is just an internal control flow. It's caught by the catch block in the same method to trigger the token estimation fallback, and never propagates to the caller. Adding a dedicated exception class would add public API surface for something that's never exposed to users. Happy to add one if you feel strongly, but wanted to flag the tradeoff.

JackYPCOnline: Adding exception doesn't add public API surface I guess? current `catch` swallow any error, I suggesting adding one since we treat countToken as a  feature module, you can reuse it elsewhere. Thus, it keeps consistency.
```
 try {                                                                                                                                
    // ...        
    if (response.inputTokens == null) {                                                                                                
      throw new CountTokensException('Bedrock CountTokens returned undefined')                                                             
    }                                                                                                                                  
    // ...                                                                                                                             
  } catch (e) {                                                                                                                        
    if (e instanceof CountTokensException) {                                                                                               
      // expected: fall back to estimation                                                                                             
    } else {                                                                                                                           
      throw e  // real bug — don't swallow it                                                                                          
    }                                                                                                                                  
  }                  
```

opieter-aws: If it's a new exported class, that would add API surface (exceptions are also documented in https://strandsagents.com/docs/api/typescript/). 

Regardless of adding an exception, I don't think we should change the catching logic. Token counting is an optimization and not a critical path, so we should always degrade gracefully and fall back to token estimation regardless of the error thrown by native token counting.

JackYPCOnline: yep, it is fine to adding Error/Exception for readability, but still, it depends on the question if tokenCounting system will throw excetion elsewhere. User can know what happens easily.
I agree with second point.

strands-ts/src/models/__tests__/anthropic.test.ts
```diff
@@ -640,4 +642,90 @@ describe('AnthropicModel', () => {
...
+
+    it('should include tool specs in request', async () => {
+      const mockCountTokens = vi.fn(async () => ({ input_tokens: 100 }))
+      const client = createCountTokensClient(mockCountTokens)
+      const model = new AnthropicModel({ client, modelId: 'claude-sonnet-4-6' })
+
+      const result = await model.countTokens(messages, { toolSpecs })
+
+      expect(result).toBe(100)
+      expect(mockCountTokens).toHaveBeenCalledWith(expect.objectContaining({ tools: expect.any(Array) }))
```
lizradway: nit/nonblocking: bumping https://github.com/strands-agents/sdk-typescript/blob/main/docs/TESTING.md#object-assertion-best-practices for this if you can assert the entire object instead with a fixture if needed


in general for all the tests in this pr

lizradway (dismissed):

strands-ts/src/errors.ts
```diff
@@ -172,6 +172,19 @@ export class SessionError extends Error {
   }
 }
 
+/**
+ * Thrown when a model provider's native token counting API fails.
+ *
+ * This error is used as internal control flow within provider `countTokens()` overrides.
+ * When caught, the provider falls back to the base class heuristic estimation.
+ */
+export class ProviderTokenCountError extends ModelError {
```
lizradway, line 181: should we introduce something similar in python? i think this is nice

JackYPCOnline (dismissed):

strands-ts/src/errors.ts
opieter-aws: Yes I will update Python implementation once this is merged

strands-ts/src/models/__tests__/anthropic.test.ts
opieter-aws: Thanks! I added it after your last comment, which was after the implementation of these tests

mkmeral (dismissed):

lizradway (approved):

strands-ts/src/index.ts
```diff
@@ -27,6 +27,7 @@ export {
   JsonValidationError,
   ConcurrentInvocationError,
   ModelThrottledError,
+  ProviderTokenCountError,
```
lizradway, line 30: does this need to be exported? will this error ever be user facing?

opieter-aws: Good catch, indeed only internal! #937

PR #888, opened 2026-04-23
feat: add DefaultModelRetryStrategy, ModelRetryStrategy, and BackoffStrategy

### LINKED ISSUE #472: [V1] Agent - Retry Strategy

notowen333:
## Description

Adds automatic retry for failed model calls, default-on. Every provider in this SDK already normalizes rate-limit signals into `ModelThrottledError` (Anthropic 429, OpenAI 429 / `rate_limit_exceeded`, Bedrock `throttlingException`, Vercel 429, Google `RESOURCE_EXHAUSTED` / `UNAVAILABLE`), but the agent loop re-throws and users are left to wire their own `AfterModelCallEvent` hook with their own backoff math. The Python SDK ships retries by default; TS did not. This PR closes that gap.

The work is split into four reusable pieces:

1. **`BackoffStrategy` interface** with `ConstantBackoff`, `LinearBackoff`, and `ExponentialBackoff` implementations — pure delay math, no retry policy. Supports four jitter modes (`none`, `full`, `equal`, `decorrelated`). `BackoffContext` is documented as an additive-only contract so new signals (e.g. `retryAfterMs`) can be added later without breaking existing strategies.

2. **`ModelRetryStrategy`** — an abstract `Plugin` base class, one-per-retry-kind. Owns `AfterModelCallEvent` hook registration and a single-agent attachment guard (instances carry per-turn state, so sharing across agents would let calls trample each other). Declares a single abstract method `retryModel(event)` that subclasses implement. No `reset()` contract — state scope is owned entirely by the subclass.

3. **`DefaultModelRetryStrategy extends ModelRetryStrategy`** — the default concrete strategy. Implements `retryModel` to drive the existing `event.retry` flow in `_invokeModel`, reading the attempt counter off `AfterModelCallEvent.attemptCount` (new — see below). Retries are gated by a private `_isRetryable(err)` check (today: `ModelThrottledError` only) and dispatched through the single configured backoff. Both are the forward-compatible extension points for adding transient-error classification or per-error-class backoff selection without a public API change. Per-turn backoff state (`_firstFailureAt`, `_lastDelayMs`) is cleared on `event.attemptCount === 1`, which the agent loop guarantees on every fresh turn.

4. **`AfterModelCallEvent.attemptCount`** — new readonly 1-indexed field on the event. The first call in a turn is `1`; each subsequent retry increments by one. The agent loop in `_invokeModel` is the single owner of this counter, so every `ModelRetryStrategy` subclass sees a consistent attempt number regardless of how many strategies are registered. Retry strategies can rely on `attemptCount === 1` as a turn-boundary marker.

## Pattern Going Forward

Each new type of retry will get its own abstract base class. The agent instance itself takes on a list of retry strategies. The type retry strategy gets an additional `|` union member for each new type. So, we get additive, combinable, and fully customizable retry strategies.

### Public API

One new field on `AgentConfig`:

```typescript
// Before: throttling failures propagate unless the user registers a retry hook.
const agent = new Agent({ model })

// After: default-on retry with sensible exponential backoff
// (6 attempts, baseMs=4s, maxMs=240s, full jitter).
const agent = new Agent({ model })

// Tune the default strategy
const agent = new Agent({
  model,
  retryStrategy: new DefaultModelRetryStrategy({
    maxAttempts: 4,
    backoff: new ExponentialBackoff({ baseMs: 500, maxMs: 60_000, jitter: 'full' }),
  }),
})

// Opt out entirely — no retry plugin is registered. (`null` signifies no strategy)
const agent = new Agent({ model, retryStrategy: null })
```

The field is typed `retryStrategy?: RetryStrategy | RetryStrategy[] | null`. Today `RetryStrategy` is a type alias for `ModelRetryStrategy` — a union that will grow arms (e.g. `| ToolRetryStrategy`) as new retry kinds are added.

Four-state on purpose:
- **Omitted** installs the default `DefaultModelRetryStrategy`.
- **A single `RetryStrategy` instance** registers that instance.
- **An array** registers all strategies in the given order. Passing two instances of the same concrete class logs a warning via `warnOnDuplicateRetryStrategyTypes` — they would collide on `plugin.name` in the plugin registry.
- **`null` or `[]`** explicitly disables retry. `null` is preferred over sentinel values like `maxAttempts: 1` because it reads as intent, not a workaround.

The array form is intentional: once future retry kinds (tool retries, etc.) land as sibling abstract classes, composing "default model retry + custom tool retry" becomes `retryStrategy: [new DefaultModelRetryStrategy(), new MyToolRetryStrategy()]` rather than requiring one subclass to own both topics.

Retry-strategy registration ordering is not load-bearing for correctness: `DefaultModelRetryStrategy` guards on `event.retry`, so a hook that already set it short-circuits the strategy regardless of registration order.

New exports from the top-level index: `RetryStrategy` (type), `ModelRetryStrategy` (abstract class), `DefaultModelRetryStrategy`, `DefaultModelRetryStrategyOptions`, `BackoffStrategy`, `BackoffContext`, `JitterKind`, `ConstantBackoff`, `ConstantBackoffOptions`, `LinearBackoff`, `LinearBackoffOptions`, `ExponentialBackoff`, `ExponentialBackoffOptions`.

No changes to any provider module. The only change to `_invokeModel` is threading the new `attemptCount` value onto each `AfterModelCallEvent`; the retry dispatch still flows through the existing `event.retry` flag. `AfterModelCallEvent` is a reverse-callback event, so user-registered `AfterModelCallEvent` hooks run **before** this strategy — a user hook that sets `event.retry = true` short-circuits the default strategy and no extra delay is stacked.

### How this scales to future retry topics

When a future topic like tool retries lands, it's a new **sibling abstract class**, not a method added to `ModelRetryStrategy`. Different retry kinds have different units of work and different turn boundaries; collapsing them into one class would force a shared reset contract that doesn't match either kind cleanly.

```ts
// Future sibling abstract class — not part of this PR.
export abstract class ToolRetryStrategy implements Plugin {
  abstract retryTool(event: AfterToolCallEvent): void | Promise<void>

  initAgent(agent: LocalAgent): void {
    // ... attachment guard, hook registration for AfterToolCallEvent ...
  }
}

// The union grows an arm; existing subclasses and the config field are unchanged.
export type RetryStrategy = ModelRetryStrategy | ToolRetryStrategy
```

`DefaultModelRetryStrategy` is untouched when this lands.

The two common user stories:

**Extend `DefaultModelRetryStrategy`** — keep the default policy (throttled-error classification, attempt-based gating, per-turn state reset) and tweak one piece:

```ts
class MyModelRetries extends DefaultModelRetryStrategy {
  override readonly name = 'my:model-retries'
  // Override retryModel / _isRetryable / etc. as needed.
}
```

**Fully custom** — subclass `ModelRetryStrategy` directly and own `retryModel` end to end:

```ts
class FullCustomModelRetries extends ModelRetryStrategy {
  readonly name = 'my:full-custom'
  override async retryModel(event: AfterModelCallEvent) {
    // Own attempt-count interpretation, classification, backoff dispatch.
    // Use event.attemptCount === 1 as the turn boundary if you need per-turn state.
  }
}
```

Either subclass inherits the single-agent attachment guard and the `AfterModelCallEvent` hook registration from `ModelRetryStrategy` for free.

### Behavior Change

Agents now retry `ModelThrottledError` by default (6 attempts, exponential backoff with `baseMs=4s`, `maxMs=240s`, full jitter). Callers that relied on immediate propagation of `ModelThrottledError` can restore prior behavior with `retryStrategy: null`.

```typescript
// Before
const agent = new Agent({ model })
// ModelThrottledError propagates immediately on throttling

// After (equivalent behavior)
const agent = new Agent({ model, retryStrategy: null })
```

## Related Issues

Resolves: #472

## Documentation PR

will follow up

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

## Checklist

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -167,11 +167,10 @@ export type AgentConfig = {
    *
    * - Omitted: a sensible default {@link DefaultModelRetryStrategy} with exponential backoff is used.
    * - Single strategy: the given strategy is used.
-   * - Array of strategies: all are registered, in the given order. Hooks fire
-   *   in reverse registration order, so later entries in the array take
-   *   priority. Duplicates of the same concrete class are dropped (only the
-   *   first of each type is kept) with a logger warning.
-   * - `[]` or `null`: retries are explicitly disabled; failures propagate to the caller.
+   * - Array of strategies: all are registered, in the given order. Passing two
+   *   instances of the same concrete class logs a warning — they will collide
+   *   on `plugin.name` when the plugin registry initializes.
+   * - `null` or `[]`: retries are explicitly disabled; failures propagate to the caller.
    */
   retryStrategy?: RetryStrategy | RetryStrategy[] | null
   /**
@@ -327,7 +326,7 @@ export class Agent implements LocalAgent, InvokableAgent {
     this._hooksRegistry = new HookRegistryImplementation()
 
     // `undefined` (omitted) → install the default; `null`/`[]` → explicit opt-out.
-    const rawRetryStrategies: RetryStrategy[] =
+    const retryStrategies: RetryStrategy[] =
       config?.retryStrategy === null
         ? []
         : config?.retryStrategy === undefined
@@ -335,29 +334,16 @@ export class Agent implements LocalAgent, InvokableAgent {
           : Array.isArray(config.retryStrategy)
             ? config.retryStrategy
             : [config.retryStrategy]
-    warnOnDuplicateRetryStrategyTypes(rawRetryStrategies)
-    const seenStrategyTypes = new Set<new (...args: never[]) => RetryStrategy>()
-    const retryStrategies = rawRetryStrategies.filter((s) => {
-      const ctor = s.constructor as new (...args: never[]) => RetryStrategy
-      if (seenStrategyTypes.has(ctor)) return false
-      seenStrategyTypes.add(ctor)
-      return true
-    })
+    warnOnDuplicateRetryStrategyTypes(retryStrategies)
 
     // Initialize plugin registry with all plugins to be initialized during initialize().
-<<<<<<< HEAD
     // Ordering notes:
     // - ModelPlugin is registered last so that on AfterInvocationEvent (which uses
     //   reverse callback ordering), it runs first — clearing messages before
     //   SessionManager saves.
-    // - Retry-strategy ordering is not load-bearing for correctness: `ModelRetryStrategy`
+    // - Retry-strategy ordering is not load-bearing for correctness: `DefaultModelRetryStrategy`
     //   guards on `event.retry`, so a user hook that already set it short-circuits
     //   the strategy regardless of registration order.
-=======
-    // Ordering is not load-bearing for retry correctness: `DefaultModelRetryStrategy`
-    // guards on `event.retry` so a user hook that already set it short-circuits
-    // the strategy regardless of registration order.
->>>>>>> b4a1aba (feat: re-design retries as one abstract class per retry type. add attemptCount to AfterModelCallEvent hook)
     this._pluginRegistry = new PluginRegistry([
       this._conversationManager,
       ...retryStrategies,
@@ -1051,77 +1037,45 @@ export class Agent implements LocalAgent, InvokableAgent {
       streamOptions.toolChoice = toolChoice
     }
 
-<<<<<<< HEAD
-    // Estimate input tokens for the upcoming model call (non-fatal if estimation fails)
-    let projectedInputTokens: number | undefined
-    try {
-      projectedInputTokens = await this._estimateInputTokens(streamOptions)
-    } catch (e) {
-      logger.debug(`error=<${e}> | token estimation failed, proceeding without estimate`)
-    }
-
-    const beforeModelCallEvent = new BeforeModelCallEvent({
-      agent: this,
-      model: this.model,
-      invocationState,
-      ...(projectedInputTokens !== undefined && { projectedInputTokens }),
-    })
-    yield beforeModelCallEvent
-
-    if (beforeModelCallEvent.cancel) {
-      const cancelText =
-        typeof beforeModelCallEvent.cancel === 'string' ? beforeModelCallEvent.cancel : 'model call denied by hook'
-      const message = new Message({ role: 'assistant', content: [new TextBlock(cancelText)] })
-      const stopData: ModelStopData = { message, stopReason: 'endTurn' }
-      const afterModelCallEvent = new AfterModelCallEvent({
-        agent: this,
-        model: this.model,
-        stopData,
-        invocationState,
-      })
-      yield afterModelCallEvent
-
-      if (afterModelCallEvent.retry) {
-        return yield* this._invokeModel(invocationState, toolChoice)
-      }
-
-      return { message, stopReason: 'endTurn' }
-    }
-
-    // Start model span within loop span context
-    const modelId = this.model.modelId
-    const modelSpan = this._tracer.startModelInvokeSpan({
-      messages: this.messages,
-      ...(modelId && { modelId }),
-      ...(this.systemPrompt !== undefined && { systemPrompt: this.systemPrompt }),
-    })
-
-    try {
-      const result = yield* this._streamFromModel(this.messages, streamOptions, invocationState)
-
-      // Accumulate token usage and model latency metrics
-      this._meter.updateCycle(result.metadata)
-
-      // End model span with usage
-      const usage = result.metadata?.usage
-      const metrics = result.metadata?.metrics
-      this._tracer.endModelInvokeSpan(modelSpan, {
-        output: result.message,
-        stopReason: result.stopReason,
-        ...(usage && { usage }),
-        ...(metrics && { metrics }),
-      })
-
-      yield new ModelMessageEvent({
-        agent: this,
-        message: result.message,
-        stopReason: result.stopReason,
-        invocationState,
-      })
-=======
     let attemptCount = 1
     while (true) {
-      yield new BeforeModelCallEvent({ agent: this, model: this.model })
+      // Estimate input tokens for the upcoming model call (non-fatal if estimation fails)
+      let projectedInputTokens: number | undefined
+      try {
+        projectedInputTokens = await this._estimateInputTokens(streamOptions)
+      } catch (e) {
+        logger.debug(`error=<${e}> | token estimation failed, proceeding without estimate`)
+      }
+
+      const beforeModelCallEvent = new BeforeModelCallEvent({
+        agent: this,
+        model: this.model,
+        invocationState,
+        ...(projectedInputTokens !== undefined && { projectedInputTokens }),
+      })
+      yield beforeModelCallEvent
+
+      if (beforeModelCallEvent.cancel) {
+        const cancelText =
+          typeof beforeModelCallEvent.cancel === 'string' ? beforeModelCallEvent.cancel : 'model call denied by hook'
+        const message = new Message({ role: 'assistant', content: [new TextBlock(cancelText)] })
+        const stopData: ModelStopData = { message, stopReason: 'endTurn' }
+        const afterModelCallEvent = new AfterModelCallEvent({
+          agent: this,
+          model: this.model,
+          attemptCount,
+          stopData,
+          invocationState,
+        })
+        yield afterModelCallEvent
+
+        if (afterModelCallEvent.retry) {
+          attemptCount += 1
+          continue
+        }
+
+        return { message, stopReason: 'endTurn' }
+      }
 
       // Start model span within loop span context
       const modelId = this.model.modelId
@@ -1132,8 +1086,7 @@ export class Agent implements LocalAgent, InvokableAgent {
       })
 
       try {
-        const result = yield* this._streamFromModel(this.messages, streamOptions)
->>>>>>> b4a1aba (feat: re-design retries as one abstract class per retry type. add attemptCount to AfterModelCallEvent hook)
+        const result = yield* this._streamFromModel(this.messages, streamOptions, invocationState)
 
         // Accumulate token usage and model latency metrics
         this._meter.updateCycle(result.metadata)
@@ -1148,26 +1101,17 @@ export class Agent implements LocalAgent, InvokableAgent {
           ...(metrics && { metrics }),
         })
 
-<<<<<<< HEAD
-      const afterModelCallEvent = new AfterModelCallEvent({
-        agent: this,
-        model: this.model,
-        stopData,
-        invocationState,
-      })
-      yield afterModelCallEvent
-
-      if (afterModelCallEvent.retry) {
-        return yield* this._invokeModel(invocationState, toolChoice)
-      }
-=======
-        yield new ModelMessageEvent({ agent: this, message: result.message, stopReason: result.stopReason })
+        yield new ModelMessageEvent({
+          agent: this,
+          message: result.message,
+          stopReason: result.stopReason,
+          invocationState,
+        })
 
         // Handle user content redaction if guardrails blocked input
         if (result.redaction?.userMessage) {
           this._redactLastMessage(result.redaction.userMessage)
         }
->>>>>>> b4a1aba (feat: re-design retries as one abstract class per retry type. add attemptCount to AfterModelCallEvent hook)
 
         const stopData: ModelStopData = {
           message: result.message,
@@ -1175,23 +1119,19 @@ export class Agent implements LocalAgent, InvokableAgent {
           ...(result.redaction && { redaction: result.redaction }),
         }
 
-        const afterModelCallEvent = new AfterModelCallEvent({ agent: this, model: this.model, attemptCount, stopData })
+        const afterModelCallEvent = new AfterModelCallEvent({
+          agent: this,
+          model: this.model,
+          attemptCount,
+          stopData,
+          invocationState,
+        })
         yield afterModelCallEvent
 
-<<<<<<< HEAD
-      // Create error event
-      const errorEvent = new AfterModelCallEvent({
-        agent: this,
-        model: this.model,
-        error: modelError,
-        invocationState,
-      })
-=======
         if (afterModelCallEvent.retry) {
           attemptCount += 1
           continue
         }
->>>>>>> b4a1aba (feat: re-design retries as one abstract class per retry type. add attemptCount to AfterModelCallEvent hook)
 
         return result
       } catch (error) {
@@ -1206,6 +1146,7 @@ export class Agent implements LocalAgent, InvokableAgent {
           model: this.model,
           attemptCount,
           error: modelError,
+          invocationState,
         })
 
         // Yield error event - stream will invoke hooks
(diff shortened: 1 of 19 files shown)


strands-ts/src/retry/model-retry-strategy.ts
```diff
@@ -0,0 +1,168 @@
...
+ * const agent = new Agent({
+ *   model,
+ *   modelRetryStrategy: new ModelRetryStrategy({ maxAttempts: 4 }),
+ * })
+ * ```
+ */
+export class ModelRetryStrategy implements Plugin {
+  readonly name = 'strands:model-retry-strategy'
+
+  private readonly maxAttempts: number
```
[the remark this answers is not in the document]
notowen333: Done — renamed all private fields (and methods) in `ConstantBackoff`, `LinearBackoff`, `ExponentialBackoff`, and `ModelRetryStrategy` to use the underscore prefix.

strands-ts/src/retry/__tests__/model-retry-strategy.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    strategy.initAgent(agent)
+
+    const e1 = makeErrorEvent(agent, new ModelThrottledError('x'))
+    const p1 = invokeTrackedHook(agent, e1)
+    await vi.advanceTimersByTimeAsync(5)
+    await p1
+
+    expect(nextDelay).toHaveBeenCalledTimes(1)
+    expect(nextDelay.mock.calls[0]![0]!.attempt).toBe(1)
+    expect(nextDelay.mock.calls[0]![0]!.lastDelayMs).toBeUndefined()
```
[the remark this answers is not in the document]
notowen333: Done — both `nextDelay` calls now assert on the full `BackoffContext` object with `toEqual`.

strands-ts/src/agent/agent.ts
```diff
@@ -136,6 +137,14 @@ export type AgentConfig = {
    * Plugins to register with the agent.
    */
   plugins?: Plugin[]
+  /**
```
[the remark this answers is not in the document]
notowen333: Added the `needs-api-review` label.

strands-ts/src/retry/model-retry-strategy.ts
```diff
@@ -0,0 +1,168 @@
...
+        'ModelRetryStrategy: instance is already attached to another agent. ' +
+          'Create a separate ModelRetryStrategy per agent.'
+      )
+    }
+    this.attachedAgent = agent
+    agent.addHook(AfterModelCallEvent, (event) => this.onAfterModelCall(event))
+    agent.addHook(AfterInvocationEvent, () => this.resetState())
+  }
+
+  private async onAfterModelCall(event: AfterModelCallEvent): Promise<void> {
```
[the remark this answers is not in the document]
notowen333: Done — renamed to `_onAfterModelCall` (along with the other private methods: `_buildContext`, `_isRetryable`, `_resetState`).

strands-ts/src/retry/model-retry-strategy.ts
```diff
@@ -0,0 +1,168 @@
...
+    }
+    return ctx
+  }
+
+  /**
+   * Returns the backoff to use for a given error. Today this always returns
+   * the configured strategy; the indirection exists so a future per-error-type
+   * map can be layered in without changing the public API.
+   */
+  private resolveBackoff(_error: Error): BackoffStrategy {
```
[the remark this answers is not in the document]
notowen333: Agreed — inlined `this._backoff.nextDelay(...)` at the call site with a short comment noting per-error-class backoff selection is a future extension. Removed the unused `_error` parameter.

strands-ts/src/retry/model-retry-strategy.ts
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * @example
+ * ```ts
+ * const agent = new Agent({
+ *   model,
+ *   modelRetryStrategy: new ModelRetryStrategy({ maxAttempts: 4 }),
+ * })
+ * ```
+ */
+export class ModelRetryStrategy implements Plugin {
```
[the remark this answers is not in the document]
notowen333: Deferring this one. Introducing an abstract `ModelRetryStrategy` base class (or dedicated `HookProvider`-extending interface) with public `isRetryable`/`getBackoff` hooks is a larger API surface and directly conflicts with the YAGNI guidance applied to `resolveBackoff` — both would only earn their keep once we actually expand retry classification (5xx, transient server errors, per-error backoff maps). Today the class only handles `ModelThrottledError` via a single backoff, so formalizing the extension interface now would lock in shape before we have the data. Users who need custom classification can still implement `Plugin` directly against the same `AfterModelCallEvent.retry` flow; when the built-in set grows we can lift that into a proper abstract base class in a follow-up.

strands-ts/src/agent/agent.ts
```diff
@@ -262,9 +271,14 @@ export class Agent implements LocalAgent, InvokableAgent {
...
 
+    // `undefined` (omitted) → install the default; `null` → explicit opt-out.
+    const modelRetryStrategy =
+      config?.modelRetryStrategy === null ? undefined : (config?.modelRetryStrategy ?? new ModelRetryStrategy())
+
     // Initialize plugin registry with all plugins to be initialized during initialize()
     this._pluginRegistry = new PluginRegistry([
       this._conversationManager,
+      ...(modelRetryStrategy ? [modelRetryStrategy] : []),
       ...(config?.plugins ?? []),
```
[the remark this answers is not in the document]
notowen333: Added a comment at the plugin-registry call site. Also noting that ordering is not load-bearing for correctness: `_onAfterModelCall` guards on `event.retry` at the top, so a user hook that already set `retry=true` short-circuits the strategy regardless of where it sits in the plugin array.

strands-ts/src/agent/agent.ts
```diff
@@ -136,6 +137,14 @@ export type AgentConfig = {
...
    */
   plugins?: Plugin[]
+  /**
+   * Retry strategy for failed model calls (e.g. throttling).
+   *
+   * - Omitted: a sensible default {@link ModelRetryStrategy} with exponential backoff is used.
+   * - Provided: the given strategy is used.
+   * - `null`: retries are explicitly disabled; failures propagate to the caller.
+   */
+  modelRetryStrategy?: ModelRetryStrategy | null
```
poshinchen: In python we're using [retry_strategy](https://github.com/strands-agents/sdk-python/blob/main/src/strands/agent/agent.py#L147), do we want to align it?

notowen333: The call here is that we'd separate out another kind of retry strategy like `toolRetryStrategy` into its own constructor argument. The reasoning would be that a tool retry strategy would probably have a substantially different shape.

We can also swap it to `retryStrategy`

I think at that point we'd want it to be plural `retryStrategies` and accept a list since I think we are sure that it won't be a single class handling all kinds of retries

poshinchen: I personally like `retryStrategies` and model and tool's RetryStrategy will be in there.

strands-ts/src/retry/__tests__/backoff-strategy.test.ts
```diff
@@ -0,0 +1,138 @@
+// All tests here are purely synchronous — strategies compute a delay number;
+// no timers fire and nothing is awaited, so tests never wait real wall time.
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import { ConstantBackoff, LinearBackoff, ExponentialBackoff, type BackoffContext } from '../backoff-strategy.js'
+
+function ctx(partial: Partial<BackoffContext> = {}): BackoffContext {
```
zastrowm, line 7: Beautiful

strands-ts/src/retry/default-model-retry-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+ * @example
+ * ```ts
+ * const agent = new Agent({
+ *   model,
+ *   retryStrategy: new DefaultModelRetryStrategy({ maxAttempts: 4 }),
+ * })
+ * ```
+ */
+export class DefaultModelRetryStrategy extends ModelRetryStrategy {
+  readonly name = 'strands:model-retry-strategy'
```
zastrowm: Default intentionally excluded from the name here?

strands-ts/src/retry/model-retry-strategy.ts
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Other retry kinds (e.g. tool retries) will land as *sibling* abstract
+ * classes, not as additional methods on this one — different retry kinds
+ * have different unit-of-work boundaries and don't share a single reset
+ * contract.
+ *
+ * State scope: the base class does not define any reset hook. Per-turn
+ * state ownership lives in the subclass; {@link DefaultModelRetryStrategy}
+ * uses `event.attemptCount === 1` as its turn boundary.
+ *
```
zastrowm: This is a weird callout; we can talk about reset, but let's not talk about a "rest hook" or say that there is not one

strands-ts/src/retry/model-retry-strategy.ts
```diff
@@ -0,0 +1,66 @@
...
+  /**
+   * Handle a post-model-call event. Fires for **every** {@link AfterModelCallEvent},
+   * including successful completions (where `event.error` is undefined) — this
+   * lets stateful subclasses use the success path as a turn-boundary signal
+   * (e.g. to reset per-turn backoff state). Error events pass through the
+   * same method; implementations typically branch on `event.error`.
+   *
+   * To request a retry, set `event.retry = true` (typically after an
+   * `await sleep(delayMs)`). Returning without setting `event.retry` lets
+   * the error propagate normally.
```
zastrowm: This is odd; I would expect the plugin to return Promise<bool> and for it to decide to retry, rather than having to set it on the hook; otherwise I think this is too thin an abstraction on top of hooks

strands-ts/src/retry/default-model-retry-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+    this._backoff =
+      opts.backoff ?? new ExponentialBackoff({ baseMs: DEFAULT_BACKOFF_BASE_MS, maxMs: DEFAULT_BACKOFF_MAX_MS })
+  }
+
+  override async retryModel(event: AfterModelCallEvent): Promise<void> {
+    // Another hook already requested retry — don't stack a second delay on top.
+    if (event.retry) return
+
+    // Clear any state left over from a prior turn at the turn boundary.
+    if (event.attemptCount === 1) this._reset()
```
zastrowm: Given that I expect every implementation is going to have to do this same logic, I feel like this should exist in the base class; same for event.retry

strands-ts/src/retry/default-model-retry-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    logger.debug(
+      `retry_delay_ms=<${delayMs}> attempt_count=<${event.attemptCount}> max_attempts=<${this._maxAttempts}> ` +
+        `| retryable model error, delaying before retry`
+    )
+
+    await sleep(delayMs)
+
+    this._lastDelayMs = delayMs
+    event.retry = true
```
zastrowm: Per my other comment, if eel like this should be `return true`

strands-ts/src/retry/default-model-retry-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+      attempt: attemptCount,
+      elapsedMs: this._firstFailureAt === undefined ? 0 : Date.now() - this._firstFailureAt,
+    }
+    if (this._lastDelayMs !== undefined) {
+      ctx.lastDelayMs = this._lastDelayMs
+    }
+    return ctx
+  }
+
+  private _isRetryable(error: Error): boolean {
```
zastrowm: *if* we think it makes sense, exposing `shouldRetry(error: Error)` as a public/overridable method would also give an easy way for folks to try different exception types, FWIW

strands-ts/src/retry/default-model-retry-strategy.ts
```diff
@@ -0,0 +1,144 @@
...
+    const maxAttempts = opts.maxAttempts ?? DEFAULT_MAX_ATTEMPTS
+    if (!Number.isInteger(maxAttempts) || maxAttempts < 1) {
+      throw new Error(`DefaultModelRetryStrategy: maxAttempts must be an integer >= 1 (got ${maxAttempts})`)
+    }
+    this._maxAttempts = maxAttempts
+    this._backoff =
+      opts.backoff ?? new ExponentialBackoff({ baseMs: DEFAULT_BACKOFF_BASE_MS, maxMs: DEFAULT_BACKOFF_MAX_MS })
+  }
+
+  override async retryModel(event: AfterModelCallEvent): Promise<void> {
```
zastrowm: Do you imagine that we'd want to add any additional properties on top of `AfterModelCallEvent` that would warrant a separate type?

strands-ts/src/retry/default-model-retry-strategy.ts
notowen333: settled on `onFirstModelAttempt` as the replacement for `reset` as to not be opinionated about what concrete classes are doing

strands-ts/src/retry/model-retry-strategy.ts
notowen333: Makes sense; I also lifted the sleep into the abstract class. the retry now returns a retryDecision which gives a retry bool as well as a sleep time

strands-ts/src/retry/default-model-retry-strategy.ts
notowen333: I think if there is anything else that we might need there's no reason not to use the existing event. Maybe others need the same

strands-ts/src/retry/default-model-retry-strategy.ts
```diff
@@ -0,0 +1,141 @@
...
+    this._maxAttempts = maxAttempts
+    this._backoff =
+      opts.backoff ?? new ExponentialBackoff({ baseMs: DEFAULT_BACKOFF_BASE_MS, maxMs: DEFAULT_BACKOFF_MAX_MS })
+  }
+
+  /**
+   * Whether `error` should be retried. Override to extend or narrow the
+   * retryable set (e.g. to also retry transient 5xx errors).
+   */
+  public isRetryable(error: Error): boolean {
```
zastrowm: I think this should be protected - external developers should not be calling this - it's more for customized strategies

strands-ts/src/retry/model-retry-strategy.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (event.error === undefined) return
+
+    const decision = await this.computeRetryDecision(event)
+    if (!decision.retry) return
+
+    await sleep(decision.waitMs)
+    event.retry = true
+  }
+
+  initAgent(agent: LocalAgent): void {
```
zastrowm, line 99: I think in other palces we explicitly talk about calling super.initAgent if you override it; might be good here

zastrowm (dismissed):

poshinchen (approved):

PR #890, opened 2026-04-24
feat: estimate input tokens before model calls

opieter-aws:
## Description                                                                                                                                                                                            
                                                                                                                                                                                                      
Adds input token estimation to the agent loop, making it available on `BeforeModelCallEvent` before every model call. This is the foundation for proactive context compression (design 0008).                                                                                                                                                                                            
                                                                                                                                                                                                      
The estimation uses the token counting strategy from the design doc: it reads `inputTokens + outputTokens` from the last assistant message's metadata as a known baseline, then estimates only new messages added since (typically tool results) via `model.countTokens()`. On cold start (no metadata available), it falls back to estimating all messages. Estimation is non-fatal — if it fails, the agent proceeds without it.                                                                                                                                                                                      
                                                                                                                                                                                                      
`BeforeModelCallEvent` now carries an optional `projectedInputTokens` field:                                                                                                                              
                                                                                                                                                                                                      
```typescript                                                                                                                                                                                             
agent.addHook(BeforeModelCallEvent, (event) => {
     console.log(event.projectedInputTokens) // e.g. 14200
})
```                                                                                                                                                                                              

No breaking changes. All new fields are optional.                                                                                                                                                         

## Related Issues                                                                                                                                                                                            
                                                                                                                                                                                                      
https://github.com/strands-agents/sdk-python/issues/555                                                                                                                                                                                               
                                                                                                                                                                                                      
## Documentation PR                                                                                                                                                                                          
                                                                                                                                                                                                      
Will do one big docs update when proactive compression is out                                                                                             
                                                                                                                                                                                                      
### Type of Change                                                                                                                                                                                            
                                                                                                                                                                                                      
New feature                                                                                                                                                                                               
                                                                                                                                                                                                      
## Testing                                                                                                                                                                                                   
                                                                                                                                                                                                      
- [x] I ran npm run check                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                      
## Checklist                                                                                                                                                                                                 
                                                                                                                                                                                                      
--------

### DIFF
--- a/strands-ts/src/agent/__tests__/agent.test.ts
+++ b/strands-ts/src/agent/__tests__/agent.test.ts
@@ -24,6 +24,7 @@ import {
   AfterToolCallEvent,
   AfterToolsEvent,
   BeforeInvocationEvent,
+  BeforeModelCallEvent,
   BeforeToolsEvent,
 } from '../../hooks/events.js'
 import { BedrockModel } from '../../models/bedrock.js'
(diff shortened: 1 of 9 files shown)


strands-ts/src/agent/agent.ts
```diff
@@ -897,7 +897,19 @@ export class Agent implements LocalAgent, InvokableAgent {
       streamOptions.toolChoice = toolChoice
     }
 
-    yield new BeforeModelCallEvent({ agent: this, model: this.model })
+    // Estimate input tokens for the upcoming model call (non-fatal if estimation fails)
+    let estimatedInputTokens: number | undefined
+    try {
+      estimatedInputTokens = await this._estimateInputTokens(streamOptions)
+    } catch {
```
JackYPCOnline: question: what exceptions or errors do you expect here? we shouldn't swallow anything unexpected?

strands-ts/src/hooks/events.ts
```diff
@@ -284,18 +284,32 @@ export class BeforeModelCallEvent extends HookableEvent {
...
   readonly model: Model
 
-  constructor(data: { agent: LocalAgent; model: Model }) {
+  /**
+   * Estimated input token count for the upcoming model call.
+   * Computed by the agent loop from message metadata and token estimation.
+   * Available for hooks and plugins (e.g. conversation managers) to make
+   * proactive decisions about context management.
+   */
+  readonly estimatedInputTokens?: number
```
lizradway: i think it would be nice for the naming to be something that explicitly suggests it is for the upcoming agent loop iteration. 

- `projectedInputTokens`
- `pendingInputTokens`
- `inputTokenEstimate`
- `preCallInputTokens`
- `nextInputTokensEstimate`


some thoughts? i kinda like `projectedInputTokens` since that is explicitly forward looking, whereas `estimated` can be retrospective.

strands-ts/src/agent/__tests__/agent.test.ts
```diff
@@ -1594,3 +1595,125 @@ describe('Agent._redactLastMessage', () => {
...
+    model.addTurn({ type: 'textBlock', text: 'Hello' })
+    const countTokensSpy = vi.spyOn(model, 'countTokens')
+    countTokensSpy.mockResolvedValue(42)
+
+    const agent = new Agent({ model, printer: false })
+    const tokenPromise = captureEstimatedTokens(agent)
+    await agent.invoke('Hi')
+
+    expect(await tokenPromise).toBe(42)
+    expect(countTokensSpy).toHaveBeenCalledWith(expect.any(Array), expect.any(Object))
```
lizradway, line 1619: nit: id recommend passing [object assertion best practices](https://github.com/strands-agents/sdk-typescript/blob/main/docs/TESTING.md#object-assertion-best-practices) through to your global agent.md for the future

strands-ts/src/hooks/events.ts
lizradway: is this pr targeting https://github.com/strands-agents/sdk-python/issues/555 or https://github.com/strands-agents/sdk-python/issues/298 or both?


it seems to me like putting this on the `beforeModelCallEvent` is targeting https://github.com/strands-agents/sdk-python/issues/298 (where the SDK needs token counts mid-invocation), but from my remembrance/understanding we wanted to expose projected input tokens to end users on `AgentResult` so they can check between invocations without writing a hook.

lizradway: for https://github.com/strands-agents/sdk-python/issues/555, you would need to surface `latestContextsize + outputTokens` from the last model response’s usage metadata onto `AgentResult` to that gives users the baseline estimate for the next call without needing a hook

lizradway: followed up offline :
- we'll be exposing this on `AgentResult` in follow up
- [proactive compression](https://github.com/strands-agents/sdk-python) will be built on `beforeModelCallEvent` with [in event loop compression](https://github.com/strands-agents/sdk-python/issues/298) built on `beforeToolCallEvent` as eventuality.

strands-ts/src/agent/agent.ts
opieter-aws: The expected errors here are `countTokens` failures. The base Model class has a default heuristic implementation, but provider overrides could throw if their token counting API is unavailable. The catch-all is because estimation is best-effort and shouldn't block the agent loop, but I'll add a `logger.debug` to make exceptions visible here. Good callout.

strands-ts/src/hooks/events.ts
opieter-aws: Thanks for the suggestions! I agree `projectedInputTokens` is the most accurate there. @mkmeral any thoughts?

opieter-aws: For https://github.com/strands-agents/sdk-python/issues/555, I agree we should also surface `inputTokens + outputTokens` from the last model response onto `AgentResult` so users can check between invocations without writing a hook. I'll add that as a new field, what do you think about `upcomingContextSize` or `projectedContextSize`? 

Regarding https://github.com/strands-agents/sdk-python/issues/298, as we discussed, the proactive context management design provides in-event-loop-cycle management because it always evaluates at the `beforeModelCall` event. That may not be the full scope of https://github.com/strands-agents/sdk-python/issues/298, but it does provide a mechanism to manage context in-loop

strands-ts/src/hooks/events.ts
```diff
@@ -284,18 +284,32 @@ export class BeforeModelCallEvent extends HookableEvent {
...
+
+  constructor(data: { agent: LocalAgent; model: Model; estimatedInputTokens?: number }) {
     super()
     this.agent = data.agent
     this.model = data.model
+    if (data.estimatedInputTokens !== undefined) {
```
lizradway: why are we exposing this field on the BeforeModelCallEvent hook? I think we should be doing the estimation in the Conversation Manager class internally by registering a new hook on the plugin without any changes needed in the agent loop.

i believe this approach ^ is what was documented in the [design doc](https://github.com/opieter-aws/docs/blob/77e851d8419ff997d44b5021c229219c265457a0/designs/0008-proactive-context-compression.md?plain=1#L55)

opieter-aws: Yes it was indeed, but this ties the estimated input token to using a conversation manager. With the proposed approach in the PR, any user can leverage the `estimatedInputTokens` field for their own application, without requiring a conversation manager or having to alter the SDK. This for example enables agent-driven context management by exposing the field to an agent.

lizradway: oh i see ok that makes a lot of sense. i dont love adding more things into the agent loop, but i don't think this could live anywhere else unfortunately.

maybe something we can consider condensing into some sort of context module once we've aligned on the high level plan? would be curious to hear your thoughts/pick your brain on that offline.


this looks good to me though and makes sense in that case!

lizradway (dismissed):

lizradway (dismissed):

lizradway (approved):

JackYPCOnline (approved):

mkmeral (approved):

PR #907, opened 2026-04-24
feat: make tool result mutable

lizradway:
## Description

Makes `AfterToolCallEvent.result` mutable so hook callbacks can replace tool results before they enter the conversation history.

Previously, `result` was `readonly` and the agent loop returned its local `toolResult` variable, ignoring any modifications hooks made to the event. This change removes `readonly` from the property and updates `executeTool()` to return `afterToolCallEvent.result`, so plugins can intercept and transform tool results.

This is a non-breaking change — existing code that reads `event.result` is unaffected. The only difference is that assigning to `event.result` now compiles and takes effect. This follows the same pattern as the existing mutable `retry` field on the same event.

Motivated by the `ContextOffloader` plugin port (sdk-python#2162), which needs to replace oversized tool results with truncated previews. This also enables the `Transform` intervention action in Interventions (https://github.com/strands-agents/sdk-typescript/pull/883).

## Related Issues

sdk-python#2162

## Documentation PR

N/A

## Type of Change

New feature

## Testing

- [ ] I ran `npm run check`
- New unit test verifies that replacing `event.result` in an `AfterToolCallEvent` hook causes the modified result to appear in the agent's conversation history
- All existing tests pass (26 hook tests, 37 events tests)

## Checklist

----

### DIFF
--- a/strands-ts/src/hooks/__tests__/events.test.ts
+++ b/strands-ts/src/hooks/__tests__/events.test.ts
@@ -220,8 +220,6 @@ describe('AfterToolCallEvent', () => {
     event.toolUse = toolUse
     // @ts-expect-error verifying that property is readonly
     event.tool = tool
-    // @ts-expect-error verifying that property is readonly
-    event.result = result
   })
 
   it('creates instance with error property when tool execution fails', () => {
--- a/strands-ts/src/hooks/__tests__/events.test.ts
+++ b/strands-ts/src/hooks/__tests__/events.test.ts
@@ -222,6 +222,21 @@ describe('AfterToolCallEvent', () => {
     event.tool = tool
   })
 
+  it('allows result to be replaced', () => {
+    const agent = new Agent()
+    const toolUse = { name: 'test', toolUseId: 'id', input: {} }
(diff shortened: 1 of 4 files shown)


opieter-aws (dismissed):

strands-ts/src/hooks/__tests__/events.test.ts
```diff
@@ -220,8 +220,6 @@ describe('AfterToolCallEvent', () => {
     event.toolUse = toolUse
     // @ts-expect-error verifying that property is readonly
     event.tool = tool
-    // @ts-expect-error verifying that property is readonly
-    event.result = result
```
zastrowm, line 224: I'm a little surprised to find no explicit test enforcing writability in here

zastrowm (dismissed):

strands-ts/src/hooks/__tests__/events.test.ts
lizradway: can add a quick unit test for tool result mutability

opieter-aws (approved):

PR #909, opened 2026-04-24
feat: centralize model defaults; emit warnings when defaults are used…

notowen333:
## Description

This PR seeks to create consistency around defaults. There are a few decisions made here:

1. Only create `defaults.ts` within the `models/` directory. This is where the most important/comparable defaults live. There aren't great candidates for other defaults which should be removed from the usage site. We have the option to add a higher level defaults file if it ever becomes applicable.
2. Add a small module that maintains a global set to keep track of which warning messages have been used within the process. Then we can avoid emitting the same warning multiple times.

The second item was informed by adding the same default model id warning we just added to python: https://github.com/strands-agents/sdk-python/blob/b340dc4d390f7effbcf75db9311fdd20cf6ae67b/src/strands/models/bedrock.py#L1184

## Related Issues

#823

## Documentation PR

None required.

## Type of Change

Refactor (defaults)
New feature (warnings)

## Testing

- [x] I ran `npm run check`

## Checklist

----


notowen333:
I'll sort out the failed tests after a comment pass

strands-ts/src/models/anthropic.ts
```diff
@@ -40,15 +40,23 @@ export class AnthropicModel extends Model<AnthropicModelConfig> {
...
-      modelId: DEFAULT_ANTHROPIC_MODEL_ID,
-      maxTokens: DEFAULT_ANTHROPIC_MAX_TOKENS,
+      modelId: MODEL_DEFAULTS.anthropic.modelId,
+      maxTokens: MODEL_DEFAULTS.anthropic.maxTokens,
       ...modelConfig,
     }
 
+    if (modelConfig.modelId === undefined) {
+      const msg = defaultModelWarningMessage(MODEL_DEFAULTS.anthropic.modelId)
+      if (claimFirstWarning(msg)) {
```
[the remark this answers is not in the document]
opieter-aws: +1 on this suggestion

opieter-aws (commented):
The setup looks good to me, I agree with the bot's suggestion to change `claimFirstWarning` into a more reusable component or add a `warnOnce` helper function

notowen333:
> The setup looks good to me, I agree with the bot's suggestion to change `claimFirstWarning` into a more reusable component or add a `warnOnce` helper function

I started with a warnOnce helper but I felt that seeing the log.warn inline was easier to parse at the call site.

I'm ok to go back in that direction too

strands-ts/src/models/google/model.ts
```diff
@@ -94,6 +92,13 @@ export class GoogleModel extends Model<GoogleModelConfig> {
 
     this._config = modelConfig
 
+    if (modelConfig.modelId === undefined) {
+      const msg = defaultModelWarningMessage(MODEL_DEFAULTS.gemini.modelId)
+      if (claimFirstWarning(msg)) {
+        logger.warn(msg)
+      }
```
zastrowm: This is the same warning throughout, right?

Let's simplify this to:

```
warnOnce(logger, defaultModelWarningMessage(MODEL_DEFAULTS.gemini.modelId))
```

since the pattern is always going to be the same

zastrowm: Oh sure, the agent says it too

strands-ts/src/logging/warn-once.ts
```diff
@@ -0,0 +1,19 @@
...
+ * repeated nudges (e.g. "using default modelId") from flooding logs
+ * when many instances are constructed.
+ *
+ * @param logger - Logger to emit the warning on
+ * @param msg - Warning message; also used as the dedupe key
+ */
+export function warnOnce(logger: Logger, msg: string): void {
+  if (warned.has(msg)) return
+  warned.add(msg)
+  logger.warn(msg)
```
opieter-aws: nit: swapping the order here ensures warning is emitted before it's saved

opieter-aws (dismissed):

opieter-aws (approved):

PR #954, opened 2026-04-28
feat: auto-populate contextWindowLimit from model ID lookup tables

### LINKED ISSUE #851: [FEATURE] Auto-populate contextWindowLimit from model ID lookup tables
### Problem Statement

PR #848 adds contextWindowLimit as an optional property on BaseModelConfig, but users must manually supply the value for every model they instantiate. There is no built-in mapping from model IDs to their known context window sizes, so callers need to look up and hard-code limits themselves.

### Proposed Solution

Add a single, provider-agnostic lookup table that maps model IDs to their context window token limits. When contextWindowLimit is not explicitly provided in the config, the model provider resolves it from this shared table based on the configured model ID.                                                                                                                                                
             
```                                                                                                                                                                                             
  // Centralized, shared across all providers                                                                                                                                                             
  const MODEL_CONTEXT_WINDOW_LIMITS: Record<string, number> = {                                                                                                                                           
    'anthropic.claude-sonnet-4-20250514-v1:0': 200_000,                                                                                                                                                   
    'claude-sonnet-4-20250514': 200_000,                                                                                                                                                                  
    'amazon.nova-pro-v1:0': 300_000,                                                                                                                                                                      
    'gpt-4o': 128_000,                                                                                                                                                                                    
    // ...                                                                                                                                                                                                
  };                                                                                                                                                                                                      
```                                                                                                                                                                                                          
 
An explicit `contextWindowLimit` in the config should always take precedence over the lookup value.  

### Use Case

- Conversation managers can implement proactive context compression (e.g. "compress at 70% of context window") without requiring users to know and supply the limit for every model.                    
- A single table avoids duplicating entries across providers and gives one place to update when new models release.                                                                                     
- Aligns with the Python SDK's approach in sdk-python#1295.                                                                                                                                             
- Reduces boilerplate for users who just want sensible defaults.

### Alternatives Solutions

_No response_

### Additional Context

- Follow-up to #848 which introduced the contextWindowLimit property on BaseModelConfig.                                                                                                                
- The Python SDK tracks the equivalent work in sdk-python#1295 (https://github.com/strands-agents/sdk-python/issues/1295).                                                                              
- Related to #823 (default model settings centralization) and #825 (restore model-aware default maxTokens).

opieter-aws:
## Description

`contextWindowLimit` is now auto-populated from a built-in lookup table when not explicitly provided in the model config. This removes the need for users to manually look up and hard-code context window sizes for every model they instantiate.

A new `MODEL_METADATA` table in `defaults.ts` maps model IDs to their known metadata (currently `contextWindowLimit`, structured as `Record<string, ModelMetadataEntry>` to support adding fields like `maxTokens` in a follow-up). Each provider constructor resolves the limit after setting the model ID. Explicit `contextWindowLimit` always takes precedence. Unknown model IDs leave it `undefined`.

For Bedrock cross-region model IDs (e.g. `us.anthropic.claude-sonnet-4-6`), the region prefix is stripped before lookup.

```typescript
// Before: users had to supply contextWindowLimit manually
const model = new BedrockModel({
  modelId: 'anthropic.claude-sonnet-4-20250514-v1:0',
  contextWindowLimit: 1_000_000,
})

// After: resolved automatically from the model ID
const model = new BedrockModel({
  modelId: 'anthropic.claude-sonnet-4-20250514-v1:0',
})

model.getConfig().contextWindowLimit // 1_000_000
```
Values sourced from https://github.com/BerriAI/litellm/blob/litellm_internal_staging/model_prices_and_context_window.json

## Related Issues

Resolves: #851
Follow-up to #848 (added contextWindowLimit to BaseModelConfig)

## Documentation PR

NA

## Type of Change

New feature

## Testing

- [x] I ran npm run check

## Checklist

### DIFF
--- a/strands-ts/src/models/anthropic.ts
+++ b/strands-ts/src/models/anthropic.ts
@@ -8,7 +8,12 @@ import type { ImageBlock, DocumentBlock } from '../types/media.js'
 import { encodeBase64 } from '../types/media.js'
 import { logger } from '../logging/logger.js'
 import { warnOnce } from '../logging/warn-once.js'
-import { MODEL_DEFAULTS, defaultMaxTokensWarningMessage, defaultModelWarningMessage } from './defaults.js'
+import {
+  MODEL_DEFAULTS,
+  defaultMaxTokensWarningMessage,
+  defaultModelWarningMessage,
+  getContextWindowLimit,
+} from './defaults.js'
 
 const CONTEXT_WINDOW_OVERFLOW_ERRORS = ['prompt is too long', 'max_tokens exceeded', 'input too long']
 const TEXT_FILE_FORMATS = ['txt', 'md', 'markdown', 'csv', 'json', 'xml', 'html', 'yml', 'yaml', 'js', 'ts', 'py']
@@ -51,6 +56,11 @@ export class AnthropicModel extends Model<AnthropicModelConfig> {
 
     if (modelConfig.maxTokens === undefined) {
       warnOnce(logger, defaultMaxTokensWarningMessage(MODEL_DEFAULTS.anthropic.maxTokens))
+    }
+
+    if (this._config.contextWindowLimit === undefined) {
+      const contextWindowLimit = getContextWindowLimit(this._config.modelId ?? MODEL_DEFAULTS.anthropic.modelId)
+      if (contextWindowLimit !== undefined) this._config.contextWindowLimit = contextWindowLimit
     }
 
     if (client) {
--- a/strands-ts/src/models/anthropic.ts
+++ b/strands-ts/src/models/anthropic.ts
@@ -1,5 +1,11 @@
 import Anthropic, { type ClientOptions } from '@anthropic-ai/sdk'
-import { Model, type BaseModelConfig, type CountTokensOptions, type StreamOptions } from '../models/model.js'
+import {
+  Model,
+  type BaseModelConfig,
+  type CountTokensOptions,
+  type StreamOptions,
+  resolveConfigMetadata,
+} from '../models/model.js'
 import type { Message, ContentBlock } from '../types/messages.js'
 import type { ModelStreamEvent } from '../models/streaming.js'
 import { createEmptyUsage } from '../models/streaming.js'
@@ -8,12 +14,7 @@ import type { ImageBlock, DocumentBlock } from '../types/media.js'
 import { encodeBase64 } from '../types/media.js'
 import { logger } from '../logging/logger.js'
 import { warnOnce } from '../logging/warn-once.js'
-import {
-  MODEL_DEFAULTS,
-  defaultMaxTokensWarningMessage,
-  defaultModelWarningMessage,
-  getContextWindowLimit,
-} from './defaults.js'
+import { MODEL_DEFAULTS, defaultMaxTokensWarningMessage, defaultModelWarningMessage } from './defaults.js'
 
 const CONTEXT_WINDOW_OVERFLOW_ERRORS = ['prompt is too long', 'max_tokens exceeded', 'input too long']
 const TEXT_FILE_FORMATS = ['txt', 'md', 'markdown', 'csv', 'json', 'xml', 'html', 'yml', 'yaml', 'js', 'ts', 'py']
@@ -58,11 +59,6 @@ export class AnthropicModel extends Model<AnthropicModelConfig> {
       warnOnce(logger, defaultMaxTokensWarningMessage(MODEL_DEFAULTS.anthropic.maxTokens))
     }
 
-    if (this._config.contextWindowLimit === undefined) {
-      const contextWindowLimit = getContextWindowLimit(this._config.modelId ?? MODEL_DEFAULTS.anthropic.modelId)
-      if (contextWindowLimit !== undefined) this._config.contextWindowLimit = contextWindowLimit
-    }
-
     if (client) {
       this._client = client
     } else {
@@ -91,7 +87,7 @@ export class AnthropicModel extends Model<AnthropicModelConfig> {
   }
 
   getConfig(): AnthropicModelConfig {
-    return this._config
+    return resolveConfigMetadata(this._config, this._config.modelId ?? MODEL_DEFAULTS.anthropic.modelId)
   }
 
   /**
--- a/strands-ts/src/models/defaults.ts
+++ b/strands-ts/src/models/defaults.ts
@@ -41,3 +41,152 @@ export function defaultModelWarningMessage(defaultModelId: string): string {
 export function defaultMaxTokensWarningMessage(defaultMaxTokens: number): string {
   return `max_tokens=<${defaultMaxTokens}> | using default maxTokens, which is subject to change | set maxTokens explicitly to pin the value`
 }
+
+/**
+ * Metadata for known model IDs.
+ *
+ * Values sourced from provider documentation and
+ * https://github.com/BerriAI/litellm/blob/litellm_internal_staging/model_prices_and_context_window.json
+ *
+ * For Bedrock models with cross-region prefixes (e.g. `us.`, `eu.`, `global.`),
+ * {@link getModelMetadata} strips the prefix before lookup so only the base model ID is needed here.
+ */
+export interface ModelMetadataEntry {
+  /**
+   * Maximum context window size in tokens (input + output combined).
+   */
+  contextWindowLimit: number
+}
+
+export const MODEL_METADATA: Record<string, ModelMetadataEntry> = {
+  // Anthropic (direct API)
+  'claude-sonnet-4-6': { contextWindowLimit: 1_000_000 },
+  'claude-sonnet-4-20250514': { contextWindowLimit: 1_000_000 },
+  'claude-sonnet-4-5': { contextWindowLimit: 200_000 },
+  'claude-sonnet-4-5-20250929': { contextWindowLimit: 200_000 },
+  'claude-opus-4-6': { contextWindowLimit: 1_000_000 },
+  'claude-opus-4-6-20260205': { contextWindowLimit: 1_000_000 },
+  'claude-opus-4-7': { contextWindowLimit: 1_000_000 },
+  'claude-opus-4-7-20260416': { contextWindowLimit: 1_000_000 },
+  'claude-opus-4-5': { contextWindowLimit: 200_000 },
+  'claude-opus-4-5-20251101': { contextWindowLimit: 200_000 },
+  'claude-opus-4-20250514': { contextWindowLimit: 200_000 },
+  'claude-opus-4-1': { contextWindowLimit: 200_000 },
+  'claude-opus-4-1-20250805': { contextWindowLimit: 200_000 },
+  'claude-haiku-4-5': { contextWindowLimit: 200_000 },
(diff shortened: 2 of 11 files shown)


strands-ts/src/models/anthropic.ts
```diff
@@ -77,7 +91,16 @@ export class AnthropicModel extends Model<AnthropicModelConfig> {
   }
 
   updateConfig(modelConfig: AnthropicModelConfig): void {
+    const modelIdChanged = modelConfig.modelId && modelConfig.modelId !== this._config.modelId
     this._config = { ...this._config, ...modelConfig }
+
+    // Re-resolve contextWindowLimit if it was auto-resolved initially and modelId changed
+    if (this._autoResolvedContextWindow && modelIdChanged && !modelConfig.contextWindowLimit) {
+      const contextWindowLimit = getContextWindowLimit(this._config.modelId!)
+      if (contextWindowLimit !== undefined) {
```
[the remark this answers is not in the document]
opieter-aws: /strands can you implement a solution to this edge case?

strands-ts/src/models/defaults.ts
```diff
@@ -41,3 +41,152 @@ export function defaultModelWarningMessage(defaultModelId: string): string {
...
+ * {@link getModelMetadata} strips the prefix before lookup so only the base model ID is needed here.
+ */
+export interface ModelMetadataEntry {
+  /**
+   * Maximum context window size in tokens (input + output combined).
+   */
+  contextWindowLimit: number
+}
+
+export const MODEL_METADATA: Record<string, ModelMetadataEntry> = {
```
lizradway: why is this exported?

lizradway: what is our maintenance strategy for this. i really don't love the idea of having to update this for every new model drop. when can we remove a deprecated model from this list or will this grow forever?

strands-ts/src/models/defaults.ts
```diff
@@ -41,3 +41,152 @@ export function defaultModelWarningMessage(defaultModelId: string): string {
...
+/**
+ * Metadata for known model IDs.
+ *
+ * Values sourced from provider documentation and
+ * https://github.com/BerriAI/litellm/blob/litellm_internal_staging/model_prices_and_context_window.json
+ *
+ * For Bedrock models with cross-region prefixes (e.g. `us.`, `eu.`, `global.`),
+ * {@link getModelMetadata} strips the prefix before lookup so only the base model ID is needed here.
+ */
+export interface ModelMetadataEntry {
```
lizradway: maybe just `ModelMetadata`?

strands-ts/src/models/defaults.ts
```diff
@@ -41,3 +41,152 @@ export function defaultModelWarningMessage(defaultModelId: string): string {
...
+/**
+ * Looks up metadata for a model ID.
+ *
+ * For Bedrock cross-region model IDs (e.g. `us.anthropic.claude-sonnet-4-6`),
+ * the region prefix is stripped before lookup.
+ *
+ * @param modelId - The model ID to look up
+ * @returns The metadata entry, or undefined if not found
+ */
+export function getModelMetadata(modelId: string): ModelMetadataEntry | undefined {
```
lizradway: similarly does this need to be exported?

strands-ts/src/models/defaults.ts
```diff
@@ -41,3 +41,152 @@ export function defaultModelWarningMessage(defaultModelId: string): string {
...
+export interface ModelMetadataEntry {
+  /**
+   * Maximum context window size in tokens (input + output combined).
+   */
+  contextWindowLimit: number
+}
+
+export const MODEL_METADATA: Record<string, ModelMetadataEntry> = {
+  // Anthropic (direct API)
+  'claude-sonnet-4-6': { contextWindowLimit: 1_000_000 },
```
lizradway: I think naming this metadata is also pretty confusing given that this only contains the context window limit. i really really hope we don't ever need to add more items to this mapping.

could this strictly be `MODEL_CONTEXT_WINDOW_LIMITS`?

strands-ts/src/models/defaults.ts
```diff
@@ -41,3 +41,152 @@ export function defaultModelWarningMessage(defaultModelId: string): string {
...
+  'gpt-5-mini': { contextWindowLimit: 272_000 },
+  'gpt-5-nano': { contextWindowLimit: 272_000 },
+  'gpt-5-pro': { contextWindowLimit: 128_000 },
+  'gpt-4.1': { contextWindowLimit: 1_047_576 },
+  'gpt-4.1-mini': { contextWindowLimit: 1_047_576 },
+  'gpt-4.1-nano': { contextWindowLimit: 1_047_576 },
+  'gpt-4o': { contextWindowLimit: 128_000 },
+  'gpt-4o-mini': { contextWindowLimit: 128_000 },
+  'gpt-4-turbo': { contextWindowLimit: 128_000 },
+  o3: { contextWindowLimit: 200_000 },
```
lizradway: why are some of these strings literals and some not?

strands-ts/src/models/defaults.ts
lizradway: I do like the idea of model metadata, but im wondering if we want to store that on the model interface instead.

strands-ts/src/models/defaults.ts
```diff
@@ -41,3 +41,152 @@ export function defaultModelWarningMessage(defaultModelId: string): string {
...
+  return undefined
+}
+
+/**
+ * Looks up the context window limit for a model ID.
+ *
+ * @param modelId - The model ID to look up
+ * @returns The context window limit in tokens, or undefined if not found
+ */
+export function getContextWindowLimit(modelId: string): number | undefined {
```
lizradway: is this helper necessary? maybe we just keep export `getModelMetadata` exported and then access the model metadata within the model override.

alternatively, i suggested above adding metadata to the model interface (i like this approach).

strands-ts/src/models/model.ts
```diff
@@ -26,6 +26,23 @@ import {
...
+/**
+ * Returns a config object with `contextWindowLimit` resolved from the lookup table
+ * when not explicitly set. Explicit values pass through unchanged.
+ *
+ * @internal
+ * @param config - The stored model config (may lack contextWindowLimit)
+ * @param modelId - The model ID to look up
+ * @returns A new config with contextWindowLimit resolved if available
+ */
+export function configWithResolvedLimit<T extends BaseModelConfig>(config: T, modelId: string): T {
```
lizradway: nit: i'm not a fan of this naming, as it is unclear what `withResolvedLimit` means


`resolveContextWindowLimit`?
`enrichConfigWithMetadata` - if we want to add all metadata eventually to config etc

strands-ts/src/models/model.ts
```diff
@@ -26,6 +26,23 @@ import {
...
+  if (config.contextWindowLimit !== undefined) return config
+  const limit = getContextWindowLimit(modelId)
+  if (limit === undefined) return config
+  return { ...config, contextWindowLimit: limit }
```
lizradway, line 44: what happens if a new model drops + isn't included in the list? we currently return undefined, but how are we planning on treating this in proactive compression?

strands-ts/src/models/defaults.ts
opieter-aws: That's a miss, thanks for spotting!

strands-ts/src/models/model.ts
opieter-aws: My plan for proactive compression is to warn once if `threshold` is set, but `contextWindowLimit` is `undefined`. This list is supposed to be a best effort convenience.

strands-ts/src/models/model.ts
opieter-aws: Updated, thanks for the suggestion!

strands-ts/src/models/defaults.ts
opieter-aws: Discussed offline, and we will go with a context window limit dict for now. If we feel the need to expand this, we can do this at a later time

strands-ts/src/models/defaults.ts
opieter-aws: Actually, this is a Prettier behavior we can't override (there's no config option for it). The unquoted `o3`/`o1` keys are Prettier's canonical style

strands-ts/src/models/defaults.ts
```diff
@@ -41,3 +41,140 @@ export function defaultModelWarningMessage(defaultModelId: string): string {
...
+  'gpt-4.1-mini': 1_047_576,
+  'gpt-4.1-nano': 1_047_576,
+  'gpt-4o': 128_000,
+  'gpt-4o-mini': 128_000,
+  'gpt-4-turbo': 128_000,
+  o3: 200_000,
+  'o3-mini': 200_000,
+  'o3-pro': 200_000,
+  'o4-mini': 200_000,
+  o1: 200_000,
```
lizradway, line 136: nonblocking nit: still inconsistent formatted for `o1`/`o3`

strands-ts/src/models/model.ts
lizradway: so proactive compression will not work? im mainly just wondering if we should have a best effort context limit fallback so compression does not silently fail if a user migrates to a newly dropped model

strands-ts/src/models/defaults.ts
```diff
@@ -41,3 +41,140 @@ export function defaultModelWarningMessage(defaultModelId: string): string {
...
+  'gemini-3.1-pro-preview': 1_048_576,
+  'gemini-3.1-flash-lite-preview': 1_048_576,
+}
+
+/**
+ * Known Bedrock cross-region routing prefixes.
+ *
+ * @see https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference.html
+ */
+const BEDROCK_REGION_PREFIXES = new Set(['us', 'eu', 'ap', 'global', 'apac', 'au', 'jp', 'us-gov'])
```
lizradway, line 155: rather than maintain a set of prefixes, can we just do a fallback check if the initial lookup fails that checks the dict for the model id stripped before the first `.`?

i would love to not have to maintain this as new bedrock regions are added

lizradway (approved):

strands-ts/src/models/defaults.ts
opieter-aws: See https://github.com/strands-agents/sdk-typescript/pull/954#discussion_r3173521270 , unfortunately Prettier enforces this

strands-ts/src/models/defaults.ts
opieter-aws: Will note as follow-up!

strands-ts/src/models/model.ts
opieter-aws: Are you okay having this discussion in https://github.com/strands-agents/sdk-typescript/pull/965 ?

strands-ts/src/models/defaults.ts
lizradway: oh oops my bad! weird lol

strands-ts/src/models/model.ts
lizradway: yes yes ofc

PR #938, opened 2026-04-28
feat: expose model on local agent

lizradway:
## Description

Exposes the `model` property on the `LocalAgent` interface so that plugins, hooks, and tools can access the agent's model provider (e.g. for token counting via `model.countTokens()`) without unsafe type casts.

The concrete `Agent` class already has `public model: Model` — this change adds `readonly model: Model` to the `LocalAgent` interface to make it visible through the typed surface that plugins receive. No implementation changes needed.

Brings TypeScript to parity with the Python SDK, where `event.agent.model` is already accessible.

## Related Issues

https://github.com/strands-agents/sdk-typescript/issues/936
https://github.com/strands-agents/sdk-typescript/issues/934

## Type of Change

New feature

## Testing

How have you tested the change?

- [x] I ran `npm run check`
- Type-checks pass cleanly (`tsc --noEmit` on `src/tsconfig.json`)
- No new tests needed — the `Agent` class already satisfies the interface, this just widens the type surface

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [X] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #965, opened 2026-04-29
feat: add proactive context compression to conversation managers

opieter-aws:
## Description

Conversation managers today are entirely reactive. They only reduce context after the model rejects a request with a `ContextWindowOverflowError`. This wastes a round-trip and causes output token starvation as input tokens approach the limit. This PR implements the proactive compression design from [0008-proactive-context-compression](https://github.com/strands-agents/docs/blob/main/designs/0008-proactive-context-compression.md), building on the `projectedInputTokens` from #890.

The `ConversationManager` base class now accepts an optional `compressionThresold` config. When set, it registers a `BeforeModelCallEvent` hook that checks `projectedInputTokens / contextWindowLimit` and calls `reduceOnThreshold()` on the subclass before the model call.
  
  ```typescript
  // Before: only reactive overflow recovery
  new SlidingWindowConversationManager({ windowSize: 50 })
  new SummarizingConversationManager()
  
  // After: proactive compression at 70% context usage
  new SlidingWindowConversationManager({ windowSize: 50, compressProactively: true })
  new SummarizingConversationManager({ compressProactively: { compressionThreshold: 0.8 } })
```  

## Related Issues

strands-agents/sdk-python#555

## Documentation PR

Will follow 

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/strands-ts/src/conversation-manager/conversation-manager.ts
+++ b/strands-ts/src/conversation-manager/conversation-manager.ts
@@ -13,6 +13,9 @@ import type { Model } from '../models/model.js'
 import { logger } from '../logging/logger.js'
 import { warnOnce } from '../logging/warn-once.js'
 
+/** Default compression threshold ratio. */
+const DEFAULT_COMPRESSION_THRESHOLD = 0.7
+
 /**
  * Options passed to {@link ConversationManager.reduce}.
  */
@@ -47,15 +50,15 @@ export type ConversationManagerThresholdOptions = {
 }
 
 /**
- * Configuration for the conversation manager base class.
+ * Configuration for proactive compression when passed as an object.
  */
-export type ConversationManagerConfig = {
+export type ProactiveCompressionConfig = {
   /**
    * Ratio of context window usage that triggers proactive compression.
-   * Value between 0 and 1 (e.g. 0.7 means compress when 70% of the context window is used).
-   * When not set, proactive compression is disabled and only reactive overflow recovery is used.
+   * Value between 0 (exclusive) and 1 (inclusive).
+   * Defaults to 0.7 (compress when 70% of the context window is used).
    */
-  compressionThreshold?: number
+  compressionThreshold: number
 }
 
 /**
@@ -68,9 +71,9 @@ export type ConversationManagerConfig = {
  * the agent loop uncaught. This makes `reduce` a critical operation — implementations
  * must be able to make meaningful progress when called with `error` set.
  *
- * Optionally, a manager can enable proactive compression by setting `compressionThreshold`
- * in the config. When set, the base class registers a `BeforeModelCallEvent` hook that
- * checks projected input tokens against the model's context window limit and calls
+ * Subclasses can enable proactive compression by passing their `compressProactively`
+ * option to the base constructor. When set, the base class registers a `BeforeModelCallEvent`
+ * hook that checks projected input tokens against the model's context window limit and calls
  * {@link ConversationManager.reduceOnThreshold} when the threshold is exceeded.
  *
  * @example
@@ -94,16 +97,22 @@ export abstract class ConversationManager implements Plugin {
 
   protected readonly _compressionThreshold: number | undefined
 
-  constructor(config?: ConversationManagerConfig) {
-    if (
-      config?.compressionThreshold !== undefined &&
-      (config.compressionThreshold <= 0 || config.compressionThreshold > 1)
-    ) {
-      throw new Error(
-        `compressionThreshold must be between 0 (exclusive) and 1 (inclusive), got ${config.compressionThreshold}`
-      )
+  /**
+   * @param compressProactively - Enable proactive compression. `true` uses the default
+   *   threshold, a config object specifies a custom threshold, `undefined`/`false` disables.
+   */
+  constructor(compressProactively?: boolean | ProactiveCompressionConfig) {
+    const threshold =
+      compressProactively === true
+        ? DEFAULT_COMPRESSION_THRESHOLD
+        : compressProactively
+          ? compressProactively.compressionThreshold
+          : undefined
+
+    if (threshold !== undefined && (threshold <= 0 || threshold > 1)) {
+      throw new Error(`compressionThreshold must be between 0 (exclusive) and 1 (inclusive), got ${threshold}`)
     }
-    this._compressionThreshold = config?.compressionThreshold
+    this._compressionThreshold = threshold
   }
 
   /**
@@ -127,7 +136,7 @@ export abstract class ConversationManager implements Plugin {
   /**
    * Proactively reduce the conversation history before a model call.
    *
-   * Called when projected input tokens exceed the configured compressionThreshold
+   * Called when projected input tokens exceed the configured compression threshold
    * of the model's context window limit. Subclasses implement this to reduce
    * context before the model call, avoiding overflow errors.
    *
@@ -144,7 +153,7 @@ export abstract class ConversationManager implements Plugin {
    * calls {@link ConversationManager.reduce} and retries the model call if reduction succeeded.
    * If `reduce` returns `false`, the error propagates out of the agent loop uncaught.
    *
-   * When `compressionThreshold` is configured and the subclass implements `reduceOnThreshold`,
+   * When a compression threshold is configured and the subclass implements `reduceOnThreshold`,
    * also registers a `BeforeModelCallEvent` hook for proactive compression.
    *
    * Subclasses that override `initAgent` MUST call `super.initAgent(agent)` to
@@ -163,7 +172,7 @@ export abstract class ConversationManager implements Plugin {
 
     if (this._compressionThreshold !== undefined && !this.reduceOnThreshold) {
       logger.warn(
-        `conversation_manager=<${this.name}> | compressionThreshold is configured but reduceOnThreshold is not implemented, proactive compression is disabled`
+        `conversation_manager=<${this.name}> | compressProactively is enabled but reduceOnThreshold is not implemented, proactive compression is disabled`
       )
     }
 
--- a/strands-ts/src/conversation-manager/conversation-manager.ts
+++ b/strands-ts/src/conversation-manager/conversation-manager.ts
@@ -44,12 +44,6 @@ export type ConversationManagerThresholdOptions = {
    * The agent instance. Mutate `agent.messages` in place to reduce history.
    */
   agent: LocalAgent
-
-  /**
-   * The model instance for the upcoming call. Used by conversation
-   * managers that perform model-based reduction (e.g. summarization).
-   */
-  model: Model
 }
 
 /**
@@ -106,7 +100,7 @@ export abstract class ConversationManager implements Plugin {
       (config.compressionThreshold <= 0 || config.compressionThreshold > 1)
     ) {
       throw new Error(
-        `compression_threshold=<${config.compressionThreshold}> | must be between 0 (exclusive) and 1 (inclusive)`
+        `compressionThreshold must be between 0 (exclusive) and 1 (inclusive), got ${config.compressionThreshold}`
       )
     }
     this._compressionThreshold = config?.compressionThreshold
@@ -193,7 +187,7 @@ export abstract class ConversationManager implements Plugin {
           logger.debug(
             `projected_tokens=<${event.projectedInputTokens}>, limit=<${contextWindowLimit}>, ratio=<${ratio.toFixed(2)}>, compression_threshold=<${this._compressionThreshold}> | compression threshold exceeded, reducing context`
           )
-          await this.reduceOnThreshold!({ agent: event.agent, model: event.model })
+          await this.reduceOnThreshold!({ agent: event.agent })
         }
       })
     }
--- a/strands-ts/src/conversation-manager/conversation-manager.ts
+++ b/strands-ts/src/conversation-manager/conversation-manager.ts
@@ -7,9 +7,11 @@
 
 import type { Plugin } from '../plugins/plugin.js'
 import type { LocalAgent } from '../types/agent.js'
-import { AfterModelCallEvent } from '../hooks/events.js'
+import { AfterModelCallEvent, BeforeModelCallEvent } from '../hooks/events.js'
 import { ContextWindowOverflowError } from '../errors.js'
 import type { Model } from '../models/model.js'
+import { logger } from '../logging/logger.js'
+import { warnOnce } from '../logging/warn-once.js'
 
 /**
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: strands-ts/src/conversation-manager/conversation-manager.ts [hunk 1]
```diff
@@ -105,5 +166,36 @@ export abstract class ConversationManager implements Plugin {
...
+      )
+    }
+
+    if (this._compressionThreshold !== undefined && this.reduceOnThreshold) {
+      agent.addHook(BeforeModelCallEvent, async (event) => {
+        const contextWindowLimit = event.model.getConfig().contextWindowLimit
+        if (contextWindowLimit === undefined) {
+          warnOnce(
+            logger,
+            `conversation_manager=<${this.name}> | contextWindowLimit is not set on the model, proactive compression is disabled | set contextWindowLimit in your model config`
```
lizradway: should we set a fallback for this instead? maybe a safe bet based on the most recent model?

opieter-aws: We could assume a 200k context window limit default, and still log a warning here. For a 1m context window that's a huge amount of context unused, but assuming a 1m limit makes the proactive compression unusable for 200k limit models.

I am not opposed to giving a default per se, but the behaviour out-of-the-box can be surprising to users.

lizradway: > We could assume a 200k context window limit default, and still log a warning here.

Personally am a fan of that approach. I think vending a default is a better devx than something not working at all because it is blocked by a config change on our behalf.

strands-ts/src/conversation-manager/conversation-manager.ts
```diff
@@ -105,5 +166,36 @@ export abstract class ConversationManager implements Plugin {
...
+          )
+          return
+        }
+
+        if (event.projectedInputTokens === undefined) {
```
lizradway: when would this happen? should we be logging here?

strands-ts/src/conversation-manager/conversation-manager.ts
```diff
@@ -34,6 +36,34 @@ export type ConversationManagerReduceOptions = {
...
+  /**
+   * The agent instance. Mutate `agent.messages` in place to reduce history.
+   */
+  agent: LocalAgent
+
+  /**
+   * The model instance for the upcoming call. Used by conversation
+   * managers that perform model-based reduction (e.g. summarization).
+   */
+  model: Model
```
lizradway: just wanted to note, we may want to find a conclusion for https://github.com/strands-agents/sdk-typescript/pull/938 before merging this through, since if we expose model on localAgent, we can remove this API parameter (which would be a breaking change).

opieter-aws: This type mirrors `ConversationManagerReduceOptions`, since it's essentially a wrapper that omits `error` type. Would we want to diverge the interfaces here?

lizradway: interesting. i think if we can remove the wrapper and just pass agent that would be best, but non-blocking if they dont figure it out today.

strands-ts/src/conversation-manager/conversation-manager.ts
opieter-aws: This shouldn't happen if using the default method with fallbacks. However, if a user overrides the method this field can be `null`. In that case we do not want to block. Good call on adding a debug log here!

strands-ts/src/conversation-manager/conversation-manager.ts
opieter-aws: #938 got merged, removed from the interface

strands-ts/src/conversation-manager/conversation-manager.ts
opieter-aws: Added

strands-ts/src/conversation-manager/conversation-manager.ts
```diff
@@ -34,6 +36,28 @@ export type ConversationManagerReduceOptions = {
   error: ContextWindowOverflowError
 }
 
+/**
+ * Options passed to {@link ConversationManager.reduceOnThreshold}.
+ */
+export type ConversationManagerThresholdOptions = {
```
lizradway: if we choose to keep a separate `reduceOnThreshold`, im wondering if we need this wrapper around `agent`? 

do we see this config expanding in the future with possible extra optional param? if not then it may make more sense to pass just standalone `agent`.

opieter-aws: Discussed that we will not update this, to keep the `reduce` interface the same with `Error` as required input parameter

strands-ts/src/conversation-manager/summarizing-conversation-manager.ts
```diff
@@ -68,6 +74,15 @@ export type SummarizingConversationManagerConfig = {
...
   summarizationSystemPrompt?: string
+
+  /**
+   * Enable proactive context compression before the model call.
+   *
+   * - `true`: compress when 70% of the context window is used (default threshold).
+   * - `{ compressionThreshold: number }`: compress at the specified ratio (0, 1].
+   * - `false` or omitted: disabled, only reactive overflow recovery is used.
+   */
+  compressProactively?: boolean | ProactiveCompressionConfig
```
lizradway: maybe just `proactiveCompression` as noun type for bool param?

`autoCompress`?
`preemptiveReduction`? -- I kinda like this one, since esp in `slidingWindowConversationManager` (and any future conversation managers we add), we are not actually "compressing" context, but reducing it (which follows more to the external method names anyways).

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
```diff
@@ -89,6 +103,16 @@ export class SlidingWindowConversationManager extends ConversationManager {
     return this._reduceContext(agent.messages, error)
   }
 
+  /**
+   * Proactively reduce context by trimming oldest messages.
+   *
+   * @param options - The threshold reduction options
+   * @returns `true` if the history was reduced, `false` otherwise
+   */
+  reduceOnThreshold({ agent }: ConversationManagerThresholdOptions): boolean {
```
lizradway: I'm still pretty hesitant to maintain a duplicate `reduceOnThreshold` on the api that is just a pass-through to `reduceContext` (especially since this means we have to maintain a separate `ConversationManagerThresholdOptions` on the api that is also duplicate).

primarily from the method name `reduceOnThreshold`, i would expect threshold as a parameter in the method. imo this is confusing, but is also a code smell that this duplication is introducing location into the method unnecessarily. 

might be good idea to have an api bar raiser check my intuition on this: @zastrowm @mkmeral any thoughts?

opieter-aws: The alternative is to make `error` an optional field on the existing `ConversationManagerReduceOptions`. That was my original design, but @zastrowm pushed back on this. Happy to reconsider depending on @zastrowm 's response.

strands-ts/src/conversation-manager/conversation-manager.ts
lizradway: reopening because i don't think we added this?

opieter-aws: Got removed during a fix that got pushed, will re-add once we have consensus on the interfaces

strands-ts/src/conversation-manager/conversation-manager.ts
```diff
@@ -105,5 +169,36 @@ export abstract class ConversationManager implements Plugin {
...
       }
     })
+
+    if (this._compressionThreshold !== undefined && !this.reduceOnThreshold) {
+      logger.warn(
+        `conversation_manager=<${this.name}> | compressProactively is enabled but reduceOnThreshold is not implemented, proactive compression is disabled`
+      )
+    }
+
+    if (this._compressionThreshold !== undefined && this.reduceOnThreshold) {
```
zastrowm: Always subscribe, do the check in the handler

strands-ts/src/conversation-manager/conversation-manager.ts
```diff
@@ -67,6 +95,26 @@ export abstract class ConversationManager implements Plugin {
    */
   abstract readonly name: string
 
+  protected readonly _compressionThreshold: number | undefined
+
+  /**
+   * @param compressProactively - Enable proactive compression. `true` uses the default
+   *   threshold, a config object specifies a custom threshold, `undefined`/`false` disables.
+   */
+  constructor(compressProactively?: boolean | ProactiveCompressionConfig) {
```
zastrowm: Make this an options object instead

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
opieter-aws: Decision: we will make `error` on `ConversationManagerReduceOptions` optional, so we do not need `reduceOnThreshold` anymore

opieter-aws:
@agent-of-mkmeral address the comments on this PR

agent-of-mkmeral:
Addressed the review comments in commit [`23f4b5d`](https://github.com/opieter-aws/sdk-typescript/commit/23f4b5d). Here's a summary:

## Changes Made

### 1. Always subscribe to `BeforeModelCallEvent` (per @zastrowm)
- Both hooks are now always registered in `initAgent()`
- The threshold check happens **inside** the handler, not as a condition for registration
- When `compressProactively` is not set, the handler is a no-op (early return)

### 2. Constructor takes options object (per @zastrowm)
- Changed from positional `compressProactively?: boolean | ProactiveCompressionConfig` param
- Now: `constructor(options?: ConversationManagerOptions)` where `ConversationManagerOptions` has `compressProactively`
- Subclasses pass `{ compressProactively: config.compressProactively }` to `super()`

### 3. Remove `reduceOnThreshold`, make `error` optional on `ReduceOptions` (per @opieter-aws decision)
- **Unified interface:** `reduce()` is now called for both reactive and proactive paths
- `error?: ContextWindowOverflowError` distinguishes reactive (error set) vs proactive (undefined)
- Removed `ConversationManagerThresholdOptions` type entirely
- Subclasses no longer need a separate `reduceOnThreshold` implementation
- `SummarizingConversationManager.reduce()` checks `error` to decide whether to rethrow or swallow

### 4. 200k context window fallback with warning (per @lizradway)
- When model doesn't report `contextWindowLimit`, defaults to `200,000`
- Logs via `warnOnce` so users know to set `contextWindowLimit` for accuracy
- Feature works out-of-the-box without requiring model config

### 5. Best-effort error handling in base class
- The `BeforeModelCallEvent` handler wraps proactive `reduce()` in try/catch
- Errors are logged and swallowed — the model call proceeds regardless

## Naming

Kept `compressProactively` for now — @lizradway's suggestions (`preemptiveReduction`, `autoCompress`) are still open for discussion. Happy to rename if you reach consensus.

---

**Tests:** All 74 conversation manager tests pass ✅  
**TypeScript:** Clean compilation with `exactOptionalPropertyTypes` ✅

strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts
```diff
@@ -55,7 +69,7 @@ export class SlidingWindowConversationManager extends ConversationManager {
    * @param config - Configuration options for the sliding window manager.
    */
   constructor(config?: SlidingWindowConversationManagerConfig) {
-    super()
+    super(config?.compressProactively !== undefined ? { compressProactively: config.compressProactively } : undefined)
```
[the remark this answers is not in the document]
opieter-aws: @agent-of-mkmeral address this and rename `compressProactively` to `proactiveCompression`

agent-of-mkmeral:
Done in commit [`be96487`](https://github.com/opieter-aws/sdk-typescript/commit/be96487). Two changes:

### 1. Renamed `compressProactively` → `proactiveCompression`
All source, tests, and JSDoc updated. Noun form aligns with @lizradway's feedback.

### 2. Simplified `super()` calls in subclass constructors
```typescript
// Before (verbose ternary):
super(config?.compressProactively !== undefined ? { compressProactively: config.compressProactively } : undefined)

// After (direct pass-through):
super(config)
```
This works because TypeScript only applies excess property checking to **object literals**, not variables. Since the subclass configs structurally satisfy `ConversationManagerOptions` (they contain `proactiveCompression?: ...`), passing them directly is type-safe even with `exactOptionalPropertyTypes: true`.

---

**Tests:** All 74 pass ✅  
**TypeScript:** Clean compilation ✅

strands-ts/src/conversation-manager/summarizing-conversation-manager.ts
```diff
@@ -97,12 +111,40 @@ export class SummarizingConversationManager extends ConversationManager {
...
    */
   async reduce({ agent, model, error }: ConversationManagerReduceOptions): Promise<boolean> {
-    const resolvedModel = this._model ?? model
+    try {
+      return await this._summarizeOldest(agent, this._model ?? model)
+    } catch (summarizationError) {
+      if (error) {
+        // Reactive: rethrow so the ContextWindowOverflowError propagates
+        logger.error(`error=<${summarizationError}> | summarization failed`)
+        const wrapped = summarizationError instanceof Error ? summarizationError : new Error(String(summarizationError))
```
zastrowm: nit - follow-up: we have a helper for this I think

zastrowm (dismissed):

strands-ts/src/conversation-manager/conversation-manager.ts
```diff
@@ -7,12 +7,26 @@
...
+/** Default context window limit fallback when the model doesn't report one. */
+const DEFAULT_CONTEXT_WINDOW_LIMIT = 200_000
 
 /**
  * Options passed to {@link ConversationManager.reduce}.
+ *
+ * When `error` is set, this is a reactive overflow recovery call — the implementation
+ * MUST remove enough history for the next model call to succeed.
+ *
+ * When `error` is undefined, this is a proactive compression call — best-effort reduction
```
lizradway, line 28: non blocking nit: i think the framing of error -> proactive versus reactive is a confusing framing, since its the location of the call that determines this.

lizradway (dismissed):
lgtm once lint passes

lizradway (approved):

strands-ts/src/conversation-manager/conversation-manager.ts
opieter-aws: Noted, it's a bit awkward with the shared interface indeed. Will see if we can clarify this better in the docs page

PR #982, opened 2026-04-30
feat(hooks): add endTurn decision field to AfterToolsEvent

mathpal:
## Motivation

`AfterToolsEvent` is the only `After*` lifecycle event without a mutable decision field. Hooks that need to halt the agent loop between tool completion and the next model call have no clean way to do so — the only workaround is `agent.cancel()`, which yields `stopReason: 'cancelled'` instead of a clean `'endTurn'`.

This is needed for use cases like budget/quota enforcement, early-exit optimization, or guardrail hooks that determine after inspecting tool results that no further model calls are necessary.

## Public API Changes

`AfterToolsEvent` gains a mutable `endTurn` field (`boolean | string`, default `false`):

```typescript
import { Agent, AfterToolsEvent } from '@strands-agents/sdk'

const agent = new Agent({ model, tools })

// Boolean: halt silently
agent.addHook(AfterToolsEvent, (event) => {
  event.endTurn = true
})

// String: halt with a final assistant message
agent.addHook(AfterToolsEvent, (event) => {
  event.endTurn = 'enough information gathered'
})
```

When `endTurn` is truthy, the agent loop exits with `stopReason: 'endTurn'`. Tool results and any assistant message are always committed to conversation history before halting. The change is backward compatible — the default value is `false`, preserving existing behavior.

## Use Cases

- **Budget/quota enforcement**: A hook inspects cumulative token usage after each tool batch and halts the loop before incurring another model call.
- **Guardrail hooks**: After tools return, a hook evaluates results against policy constraints and stops the agent cleanly.
- **Early-exit optimization**: When tool results already contain a complete answer, skip the redundant model round-trip.

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -906,14 +906,12 @@ export class Agent implements LocalAgent, InvokableAgent {
           // Hook requested halt: exit without calling the model again
           const { afterToolsEvent } = toolsResult
           if (afterToolsEvent.endTurn) {
-            let lastMessage: Message
-            if (typeof afterToolsEvent.endTurn === 'string') {
-              lastMessage = new Message({ role: 'assistant', content: [new TextBlock(afterToolsEvent.endTurn)] })
-              yield this._appendMessage(lastMessage, invocationState)
-            } else {
-              // Boolean true: lastMessage is the assistant message with toolUseBlock items, not natural-language text.
-              lastMessage = assistantMessage
-            }
+            const endTurnText =
+              typeof afterToolsEvent.endTurn === 'string'
+                ? afterToolsEvent.endTurn
+                : 'Turn ended early by hook after tool execution'
+            const lastMessage = new Message({ role: 'assistant', content: [new TextBlock(endTurnText)] })
+            yield this._appendMessage(lastMessage, invocationState)
 
             result = new AgentResult({
               stopReason: 'endTurn',
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -202,6 +202,9 @@ const DEFAULT_AGENT_NAME = 'Strands Agent'
 /** Default identifier assigned to agents when none is provided. */
 const DEFAULT_AGENT_ID = 'agent'
 
+/** Result returned by tool-execution generators, threading the AfterToolsEvent back to the main loop. */
+type ToolsExecutionResult = { message: Message; afterToolsEvent: AfterToolsEvent }
+
 /**
  * Orchestrates the interaction between a model, a set of tools, and MCP clients.
  * The Agent is responsible for managing the lifecycle of tools and clients
@@ -866,12 +869,21 @@ export class Agent implements LocalAgent, InvokableAgent {
           }
 
           // Execute tools
-          const toolResultMessage = yield* this.executeTools(
+          const toolsResult = yield* this.executeTools(
             assistantMessage,
             this._toolRegistry,
             invocationState,
             completedToolResults
           )
+
+          // When the consumer breaks the stream (e.g. agent.cancel() + break),
+          // yield* returns undefined because the inner generator was closed.
+          if (!toolsResult) {
+            this._meter.endCycle(cycleStartTime)
+            this._tracer.endAgentLoopSpan(cycleSpan)
+            continue
+          }
+          const toolResultMessage = toolsResult.message
 
           /**
            * Deferred append: both messages are added AFTER tool execution completes.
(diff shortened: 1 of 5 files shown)


strands-ts/src/hooks/events.ts
```diff
@@ -711,6 +711,15 @@ export class AfterToolsEvent extends HookableEvent {
...
   readonly invocationState: InvocationState
 
+  /**
+   * When set to `true`, the agent loop halts after this tool batch completes
+   * without calling the model again. When set to a string, that string is
+   * appended as a final assistant message before halting.
+   *
+   * In both cases `stopReason` on the returned `AgentResult` is `'endTurn'`.
+   */
+  endTurn: boolean | string = false
```
notowen333, line 752: Instead of overloading one property with two types and meanings, I think splitting the meanings into their own variables will lead to more explicit usage:

```
endTurn?: boolean
endTurnMessage?: string
```

strands-ts/src/agent/agent.ts
```diff
@@ -822,7 +825,16 @@ export class Agent implements LocalAgent, InvokableAgent {
           }
 
           // Execute tools
-          const toolResultMessage = yield* this.executeTools(modelResult.message, this._toolRegistry, invocationState)
+          const toolsResult = yield* this.executeTools(modelResult.message, this._toolRegistry, invocationState)
+
+          // When the consumer breaks the stream (e.g. agent.cancel() + break),
+          // yield* returns undefined because the inner generator was closed.
+          if (!toolsResult) {
+            this._meter.endCycle(cycleStartTime)
```
notowen333, line 882: Ok with the repetition but we could consider a helper for ending a cycle

strands-ts/src/agent/agent.ts
```diff
@@ -822,7 +825,16 @@ export class Agent implements LocalAgent, InvokableAgent {
           }
 
           // Execute tools
-          const toolResultMessage = yield* this.executeTools(modelResult.message, this._toolRegistry, invocationState)
+          const toolsResult = yield* this.executeTools(modelResult.message, this._toolRegistry, invocationState)
+
+          // When the consumer breaks the stream (e.g. agent.cancel() + break),
+          // yield* returns undefined because the inner generator was closed.
```
notowen333, line 880: Makes sense... wondering if it's possible to catch the broken stream and return something structured

strands-ts/src/agent/agent.ts
```diff
@@ -836,6 +848,26 @@ export class Agent implements LocalAgent, InvokableAgent {
           this._meter.endCycle(cycleStartTime)
           this._tracer.endAgentLoopSpan(cycleSpan)
 
+          // Hook requested halt: exit without calling the model again
+          if (toolsResult.endTurn) {
+            let lastMessage: Message
+            if (typeof toolsResult.endTurn === 'string') {
```
notowen333: It's here that I think the split improves the code

strands-ts/src/agent/agent.ts
```diff
@@ -198,6 +198,9 @@ const DEFAULT_AGENT_NAME = 'Strands Agent'
 /** Default identifier assigned to agents when none is provided. */
 const DEFAULT_AGENT_ID = 'agent'
 
+/** Result returned by tool-execution generators, threading the AfterToolsEvent decision back to the main loop. */
+type ToolsExecutionResult = { message: Message; endTurn: boolean | string }
```
notowen333: Should we just include the whole event instead of just the endTurn field?

notowen333 (changes requested):
The added field on the event makes sense to me. A few comments on impl details

strands-ts/src/hooks/events.ts
notowen333: Coupling them because message is dependent on endTurn like

```
endTurn?: {
  message?: string
}
```

Could also make sense but I think `endTurn: {}` is a bit unexpected

strands-ts/src/agent/agent.ts
mathpal: the split makes it a bit weird to deal with. for example: what happens if `endTurnMessage` is set, but `endTurn` is not? 

ideally, we should have hidden the action behind methods like:

```
event.halt()
event.halt("stop")
```

but that might be a breaking change or a special case of this specific hook, both of which aren't ideal. I think sticking to the current approach is the best middle ground that I can see without breaking api.

strands-ts/src/agent/agent.ts
mathpal: sure.

strands-ts/src/agent/agent.ts
mathpal: i don't see a good way to do that or any major issues in not doing this.. did you have something in mind?

notowen333:
Team is aligned on the overloaded `bool | string` in addition to the `endTurn` name. 

Noting also that `cancel` on `BeforeModelCallEvent` could be used for a similar purpose but methodologically, we are ok with expanding bells and whistles on hook events

notowen333:
Hey we're having an issue with CI can you rebase and commit again

notowen333:
Sorry for the churn here: for consistency can we add a default message that we append when endTurn is `true` and no string is provided. Similar to BeforeModelCallEvent

`const cancelText =
        typeof beforeModelCallEvent.cancel === 'string' ? beforeModelCallEvent.cancel : 'model call denied by hook'`

Something like `Turn ended early by hook after tool execution`

notowen333 (approved):

Issue #771, opened 2026-04-01
[FEATURE] Direct tool calling


notowen333:
### Problem Statement

Allow direct tool calling like in python. `agent.tool.calculator(expression="765 / 987", record_direct_tool_call=False)`

### Use Case

Control flows that want deterministic tool calls in the conversation history.

### Additional Context

It might be an acknowledged gap to not support this feature in TS

notowen333:
Being addressed in:

https://github.com/strands-agents/sdk-typescript/pull/985

Issue #471, opened 2026-02-03
Agent - Direct Tool Calls


pgrayy:
## Summary

Allow users to call tools directly through the agent instance. Add configurations around adding tool results to message history.

## Usage

> ⚠️ **Agent-generated example** - This is a preliminary API sketch and has not been finalized.

```typescript
import { Agent } from '@strands-agents/sdk'
import { calculator } from './tools'

const agent = new Agent({
  tools: [calculator],
})

// Direct tool invocation
const result = await agent.tool.calculator({
  expression: '2 + 2',
})

// With options for message history
const result = await agent.tool.calculator(
  { expression: '2 + 2' },
  { addToHistory: true }
)
```

## Documentation

https://strandsagents.com/latest/documentation/docs/api-reference/python/agent/agent/?h=#strands.agent.agent

## Reference

https://github.com/strands-agents/sdk-python/blob/main/src/strands/tools/_caller.py

PR #985, opened 2026-05-01
feat: add direct tool calling via agent.tool accessor

agent-of-mkmeral:
## Summary

Adds programmatic tool invocation to the TS SDK — Python parity for `agent.tool.tool_name(...)`.

## Usage (customer-facing)

```typescript
const agent = new Agent({ tools: [calculator] })

// Get the result
const result = await agent.tool.calculator.invoke({ a: 5, b: 3 })

// Or stream events
for await (const e of agent.tool.calculator.stream({ a: 5, b: 3 })) { ... }

// Skip recording in conversation history (per call or globally)
await agent.tool.calculator.invoke({ a: 5, b: 3 }, { recordDirectToolCall: false })
new Agent({ tools, recordDirectToolCall: false })
```

> The `!` you'll see in this PR's tests (`agent.tool.calc!.invoke(...)`) is only required because our internal `tsconfig` has `noUncheckedIndexedAccess: true`. Customer projects won't need it.

## What's added

- `agent.tool` accessor returning a `ToolHandle` per tool with `.invoke()` / `.stream()`
- Underscore-to-hyphen and case-insensitive name resolution (via new `ToolRegistry.resolve`)
- 3-message recording (assistant ToolUse → user ToolResult → assistant ack) with `MessageAddedEvent` hooks fired
- `AgentConfig.recordDirectToolCall` global default + `DirectToolCallOptions.recordDirectToolCall` per-call
- `Agent.isInvoking` getter + `ConcurrentInvocationError` guard against direct calls during the agent loop
- `ToolNotFoundError` (typed, with `toolName`)

## Files

- `strands-ts/src/agent/tool-caller.ts` (new)
- `strands-ts/src/agent/agent.ts` (`tool` getter, `isInvoking`, `recordDirectToolCall`, encapsulated message-mutation helper passed to ToolCaller)
- `strands-ts/src/registry/tool-registry.ts` (`resolve`)
- `strands-ts/src/errors.ts` (`ToolNotFoundError`)
- `strands-ts/src/index.ts` (exports)
- `strands-ts/src/agent/__tests__/tool-caller.test.ts` (25 tests)
- `AGENTS.md`

## Tests

25/25 tool-caller ✅ · 361/361 agent ✅ · type-check / lint / format ✅

cc @mkmeral @notowen333 @zastrowm


strands-ts/src/agent/agent.ts
```diff
@@ -185,6 +187,12 @@ export type AgentConfig = {
    * Optional unique identifier for the agent. Defaults to "agent".
    */
   id?: string
+  /**
+   * Whether to record direct tool calls in message history.
+   * When true, direct tool calls via `agent.tool.toolName()` are recorded in the
+   * conversation history for context continuity. Defaults to true.
+   */
+  recordDirectToolCall?: boolean
```
mkmeral: why is this an agent param?

mkmeral (commented):
also what about wasm, wit and python bindings?

mkmeral:
@agent-of-mkmeral  update PR based on reviews. also add and test wit python bindings

agent-of-mkmeral:
## API Design Consideration: Choose One Pattern, Not Three

After evaluating the `!` assertion ergonomics, we want to raise a design question: **should we ship the current dot-access Proxy approach, or choose a cleaner alternative?**

We don't want multiple ways to do the same thing. We should pick ONE pattern for direct tool calling in the TS SDK.

<details>
<summary>The 3 alternatives and their tradeoffs</summary>

### Option 1: Current Proxy dot-access (this PR)
```typescript
const result = await agent.tool.calculator!({ a: 5, b: 3 })
```

| Pros | Cons |
|------|------|
| Direct Python parity (`agent.tool.calculator()`) | Requires `!` due to `noUncheckedIndexedAccess: true` in tsconfig |
| Familiar dot-notation | No IDE autocomplete for tool names |
| Proxy is idiomatic JS pattern | The `!` is confusing — users might think the value can be null (it can't) |

---

### Option 2: `agent.callTool('calculator', {...})`
```typescript
const result = await agent.callTool('calculator', { a: 5, b: 3 })
```

| Pros | Cons |
|------|------|
| No `!` needed — clean, explicit | Different API shape from Python SDK |
| Full type safety on return type | Slightly more verbose |
| IDE autocomplete works (method signature) | Tool name as string (no static checking) |
| Simple implementation (no Proxy needed) | Less "magical" (could be a pro) |

---

### Option 3: `agent.tool('calculator')({...})` — callable accessor
```typescript
const result = await agent.tool('calculator')({ a: 5, b: 3 })
```

| Pros | Cons |
|------|------|
| No `!` needed | Unusual double-call pattern |
| Tool name as parameter (no index access) | Different from Python parity |
| `tool` getter becomes a function | Might confuse users ("why two calls?") |
| Proxy can intercept `apply` trap | Slightly harder to understand at first glance |

</details>

### Recommendation

The `!` in Option 1 is a TypeScript-specific ergonomic issue caused by `noUncheckedIndexedAccess: true` in the SDK's tsconfig. It cannot be eliminated while keeping dot-access syntax (this is a fundamental TS limitation, not a fixable implementation detail).

Given we want **one clean way** to do this, the question is:
- Do we value **Python parity** (Option 1, accept the `!`) — or
- Do we value **TS-native ergonomics** (Option 2 or 3, no `!`) and diverge slightly from Python?

Option 2 (`callTool`) is the most TS-idiomatic — explicit, typed, no magic. Option 3 (`agent.tool('calculator')`) is a middle ground that keeps `tool` as the namespace but avoids index access.

Would love to hear thoughts on which direction to take before we continue iterating on this PR.

cc @mkmeral

---

agent-of-mkmeral:
## ✅ Updated: All review feedback addressed + WIT/Python bindings added

Pushed commit `e3c40f5` with the following changes:

### Review Feedback Addressed

| # | Review Comment | Resolution |
|---|---|---|
| 1 | @mkmeral: "why is this an agent param?" (`recordDirectToolCall` on `AgentConfig`) | ✅ **Removed from AgentConfig** — now per-call only via `DirectToolCallOptions`. Default is `true`. |
| 2 | `_toolCaller` should be `readonly` | ✅ `private readonly _toolCaller` |
| 3 | Redundant null check after `_findNormalizedToolName` | ✅ Removed dead code, uses `!` assertion |
| 4 | `Math.random()` → `crypto.randomUUID()` | ✅ Guaranteed unique tool use IDs |
| 5 | `userMessageOverride` is speculative | ✅ **Removed** — simpler API, can add later |
| 6 | `ToolCaller` shouldn't be exported as a value | ✅ Now `export type { ToolCaller }` only |
| 7 | Concurrency test doesn't test the actual scenario | ✅ Added real test: mocks `isInvoking=true`, asserts `ConcurrentInvocationError` |
| 8 | Tests should use `toStrictEqual` | ✅ Full object assertions where applicable |
| 9 | AGENTS.md not updated | ✅ Added `tool-caller.ts` to directory listing |

### WIT/Python Bindings (New)

| Layer | Change |
|---|---|
| **WIT** (`wit/agent.wit`) | Added `call-tool-direct-args`, `call-tool-direct-result` types and `call-tool-direct` method on `agent` resource |
| **WASM** (`strands-wasm/entry.ts`) | Implemented `callToolDirect` → routes to `agent.tool[name](input, opts)` |
| **Python host** (`_wasm_host.py`) | Added `call_tool_direct` + `call_tool_direct_async` methods |
| **Python proxy** (`agent/__init__.py`) | `_ToolProxy.invoke()` now accepts `record_direct_tool_call` kwarg and routes through WASM when available |

### Test Results

- **16/16 tool-caller tests pass** ✅
- **2268/2272 total tests pass** (4 pre-existing OTEL config failures due to `OTEL_SERVICE_NAME` env conflict — not related to this PR)
- **TypeScript compiles cleanly** (no type errors)

### API Design Note

The `!` non-null assertion is still required due to `noUncheckedIndexedAccess: true` in tsconfig:
```typescript
await agent.tool.calculator!({ a: 5, b: 3 })
```
This is a fundamental TypeScript limitation — the Proxy guarantees values at runtime but TS can't statically verify this for index access. My earlier comment about the 3 design options still stands for discussion.

---

agent-of-mkmeral:
**TL;DR**: PASS — 1 medium-severity UX issue found. 46 adversarial tests written, all pass. The implementation is solid for normal tools, MCP tools, concurrency, and edge cases.

<details>
<summary>Adversarial Testing Report</summary>

**Scope**: Direct tool calling via `agent.tool` Proxy accessor — tested normal tools, MCP-like tools, Proxy safety, concurrency, parameter filtering, error propagation, tool registry mutations, and streaming tools.

**Baseline**: 2314/2318 tests pass (4 pre-existing OTEL config failures unrelated to PR)
**Tests written**: 46
**Tests passing**: 46
**Tests failing (findings)**: 0 (finding is behavioral, not a test failure)

| # | Category | Severity | Description |
|---|----------|----------|-------------|
| 1 | Unhandled Edge Case | Medium | Proxy "then" trap makes `agent.tool` accidentally thenable |

<details>
<summary>Finding 1 — Unhandled Edge Case (Medium): Proxy "then" trap makes agent.tool thenable</summary>

**Reproduction**:
```typescript
const agent = new Agent({ model, tools: [] })

// This is what happens when a user does: await agent.tool
// JS runtime checks .then to determine if object is thenable
console.log(typeof agent.tool.then) // "function" — Proxy returns a callable!

// So `await agent.tool` throws:
// Error: "Tool 'then' not found"
// This is extremely confusing for users
```

**Observed behavior**: The Proxy's `get` trap returns a function for ANY string property, including `"then"`. This makes `agent.tool` satisfy JavaScript's thenable interface. If a user accidentally writes `const tools = await agent.tool` (common mistake with async code), they get a cryptic `"Tool 'then' not found"` error instead of a helpful message.

**Expected behavior**: `await agent.tool` should either:
- Return the proxy as-is (not be thenable), or
- Throw a clear error like "agent.tool is not awaitable"

**Fix**: Add a special case in the Proxy get trap:
```typescript
get(target: ToolCaller, prop: string | symbol): ToolCallerFn | unknown {
  if (typeof prop === 'symbol' || prop in target) {
    return (target as unknown as Record<string | symbol, unknown>)[prop]
  }
  
  // Prevent accidental thenable behavior
  if (prop === 'then') return undefined
  
  return (input?: JSONValue, options?: DirectToolCallOptions): Promise<ToolResultBlock> => {
    return target._callTool(prop, input ?? {}, options)
  }
}
```

This is a well-known Proxy pattern — Prisma, tRPC, and other Proxy-based libraries all handle the `then` trap explicitly.

**Impact**: Any user who writes `await agent.tool` (instead of `await agent.tool.name()`) gets a cryptic error. This is especially likely when refactoring async code.

</details>

<details>
<summary>What Survived (46 tests passed)</summary>

✅ **MCP tools**: Hyphenated MCP tools work correctly via underscore normalization (`agent.tool.web_search` → `web-search`). Missing required params pass through (no client-side validation).

✅ **Proxy safety**: `constructor`, `__proto__`, `toString`, `Symbol.iterator`, `Symbol.toPrimitive` all handled correctly — no prototype pollution, no crashes.

✅ **Edge case inputs**: null, undefined, deeply nested objects, special characters (unicode, XSS payloads, SQL injection strings), arrays, very large strings (100KB), numeric/boolean primitives — all handled.

✅ **Concurrency**: 10 sequential calls record correctly (40 messages). Parallel calls with `recordDirectToolCall: false` don't interfere. Parallel calls WITH recording produce valid message sequences (atomic 4-message push).

✅ **Tool registry mutations**: Removing a tool after getting the function reference correctly throws. Adding tools dynamically after agent creation works.

✅ **Parameter filtering**: Empty properties filters everything. No schema passes everything through. Array/null inputs handled gracefully (returned as-is when non-object).

✅ **Error propagation**: Non-Error throws (`throw 'string'`), `throw undefined`, mid-stream errors all propagate correctly. Failed tools don't record messages.

✅ **Name normalization**: Double hyphens, pure underscore names, partial name non-matching, single-char names, max-length (64 char) names all work.

✅ **ToolContext correctness**: Agent reference, input passthrough, empty invocationState, proper toolUseId format — all verified.

✅ **Streaming**: Generator yields are fully consumed before returning the final result.

✅ **Accessor stability**: `agent.tool` returns same reference. Individual tool functions are fresh per access (Proxy behavior).

</details>

</details>

---

agent-of-mkmeral:
## ✅ Second-round review feedback addressed (commit `3d78f1a`)

All remaining issues from the automated review and adversarial testing have been fixed:

| # | Issue | Fix |
|---|---|---|
| 1 | **Proxy `then` trap** — `await agent.tool` throws cryptic "Tool 'then' not found" | ✅ Returns `undefined` for `then` property, preventing accidental thenable behavior. Well-known pattern used by Prisma, tRPC, etc. |
| 2 | **Proxy property collision** — `prop in target` fallthrough lets inherited properties shadow tool names | ✅ Removed `prop in target` check entirely. Now only symbols pass through; all string properties return tool caller functions. |
| 3 | **Python `except Exception: pass`** — silently swallows all WASM errors including legitimate failures | ✅ Narrowed to `(AttributeError, NotImplementedError, ImportError)` — only catches cases where WASM backend isn't functional. |
| 4 | **Python formatting regression** — `f"No tool named ..."` has extra escapes and space | ✅ Reverted to clean formatting |
| 5 | **Spurious double blank line** in `agent.ts` line 253-254 | ✅ Removed extra blank line |

### Test Results
- **16/16 tool-caller tests pass** ✅
- **303/303 agent tests pass** (12 test files) ✅
- TypeScript compiles cleanly (no source errors)

### Open Design Question

The `!` non-null assertion (`agent.tool.calculator!({...})`) question from my [earlier comment](https://github.com/strands-agents/sdk-typescript/pull/985#issuecomment-4367738554) is still open. Options:
1. **Keep dot-access Proxy** (this PR) — Python parity, requires `!`
2. **`agent.callTool('name', {...})`** — TS-idiomatic, no `!`
3. **`agent.tool('name')({...})`** — middle ground

cc @mkmeral — would love your input on which direction to take.

---

mkmeral:
^ we will keep this PR. in our repository, we have a flag, that causes TS to enforce this. so this is not the customer experience, right? @agent-of-mkmeral

agent-of-mkmeral:
Correct! The `noUncheckedIndexedAccess: true` flag is in this repo's tsconfig, so the `!` is only needed in our tests/internal code. Customers using the SDK will have their own tsconfig (typically without that flag), so for them it's just:

```typescript
const result = await agent.tool.calculator({ a: 5, b: 3 })
```

No `!` needed. The exported type `Record<string, ToolCallerFn>` resolves cleanly when `noUncheckedIndexedAccess` is off (which is the default).

✅ Design question resolved — keeping Option 1 (Proxy dot-access). Will keep this PR as-is.

**Current state:**
- All review feedback addressed (commits `e3c40f5`, `3d78f1a`)
- WIT/Python bindings added and tested
- 46 adversarial tests pass
- One note: PR shows **merge conflicts** (mergeable: CONFLICTING) — likely from changes on `main` since we branched. Want me to rebase?

---

agent-of-mkmeral:
**TL;DR:** Rebased on latest `main`, resolved merge conflicts, and addressed all critical/major findings from [issue #162 review](https://github.com/agent-of-mkmeral/strands-coder-private/issues/162).

<details>
<summary>Changes in this update (commit 9ac526f)</summary>

### Rebase
- Resolved merge conflicts with the new **interrupt system** feature (`#784`) that added `Interruptible` to `ToolContext`
- Added proper `interrupt` handler in direct tool call context (throws clear error: "Interrupts are not supported in direct tool calls")

### Critical Finding: Streaming Generator Test
Added test that creates a multi-yield tool (3 progress events before final result) and verifies:
- All yields are consumed (generator fully iterated)
- Final result is correctly returned after all yields

### Major Finding: Parameter Filtering Contract
Added test that proves the core contract:
- Tool receives ALL parameters (including `extra` params not in schema)
- Only the RECORDED message history is filtered to schema-defined params
- Filtering ≠ restriction

### Additional Tests
- **Dynamic tool registration** — tool added via `agent.toolRegistry.add()` after creation works with `agent.tool`

### Lint Fixes
- `crypto.randomUUID()` → `globalThis.crypto.randomUUID()` (ESLint `no-undef`)
- Removed unused `vi` import

### Results
- **19/19 tool-caller tests pass** ✅
- **327/327 agent tests pass** ✅
- Type check ✅, lint ✅, build ✅
- PR should now be **mergeable** (no more conflicts)

</details>

cc @mkmeral

---

strands-ts/src/agent/tool-caller.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    // Create filtered tool use block
+    const filteredToolUse = new ToolUseBlock({
+      toolUseId: toolUse.toolUseId,
+      name: toolUse.name,
+      input: filteredInput,
+    })
+
+    // Create the message sequence
+    const userMsg = new Message({ role: 'user', content: userContent })
```
mkmeral: aren't we adding the tool input params twice to the messages array? once through user content, and another through tool use, right?

strands-ts/src/agent/agent.ts
```diff
@@ -479,6 +483,27 @@ export class Agent implements LocalAgent, InvokableAgent {
...
+   *
+   * Enables calling tools directly without model inference:
+   * ```typescript
+   * const result = await agent.tool.calculator({ a: 5, b: 3 })
+   * ```
+   *
+   * Supports underscore-to-hyphen normalization: `agent.tool.my_tool()` matches a tool named `my-tool`.
+   */
+  get tool(): ToolCallerProxy {
+    return this._toolCaller as unknown as ToolCallerProxy
```
notowen333: Is this unknown cast avoidable?

strands-ts/src/agent/tool-caller.ts
```diff
@@ -0,0 +1,249 @@
...
+  /**
+   * Whether to record this tool call in the agent's message history.
+   * Defaults to `true`. Set to `false` to execute the tool without
+   * affecting conversation context.
+   */
+  recordDirectToolCall?: boolean
+}
+
+/**
+ * Type for the function returned by the Proxy get trap.
```
notowen333: returned by the Proxy get trap?

strands-ts/src/agent/tool-caller.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  constructor(agent: Agent) {
+    this._agent = agent
+
+    // Return a Proxy that intercepts property access to resolve tool names
+    return new Proxy(this, {
+      get(target: ToolCaller, prop: string | symbol): ToolCallerFn | unknown {
+        // Pass through symbol properties (Symbol.toPrimitive, Symbol.iterator, etc.)
+        if (typeof prop === 'symbol') {
+          return (target as unknown as Record<symbol, unknown>)[prop]
```
notowen333: consider

```
    if (typeof prop === 'symbol') {
      return Reflect.get(target, prop, receiver)
    }
```

strands-ts/src/agent/tool-caller.ts
```diff
@@ -0,0 +1,249 @@
...
+   * @throws Error if no tool with the given name exists
+   */
+  private _findNormalizedToolName(name: string): string {
+    // Direct match
+    if (this._agent.toolRegistry.get(name)) {
+      return name
+    }
+
+    // Underscore-to-hyphen normalization
+    if (name.includes('_')) {
```
notowen333: should we also normalize on case? Any other special chars we want to handle?

strands-ts/src/agent/tool-caller.ts
```diff
@@ -0,0 +1,249 @@
...
+   * 3. A user message with the tool result
+   * 4. An assistant message acknowledging the tool call
+   */
+  private _recordToolExecution(toolUse: ToolUse, toolResult: ToolResultBlock): void {
+    const filteredInput = this._filterToolParameters(toolUse.name, toolUse.input)
+
+    // Create user message describing the tool call
+    const inputParameters = JSON.stringify(filteredInput)
+    const userContent = [
+      new TextBlock(`agent.tool.${toolUse.name} direct tool call.\nInput parameters: ${inputParameters}\n`),
```
notowen333: I think we can clean tighten this string. Specifically making clear that direct tool call means that it was generated from declarative code outside of model invocation

strands-ts/test/integ/direct-tool-call.test.ts
```diff
@@ -0,0 +1,241 @@
...
+describe.skipIf(bedrock.skip)('Direct Tool Calling Integration', () => {
+  describe('with vended tools', () => {
+    it('calls calculator tool directly and returns result', async () => {
+      const agent = new Agent({
+        model: bedrock.createModel(),
+        tools: [calculatorTool],
+        printer: false,
+      })
+
+      const result = await agent.tool.calculator!({ operation: 'multiply', a: 7, b: 8 })
```
notowen333: hm are there any other approaches where we don't end up with a forced `!` at every callsite?

notowen333 (changes requested):
Mostly small comments. Didn't give attention to the wasm/wit/py-adapter changes

The `.tool!` pattern is the biggest sticking point on my side. I'm OK to sign off on it if it's the least of many evils :)

strands-ts/test/integ/direct-tool-call.test.ts
mkmeral: we need `!` because of our sdk tsconfig. customers would not need this unless they explicitly change ts configs

strands-ts/src/agent/tool-caller.ts
mkmeral: wdym? this is the behavior we have in python

mkmeral:
@agent-of-mkmeral check comments/reviews ^

strands-ts/src/agent/tool-caller.ts
notowen333: I don't think that "direct tool call" is unambiguous language for the model is all I'm saying

strands-ts/test/integ/direct-tool-call.test.ts
notowen333: great :)

strands-ts/src/agent/__tests__/tool-caller.test.ts
```diff
@@ -0,0 +1,437 @@
...
+          })
+      )
+      const model = new MockMessageModel().addTurn({ type: 'textBlock', text: 'Hello' })
+      const agent = new Agent({ model, tools: [tool] })
+
+      await agent.tool.id_tool!()
+      await agent.tool.id_tool!()
+
+      // Both calls recorded — tool use IDs in messages should be different
+      const toolUse1 = agent.messages[1]!.content[0] as ToolUseBlock
```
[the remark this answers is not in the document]
mkmeral: we should do better assertions in tests

strands-ts/src/agent/agent.ts
```diff
@@ -296,6 +298,8 @@ export class Agent implements LocalAgent, InvokableAgent {
   _interruptState: InterruptState
   /** Strategy for executing tool calls from a single assistant turn. */
   private readonly _toolExecutor: ToolExecutorStrategy
+  /** Direct tool caller. Uses `!` because the ToolCaller constructor returns a Proxy, confusing TS control flow. */
+  private readonly _toolCaller!: ToolCaller
```
notowen333: Instead of explaining the awkwardness with a comment:

Can we encapsulate the proxy cast inside `ToolCaller`? Right now `Agent` has to handle the proxy implementation detail in three places: `_toolCaller!`, `new ToolCaller(this) as unknown as ToolCaller`, and `this._toolCaller as unknown as ToolCallerProxy`.

In the agent:

```
private readonly _toolCaller: ToolCallerProxy

this._toolCaller = ToolCaller.create(this)

get tool(): ToolCallerProxy {
  return this._toolCaller
}
```

In the tool caller:

```ts
// tool-caller.ts
static create(agent: Agent): ToolCallerProxy {
  return new ToolCaller(agent) as unknown as ToolCallerProxy
}

File: strands-wasm/entry.ts [hunk 11]
```diff
@@ -644,6 +644,26 @@ class AgentImpl {
     if (!this.sessionManager) return { tag: 'err', val: { tag: 'no-session-configured' } }
     return { tag: 'err', val: { tag: 'internal', val: 'deleteSession not yet implemented' } }
   }
+
+  /**
+   * Call a tool directly without model inference, via the TS SDK's ToolCaller.
+   * Used by Python bindings for `agent.tool.my_tool()` support.
+   */
+  async callToolDirect(name: string, inputJson: string, recordInHistory: boolean): Promise<{ tag: 'ok'; val: { status: string; content: string } } | { tag: 'err'; val: AgentError }> {
```
notowen333: Again, would avoid any WASM related updates in this PR

notowen333 (changes requested):

strands-ts/src/agent/tool-caller.ts
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  /**
+   * Records a tool execution in the agent's message history.
+   *
+   * Creates a sequence of messages that represent the tool execution:
+   * 1. A user message with the tool use block (model sees what was called)
+   * 2. An assistant message acknowledging the tool use
+   * 3. A user message with the tool result
+   * 4. An assistant message acknowledging the result
```
poshinchen: the actual code below only adds three of these?

strands-ts/src/agent/tool-caller.ts
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    // 3. Case-insensitive match
+    const lowerName = name.toLowerCase()
+    const caseMatch = tools.find((t) => t.name.toLowerCase() === lowerName)
+    if (caseMatch) {
+      return caseMatch.name
+    }
+
+    throw new Error(`Tool '${name}' not found`)
```
poshinchen: Do we want to have a typed error?

notowen333:
@agent-of-mkmeral can you address the most recent comments?

strands-ts/src/agent/tool-caller.ts
```diff
@@ -0,0 +1,270 @@
...
+        // a callable that throws "Tool 'then' not found", which is confusing.
+        if (prop === 'then') {
+          return undefined
+        }
+
+        // Return a function that executes the named tool.
+        // We intentionally do NOT fall through to `prop in target` here — that would
+        // cause tool names that collide with inherited Object properties (e.g.,
+        // 'constructor', 'toString', 'valueOf') to return the wrong value.
+        return (input?: JSONValue, options?: DirectToolCallOptions): Promise<ToolResultBlock> => {
```
zastrowm: Given that we've standardized on `agent.invoke` in TS, does it make sense to do the same here?:

The benefit is that we'd then be able to the returned values as typed objects and expose the other functions as well

```
await agent.tool.calculator.invoke()
await agent.tool.calculator.stream()
```

strands-ts/src/agent/tool-caller.ts
```diff
@@ -0,0 +1,270 @@
...
+        // Pass through symbol properties (Symbol.toPrimitive, Symbol.iterator, etc.)
+        // Uses Reflect.get for proper receiver forwarding.
+        if (typeof prop === 'symbol') {
+          return Reflect.get(target, prop, receiver)
+        }
+
+        // Prevent accidental thenable behavior — if a user writes `await agent.tool`
+        // the JS runtime checks for `.then`. Without this guard, the Proxy would return
+        // a callable that throws "Tool 'then' not found", which is confusing.
+        if (prop === 'then') {
```
zastrowm, line 144: FWIW, this means that if we add a tool named `then` we can't access it :)

strands-ts/src/agent/tool-caller.ts
```diff
@@ -0,0 +1,270 @@
...
+    // tool use block (not duplicated as text).
+    const toolUseMsg = new Message({ role: 'assistant', content: [filteredToolUse] })
+    const toolResultMsg = new Message({ role: 'user', content: [toolResult] })
+    const assistantMsg = new Message({
+      role: 'assistant',
+      content: [new TextBlock(`agent.tool.${toolUse.name} was called.`)],
+    })
+
+    // Add to message history
+    this._agent.messages.push(toolUseMsg, toolResultMsg, assistantMsg)
```
zastrowm: Fire the message added hook?

strands-ts/src/agent/tool-caller.ts
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  /**
+   * Filters input parameters to only include those defined in the tool specification.
+   */
+  private _filterToolParameters(toolName: string, input: JSONValue): JSONValue {
```
zastrowm: What's the purpose of this?  IMHO this is the tools responsibility, not ours

strands-ts/src/agent/tool-caller.ts
```diff
@@ -0,0 +1,270 @@
...
+   * Resolution order:
+   * 1. Exact match
+   * 2. Underscore-to-hyphen substitution (e.g. `my_tool` → `my-tool`)
+   * 3. Case-insensitive match
+   *
+   * @param name - The name to look up
+   * @returns The actual registered tool name
+   * @throws Error if no tool with the given name exists
+   */
+  private _findNormalizedToolName(name: string): string {
```
zastrowm: Can/should this just be done by the tool registry (I think it should)

https://github.com/strands-agents/sdk-typescript/blob/081daec7abf7fd8dcd46345699c9edcdc5ce7854/strands-ts/src/registry/tool-registry.ts#L111

strands-ts/src/agent/tool-caller.ts
notowen333: let's add a comment here but no-op

strands-ts/src/agent/agent.ts
```diff
@@ -2044,6 +2077,21 @@ export class Agent implements LocalAgent, InvokableAgent {
...
 
+  /**
+   * Appends a message to the conversation history and fires MessageAddedEvent hooks.
+   *
+   * Use this from non-streaming contexts (e.g., direct tool calls via {@link ToolCaller})
+   * where the caller cannot yield events into the agent stream.
+   *
+   * @param message - The message to append
+   * @param invocationState - Optional invocation state (defaults to empty)
+   * @internal
```
zastrowm: Not sure I like this FWIW; we should be very careful of how messages are exposed to callers; are internal methods hidden?

We can lean towards a private method (we use that elsewhere) or passing in a helper into ToolCallerProxy since we own it

notowen333 (dismissed):

strands-ts/src/agent/agent.ts
```diff
@@ -486,6 +495,34 @@ export class Agent implements LocalAgent, InvokableAgent {
     return this._toolRegistry
   }
 
+  /**
+   * Whether the agent is currently processing an invocation.
+   */
+  get isInvoking(): boolean {
```
zastrowm, line 511: Sanity check if we want to expose this; I don't think we should be doing it just for tool calling but I'm not opposed to it

strands-ts/src/agent/__tests__/tool-caller.test.ts
zastrowm, line 1: PR description is out of data I think now - per the invoke/stream apis?

strands-ts/src/agent/agent.ts
mkmeral: Overall, I think we should

strands-ts/src/agent/__tests__/tool-caller.test.ts
mkmeral: @agent-of-mkmeral update PR description based on latest info

mkmeral: Agent has updated it, but is this the DevX that we want? I cringed when I read the PR description for updated DevX, it feels like we are writing couple paragraphs to call a tool 😅

zastrowm: For invoke/stream? do we know how often it's used?  Given that it's mostly pre-filling I think I'm okay with it as it's unified with normal tool calling (which has invoke/stream) - the "!" is what annoys me more TBH

agent-of-mkmeral: Tightened the PR description — cut it from ~80 lines down to a usage block + a short "what's added" + files + tests. The verbose "design decisions" tour is gone (it was just relitigating choices already settled in this thread).

On the `!`: that's purely an artifact of *our* `tsconfig.base.json` having `noUncheckedIndexedAccess: true`. Three options if we want to make it nicer:

1. **Status quo** — keep `ToolCallerProxy = Record<string, ToolHandle>`. Customer tsconfigs default to `noUncheckedIndexedAccess: false`, so they write `agent.tool.calc.invoke()` with no `!`. Strict customers (and our own tests) eat the `!`.
2. **Throwing accessor** — expose `agent.tool('calc').invoke(...)`. Single function call, no `!` ever, but loses dot-completion in IDEs and diverges from Python's `agent.tool.calc(...)`.
3. **`getTool` escape hatch** — add `agent.getTool('calc').invoke(...)` *alongside* dot-access. Strict-mode users get a `!`-free path. Two ways to do the same thing though.

My preference is **#1 + acknowledge in docs**. Customer DevX is unaffected, and the `!` only shows up when devs explicitly opt into stricter typing. Happy to do #3 in a follow-up if it bites people.

As for `.invoke`/`.stream` length — agreed it's wordier than `agent.tool.calc({...})` but the unified shape with `agent.invoke` / `agent.stream` is what made @zastrowm's earlier comment land for me. The 90% case is still one line.

mkmeral: > For invoke/stream?

yep. I kinda get the reasoning, but I don't like how it's used. 

@agent-of-mkmeral is there any issue in sdk repos about asking for streaming in direct tool calling?

agent-of-mkmeral: Yep — searched both SDK repos. There's an explicit feature request, plus prior alignment from @zastrowm:

**[sdk-python#1436 — `[FEATURE] Support for Direct Method Call Streaming`](https://github.com/strands-agents/sdk-python/issues/1436)** (open, 3 reactions)
- Filed by @larrylp2: motivating use-case is *Agents-as-Tools* — direct calls to sub-agents need to stream so the parent can surface progress
- @zastrowm on it: *"I'm inclined — I would expect new methods off of the tool caller that can stream results. Both synchronous and asynchronous method calls probably."*
- @emaan-c posted a design doc with the exact `.invoke() / .stream() / .stream_async()` shape this PR uses, and a PR is up against sdk-python

Related:
- [sdk-typescript#771](https://github.com/strands-agents/sdk-typescript/issues/771) — base "direct tool calling" feature request from @notowen333 (this is what PR #985 closes)
- [sdk-typescript#471](https://github.com/strands-agents/sdk-typescript/issues/471) — original Agent direct-tool-calls task
- [sdk-python#1437](https://github.com/strands-agents/sdk-python/issues/1437) — `Support for interrupts in direct method calls` (another reason to have a stream — interrupts are easier to plumb through events than a single resolved result)
- [sdk-python#2261](https://github.com/strands-agents/sdk-python/issues/2261) — `Concurrent Direct Tool Calls`

So there IS demand, AND the Python side is converging on the same `.invoke() / .stream()` shape we have here. Keeping it gives us TS-Python parity on the streaming API the moment Python lands theirs.

If the worry is "is anyone actually going to use `.stream()` on day one?" — probably not heavily. The two clearest customers are:
1. **Agents-as-tools** (the #1436 use case) — surfacing sub-agent events to the parent
2. **Long-running tools** with progress callbacks (DB queries, API scrapes, etc.)

Given this, my read is: keep `.invoke()` / `.stream()` in this PR, ship parity with Python's direction. If `.stream()` ends up being ghost code we can deprecate later, but removing it now means we'd just add it back when Python's #1436 PR lands.

mkmeral: @agent-of-mkmeral  let's keep stream and invoke as is, but I would also like to add a sugar to improve devx. Can we also support `agent.tool.file_read("/path")` for simpler usage to cover majority of the use cases? 

So we'd have
```
agent.tool.file_read("/path")
agent.tool.file_read.invoke("/path")
agent.tool.file_read.stream("/path")
```

emaan-c: Python side is aligned here. strands-agents/sdk-python#1955 on sdk-python implements the same `.stream() / .stream_async()` pattern (rebased and CI-green). Happy to coordinate on any API naming decisions so both SDKs ship consistently.

agent-of-mkmeral: ## ✅ Call sugar added (commit `13e0226`)

@mkmeral done — `agent.tool.foo(input)` now works as sugar for `agent.tool.foo.invoke(input)`. All three forms are supported:

```typescript
await agent.tool.calculator!({ a: 5, b: 3 })           // call sugar — shortest, 90% case
await agent.tool.calculator!.invoke({ a: 5, b: 3 })    // explicit — Python parity
for await (const e of agent.tool.calculator!.stream({ a: 5, b: 3 })) { ... }  // streaming
```

### How it works

`ToolHandle` is now a callable function with `.invoke()` and `.stream()` attached as own properties. The callable form and `.invoke()` share the same private `_callTool()` implementation, so behavior is guaranteed identical — there's literally no way for them to drift.

```typescript
export interface ToolHandle {
  // Callable — sugar for invoke
  (input?: JSONValue, options?: DirectToolCallOptions): Promise<ToolResultBlock>
  // Explicit
  invoke(input?: JSONValue, options?: DirectToolCallOptions): Promise<ToolResultBlock>
  // Streaming
  stream(...): AsyncGenerator<ToolStreamEvent, ToolResultBlock, undefined>
}
```

### Tests
- **29/29 tool-caller** ✅ (+4 new for call sugar: equivalence with `.invoke()`, options arg, ToolNotFoundError, no-input default)
- **365/365 agent** ✅
- type-check / lint / prettier all clean

### @emaan-c — re: sdk-python#1955 alignment

Thanks for the heads-up! TS now ships:
- `.invoke()` / `.stream()` (matches your design from sdk-python#1436)
- Plus a callable shortcut `agent.tool.foo(input)` for the common case

If Python ends up adding the same callable sugar (which would also keep parity with `agent.tool.calculator(...)` from existing Python docs), we're aligned. If Python keeps to `.stream()` / `.stream_async()` only, the explicit `.invoke()` / `.stream()` methods on TS still match 1:1. Either way, no naming conflict — happy to coordinate further if anything shifts.

PR is ready for re-review.

mkmeral: @agent-of-mkmeral revert the last comment, we don't want the sugar, we decided against that.

also kick off another agent to review https://github.com/strands-agents/sdk-python/pull/1955

agent-of-mkmeral: Reverted in `6f2848f` — call sugar is gone, back to `.invoke()` / `.stream()` only.

```typescript
await agent.tool.calculator!.invoke({ a: 5, b: 3 })
for await (const e of agent.tool.calculator!.stream({ a: 5, b: 3 })) { ... }
```

Also kicked off a separate agent to review sdk-python#1955 — will post findings there.

**Verification:** 25/25 tool-caller ✅ · 361/361 agent ✅ · type-check / lint / format clean.

---

notowen333:
@agent-of-mkmeral please address the review agents most recent comments

notowen333 (approved):

mkmeral (approved):

PR #999, opened 2026-05-05
fix: cache unsupported models for bedrocks token counting

opieter-aws:
## Description

When a Bedrock model does not support the `CountTokens` API, it returns a `ValidationException` with the message "The provided model doesn'\''t support counting tokens". Today, every invocation of `countTokens` makes this failing API call before falling back to the heuristic estimator, causing polluted debug logs and Bedrock error logging.

This change caches the model ID on the first failure and skips the API call entirely on subsequent invocations, falling back directly to the base class heuristic.

## Related Issues

NA

## Documentation PR

NA

## Type of Change

Bug fix

## Testing

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -90,6 +90,12 @@ const BEDROCK_CONTEXT_WINDOW_OVERFLOW_MESSAGES = [
 ]
 
 /**
+ * Cache of model IDs that do not support the CountTokens API.
+ * Prevents repeated failing API calls for models that will never support token counting.
+ */
+const UNSUPPORTED_COUNT_TOKENS_MODELS = new Set<string>()
+
+/**
  * Mapping of Bedrock stop reasons to SDK stop reasons.
  */
 const STOP_REASON_MAP = {
@@ -327,6 +333,14 @@ export class BedrockModel extends Model<BedrockModelConfig> {
   private _client: BedrockRuntimeClient
 
   /**
+   * Clears the cache of model IDs that do not support the CountTokens API.
+   * After calling this, the next countTokens invocation will attempt the API again.
+   */
+  static clearCountTokensCache(): void {
+    UNSUPPORTED_COUNT_TOKENS_MODELS.clear()
+  }
+
+  /**
    * Creates a new BedrockModel instance.
    *
    * @param options - Optional configuration for model and client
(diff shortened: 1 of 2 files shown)


strands-ts/src/models/bedrock.ts
```diff
@@ -89,6 +89,12 @@ const BEDROCK_CONTEXT_WINDOW_OVERFLOW_MESSAGES = [
   'prompt is too long',
 ]
 
+/**
+ * Cache of model IDs that do not support the CountTokens API.
+ * Prevents repeated failing API calls for models that will never support token counting.
+ */
+const UNSUPPORTED_COUNT_TOKENS_MODELS = new Set<string>()
```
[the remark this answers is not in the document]
opieter-aws: The module-level cache is intentional. The goal is to avoid the failing API call across all instances for the same model ID, since model support is a property of the service, not the instance.

strands-ts/src/models/bedrock.ts
```diff
@@ -499,7 +519,18 @@ export class BedrockModel extends Model<BedrockModelConfig> {
       logger.debug(`total_tokens=<${response.inputTokens}> | native token count`)
       return response.inputTokens
     } catch (error) {
-      logger.debug(`error=<${error}> | native token counting failed, falling back to estimation`)
+      if (
+        error instanceof Error &&
+        error.name === 'ValidationException' &&
+        error.message.includes("doesn't support counting tokens")
```
notowen333, line 527: nit: would lift this into a constant/a list up at the top with the cache

strands-ts/src/models/bedrock.ts
```diff
@@ -326,6 +332,14 @@ export class BedrockModel extends Model<BedrockModelConfig> {
   private _config: BedrockModelConfig
   private _client: BedrockRuntimeClient
 
+  /**
+   * Clears the cache of model IDs that do not support the CountTokens API.
+   * After calling this, the next countTokens invocation will attempt the API again.
+   */
+  static clearCountTokensCache(): void {
```
notowen333, line 341: Is there a case where we'd need this? It seems like from the impl that we'd never need it

mehtarac (dismissed):

strands-ts/src/models/bedrock.ts
opieter-aws: It gives control to the user to clear the cache and try native token counting again, this would be unaccessible otherwise

notowen333 (dismissed):

mehtarac (approved):

notowen333 (approved):

Issue #897, opened 2026-04-24
[BUG] Swarm and Graph maxSteps default to Infinity — no runaway protection


opieter-aws:
### Steps to Reproduce

**Swarm**
`swarm.ts:121`: `maxSteps: config.maxSteps ?? Infinity`. Python defaults to `max_handoffs=20, max_iterations=20, execution_timeout=900s, node_timeout=300s`.

**Graph**
`graph.ts:113-114`: `maxSteps` defaults to `Infinity`, `maxConcurrency` defaults to `Infinity`. Python requires at least one of `max_node_executions` or `execution_timeout` to be set.

**No cycle detection**
Python has `repetitive_handoff_detection_window` to detect agents cycling between each other. TypeScript has nothing — agents can cycle indefinitely until `maxSteps` (which is Infinity).

### Expected Behavior

- `maxSteps` should have a finite default (e.g., 20 like Python) for both Swarm and Graph
- Exceeding the limit should return a FAILED result, not throw an exception
- Consider adding timeout support and/or cycle detection

### Actual Behavior

- A swarm or graph with buggy handoff/routing logic loops forever
- No timeout protection if an agent hangs
- No cycle detection for repetitive handoffs

### Additional context

This was a finding from an agentic cross-SDK consistency scan between Python and TypeScript SDKs #789. The finding has been triaged, but verify that this needs an actual fix.

Luffy2208:
@opieter-aws @notowen333 Can i work on this?

Luffy2208:
### Summary

Based on the current implementation and the changes already explored, there are two clear parts to this issue:

---

### 1. Immediate / Scoped Fixes (can be done quickly)

The following can be implemented safely without expanding the public API surface:

- Set a **finite default for `maxSteps`** (e.g., 20) in both Swarm and Graph instead of `Infinity`
- Replace step-limit **exceptions with a structured `FAILED` `MultiAgentResult`**
- Preserve existing **event + tracing flow** (no breaking changes)
- Update tests to expect **failed results instead of thrown errors**
- Add basic coverage for **infinite loop protection via maxSteps**

These are **localized changes** and align directly with the current architecture.

---

### 2. Extended Work (timeouts + cycle detection)

The remaining items (execution timeout, node timeout, repetitive/cycle detection) are **not isolated fixes**. They require:

- **New config surface** (e.g., timeouts, detection parameters)
- **Cancellation propagation** (likely via `AbortSignal`)
- Changes across multiple files:
[...]

Issue #893, opened 2026-04-24
[BUG] Bedrock provider missing timeout, throttle detection on initial request, and default retry strategy


opieter-aws:
### Steps to Reproduce

Gap in the Bedrock model provider compared to the Python SDK:

No read timeout (scan finding #2): The Python SDK sets a 120-second read timeout (`DEFAULT_READ_TIMEOUT = 120` in `bedrock.py:57`). The TypeScript SDK creates `BedrockRuntimeClient` with no timeout, so stuck connections hang indefinitely.

### Expected Behavior

A default read timeout (e.g., `requestTimeout: 120_000`) should be set on the `BedrockRuntimeClient`

### Actual Behavior

No timeout — stuck connections hang forever

### Possible Solution

```typescript
new BedrockRuntimeClient({ ...config, requestTimeout: 120_000 })
```

### Additional context

This was a finding from an agentic cross-SDK consistency scan between Python and TypeScript SDKs #789. The finding has been triaged, but verify that this needs an actual fix.

notowen333:
PR is merged

PR #1008, opened 2026-05-05
feat: add timeouts to graph/swarm; bedrock request timeout

notowen333:
## Description

Two independent reliability improvements:

**Bedrock**: the AWS SDK's `requestTimeout` defaults to `0` (disabled), so a stuck Bedrock connection can hang a process indefinitely. We now inject a 120s default into the `BedrockRuntimeClient`'s `requestHandler`. Callers can override via `clientConfig.requestHandler.requestTimeout`. A user-supplied handler *instance* is left untouched because its timeouts are baked in at construction.

**Swarm / Graph**: add a `timeout` (wall-clock ceiling for the whole invocation) and `nodeTimeout` (fallback per-node ceiling) on `SwarmConfig` / `GraphConfig`, plus a per-node `timeout` override on `AgentNodeOptions`. Per-node enforcement composes via `AbortSignal`, forwarded into the agent's existing `cancelSignal` channel — cancellation is cooperative, so tools that don't check the signal can still run past the deadline (documented). All new fields default to `Infinity` (no behavior change); a one-shot `warnOnce` fires when both `maxSteps` and `timeout` are unbounded.

## Related Issues

#897  #893 

## Documentation PR

follow up

## Type of Change

New feature

## Public API Changes

**Bedrock** — `BedrockModelOptions.clientConfig.requestHandler.requestTimeout` now defaults to 120s:

```typescript
// Before: no default — stuck connections could hang indefinitely
new BedrockModel({ region: 'us-west-2' })

// After: 120s default applied automatically; override as needed
new BedrockModel({
  region: 'us-west-2',
  clientConfig: { requestHandler: { requestTimeout: 300_000 } },
})
```

Backward compatible for any request that completes in under 120s. Requests that legitimately take longer (rare, and usually streaming — which doesn't go through this path) will need to raise the limit.

**Swarm / Graph** — new `timeout` and `nodeTimeout` on orchestrator config; new per-node `timeout` on `AgentNodeOptions`:

```typescript
// Bound a swarm's total wall-clock and each node's execution
const swarm = new Swarm({
  nodes: [researcher, writer],
  timeout: 60_000,      // entire swarm must finish in 60s
  nodeTimeout: 20_000,  // any single node has 20s
})

// Per-node override for a known-slow agent
const graph = new Graph({
  nodes: [
    { agent: fastClassifier },                        // uses nodeTimeout
    { agent: slowResearcher, timeout: 120_000 },      // overrides
  ],
  edges: [['fastClassifier', 'slowResearcher']],
  nodeTimeout: 30_000,
})
```

All three default to `Infinity` (no limit). Set `nodeTimeout: Infinity` or a per-node `timeout: Infinity` to opt out while still using other defaults. Enforcement is via `AbortSignal` — cancellation is cooperative, so a tool that neither polls its cancel signal nor forwards it to a cancellable API (e.g. `fetch(url, { signal })`) can run past the deadline.

## Use Cases

- **Bound runaway swarms** — an LLM caught in a handoff loop no longer holds the process open forever; `timeout` throws, `nodeTimeout` interrupts a stuck agent.
- **Mix fast and slow agents in one graph** — per-node `timeout` override lets a known-slow research agent use 2 minutes while the rest of the graph stays tight at 20s.
- **Bedrock clients behind flaky networks** — the default 120s timeout surfaces a hung connection as an error instead of waiting indefinitely, matching what most users expect out of the box.

## Breaking Changes

None intended. The Bedrock 120s default is technically new behavior — a previously-working non-streaming Bedrock request that takes longer than 120s will now fail. In practice, any such request was either streaming (routed through the event-stream handler, unaffected) or already hung. Callers who need a longer ceiling can raise it via `clientConfig.requestHandler.requestTimeout`.

Swarm / Graph timeouts default to `Infinity`, so no behavior change for existing callers beyond an informational `warnOnce` when both `maxSteps` and `timeout` are unbounded.

### DIFF
--- a/strands-ts/src/__fixtures__/agent-helpers.ts
+++ b/strands-ts/src/__fixtures__/agent-helpers.ts
@@ -191,16 +191,17 @@ export function expectAgentResult(options: AgentResultMatcher): AgentResult {
 }
 
 /**
- * Creates a minimal InvokableAgent that sleeps for `delayMs` before resolving.
- * Respects `cancelSignal` by aborting the sleep early. Used to exercise timeout
- * behavior deterministically without spinning up a full Agent.
+ * Creates a minimal InvokableAgent that sleeps for `delayMs` before resolving,
+ * aborting the sleep early when the invocation's `cancelSignal` fires. Used to
+ * exercise timeout and cancellation behavior deterministically without spinning
+ * up a full Agent.
  *
  * @param id - The agent id
  * @param delayMs - How long the agent should sleep before returning
  * @param structuredOutput - Optional structured output (e.g. a swarm handoff). When present,
  *   its `message` field is used as the assistant text.
  */
-export function createSlowAgent(
+export function createCancellableAgent(
   id: string,
   delayMs: number,
   structuredOutput: { agentId?: string; message: string } = { message: 'done' }
--- a/strands-ts/src/__fixtures__/agent-helpers.ts
+++ b/strands-ts/src/__fixtures__/agent-helpers.ts
@@ -5,14 +5,14 @@
 
 import { expect } from 'vitest'
 import type { Agent } from '../agent/agent.js'
-import type { AgentResult } from '../types/agent.js'
+import type { AgentResult, InvokableAgent, InvokeArgs, InvokeOptions } from '../types/agent.js'
 import type { StopReason } from '../types/messages.js'
 import { Message, TextBlock } from '../types/messages.js'
 import type { Role } from '../types/messages.js'
 import { StateStore } from '../state-store.js'
 import type { JSONValue } from '../types/json.js'
 import { ToolRegistry } from '../registry/tool-registry.js'
-import type { HookableEvent } from '../hooks/events.js'
+import type { HookableEvent, StreamEvent } from '../hooks/events.js'
 import type { HookableEventConstructor, HookCallback } from '../hooks/types.js'
 import { expectLoopMetrics, type LoopMetricsMatcher } from './metrics-helpers.js'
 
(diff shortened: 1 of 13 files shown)


strands-ts/src/multiagent/graph.ts
```diff
@@ -117,9 +128,15 @@ export class Graph implements MultiAgent {
...
       maxSteps: config.maxSteps ?? Infinity,
+      timeout: config.timeout ?? Infinity,
+      nodeTimeout: config.nodeTimeout ?? Infinity,
     }
     this._validateConfig()
 
+    if (this.config.maxSteps === Infinity && this.config.timeout === Infinity) {
+      warnOnce(logger, 'graph has no maxSteps or timeout set; execution is unbounded')
+    }
+
```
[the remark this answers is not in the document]
notowen333: A log line is not a breaking change, there's no need to add any additional logic here

strands-ts/src/__fixtures__/agent-helpers.ts
```diff
@@ -189,3 +189,49 @@ export function expectAgentResult(options: AgentResultMatcher): AgentResult {
...
+ * Creates a minimal InvokableAgent that sleeps for `delayMs` before resolving.
+ * Respects `cancelSignal` by aborting the sleep early. Used to exercise timeout
+ * behavior deterministically without spinning up a full Agent.
+ *
+ * @param id - The agent id
+ * @param delayMs - How long the agent should sleep before returning
+ * @param structuredOutput - Optional structured output (e.g. a swarm handoff). When present,
+ *   its `message` field is used as the assistant text.
+ */
+export function createSlowAgent(
```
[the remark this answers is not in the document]
poshinchen: /strands is there a better name than `createSlowAgent`?

strands-ts/src/multiagent/swarm.ts
```diff
@@ -232,14 +253,37 @@ export class Swarm implements MultiAgent {
...
+          state,
+          handoff,
+          multiAgentSpan,
+          invocationState,
+          execController?.signal
+        )
         handoff = nodeResult.structuredOutput as HandoffResult | undefined
 
+        if (execController?.signal.aborted) {
+          throw new Error(`timeout=<${this.config.timeout}>, swarm_id=<${this.id}> | swarm exceeded wall-clock budget`)
```
poshinchen: nit: should the Error message be different from line 267?

strands-ts/src/multiagent/swarm.ts
```diff
@@ -307,16 +353,32 @@ export class Swarm implements MultiAgent {
 
     const nodeInput = this._resolveNodeInput(input, handoff)
 
+    const nodeTimeout = node.timeout ?? this.config.nodeTimeout
```
poshinchen, line 358: nit: these lines are nearly identical in the graph too, could be extracted.

strands-ts/src/multiagent/nodes.ts
```diff
@@ -154,16 +167,25 @@ export interface AgentNodeOptions {
...
+    const { agent, timeout, ...config } = options
 
     super(agent.id, {
       ...config,
       ...(agent.description !== undefined && { description: agent.description }),
     })
 
     this._agent = agent
+    if (timeout !== undefined) {
+      this.timeout = timeout
```
poshinchen, line 190: nit: We have _validateConfig() to check `timeout` and `nodeTimeout` at orchestrator level, but the node level is missing. Then users who set this timeout to 0 or -1 will fire AbortController immediately. However it might not be a blocker since user wouldn't do that.

poshinchen (dismissed):

strands-ts/src/multiagent/swarm.ts
notowen333: tried extracting and I don't think it makes the code clearer. The shapes are too complicated for a few lines repetition for now. Can consider in the future

poshinchen (approved):

PR #1015, opened 2026-05-06
feat: add optional preserveContext property to AgentNode

### LINKED ISSUE #905: [FEATURE] Graph: support stateful node execution (opt-out of snapshot/restore)
### Problem Statement

TypeScript's `AgentNode` always takes a snapshot before execution and restores it in a `finally` block (`nodes.ts:170-201`). There is no way to opt out — nodes are always stateless.

Python's Graph has `reset_on_revisit` (default: `False`), allowing agent state to accumulate across executions. This enables stateful graph patterns where a node builds on its previous work when revisited.

### Use Case

Graph workflows where a node is visited multiple times and should retain context from previous visits. For example:
- An "analyst" node that accumulates findings across multiple passes
- A "reviewer" node that remembers previous review feedback
- Iterative refinement patterns where a node improves its output each visit

### Proposed Solution

Add a `resetOnRevisit` option to `AgentNode` (default: `true` to preserve current safe behavior). When `false`, skip the snapshot/restore cycle.

### Additional Context

The current snapshot/restore approach is a safer default, but the lack of opt-out blocks a significant class of graph workflows.

This was a finding from an agentic cross-SDK consistency scan between Python and TypeScript SDKs #789. The finding has been triaged, but verify that this needs an actual fix.

notowen333:
## Description

`AgentNode` always snapshots the wrapped agent's messages/appState/modelState before each execution and restores them in `finally`, so the node is stateless across visits. This blocks a useful class of graph workflows where a node is revisited and should build on its previous work.

Add an opt-in `preserveContext` flag on `AgentNode` that skips the snapshot/restore cycle. Default is `false` to preserve today's isolation semantics. The flag lives on the node (not the graph) because the snapshot logic is already self-contained in `AgentNode.handle` and the choice is a per-node design decision — one graph can mix accumulator nodes with stateless ones without any orchestrator plumbing. This also makes it work for Swarm for free.

```typescript
// Default: stateless across visits (unchanged)
new AgentNode({ agent })

// Opt in: agent accumulates state across executions
new AgentNode({ agent, preserveContext: true })
```

The flag only applies when the wrapped agent is an `Agent` instance; non-`Agent` `InvokableAgent`s were never snapshotted and aren't affected.

## Related Issues

Resolves #905

## Documentation PR

Follow up

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/strands-ts/src/multiagent/__tests__/nodes.test.ts
+++ b/strands-ts/src/multiagent/__tests__/nodes.test.ts
@@ -132,6 +132,15 @@ describe('AgentNode', () => {
       const timedNode = new AgentNode({ agent, timeout: Infinity })
       expect(timedNode.timeout).toBe(Infinity)
     })
+
+    it('defaults stateful to false', () => {
+      expect(node.stateful).toBe(false)
+    })
+
+    it('stores the stateful flag when provided', () => {
+      const statefulNode = new AgentNode({ agent, stateful: true })
+      expect(statefulNode.stateful).toBe(true)
+    })
   })
 
   describe('handle', () => {
@@ -169,6 +178,21 @@ describe('AgentNode', () => {
 
       expect(agent.messages).toStrictEqual(messagesBefore)
       expect(agent.appState.getAll()).toStrictEqual(stateBefore)
+    })
+
+    it('retains agent messages across executions when stateful', async () => {
+      const model = new MockMessageModel().addTurn(new TextBlock('reply-1')).addTurn(new TextBlock('reply-2'))
+      const statefulAgent = new Agent({ model, printer: false, id: 'stateful-agent' })
+      const statefulNode = new AgentNode({ agent: statefulAgent, stateful: true })
+      const statefulState = new MultiAgentState({ nodeIds: ['stateful-agent'] })
+
+      await collectGenerator(statefulNode.stream([new TextBlock('first')], statefulState))
+      const messagesAfterFirst = statefulAgent.messages.length
+      expect(messagesAfterFirst).toBeGreaterThan(0)
+
+      await collectGenerator(statefulNode.stream([new TextBlock('second')], statefulState))
+
+      expect(statefulAgent.messages.length).toBeGreaterThan(messagesAfterFirst)
     })
 
     it('passes structuredOutputSchema from options to the agent', async () => {
--- a/strands-ts/src/multiagent/__tests__/nodes.test.ts
+++ b/strands-ts/src/multiagent/__tests__/nodes.test.ts
@@ -1,6 +1,7 @@
 import { beforeEach, describe, expect, it } from 'vitest'
 import { z } from 'zod'
 import { Agent } from '../../agent/agent.js'
+import { BeforeInvocationEvent } from '../../hooks/events.js'
 import type { MultiAgentInput } from '../multiagent.js'
 import { MockMessageModel } from '../../__fixtures__/mock-message-model.js'
 import { collectGenerator } from '../../__fixtures__/model-test-helpers.js'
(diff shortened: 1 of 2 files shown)


strands-ts/src/multiagent/nodes.ts
```diff
@@ -189,6 +211,7 @@ export class AgentNode extends Node {
       }
       this.timeout = timeout
     }
+    this.stateful = stateful ?? false
```
zastrowm: Can we throw if the agent isn't Agent here?  That would surface a foot-gun

strands-ts/src/multiagent/nodes.ts
```diff
@@ -156,13 +156,30 @@ export interface AgentNodeOptions {
    * this deadline.
    */
   timeout?: number
+  /**
+   * When `true`, the wrapped agent accumulates state (messages, appState,
+   * modelState) across node executions. Useful for graph patterns where a
+   * node is revisited and should build on its previous work (e.g., an
+   * analyst that accumulates findings, or iterative refinement).
```
pgrayy, line 180: Would would want to revert back to the original state after invocation? Is this mainly for cycles within an invocation? Or are there cross-invocation concerns as well?

strands-ts/src/multiagent/nodes.ts
```diff
@@ -156,13 +156,30 @@ export interface AgentNodeOptions {
...
+   *
+   * When `false` (default), the agent's state is snapshotted before each
+   * execution and restored in `finally`, so the node is stateless across
+   * visits.
+   *
+   * Only applies when the wrapped agent is an {@link Agent} instance;
+   * non-`Agent` `InvokableAgent`s are never snapshotted regardless of this
+   * flag.
+   */
+  stateful?: boolean
```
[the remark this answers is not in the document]
notowen333: `stateful` is a much better name; we don't want to propagate the awkward property (which could have different semantics) than python

strands-ts/src/multiagent/nodes.ts
notowen333: Using `stateful=true` would be used when we want the Agent to retain its context from the previous invocation on a future invocation.

The default false would be what you said: `revert back to the original state after invocation`

This would not be relevant to cycles within the invocation of the node

strands-ts/src/multiagent/nodes.ts
notowen333: Done — `AgentNode` constructor now throws when `stateful: true` is paired with a non-`Agent` `InvokableAgent`. Test added.

strands-ts/src/multiagent/__tests__/nodes.test.ts
```diff
@@ -171,6 +180,21 @@ describe('AgentNode', () => {
...
+      const statefulNode = new AgentNode({ agent: statefulAgent, stateful: true })
+      const statefulState = new MultiAgentState({ nodeIds: ['stateful-agent'] })
+
+      await collectGenerator(statefulNode.stream([new TextBlock('first')], statefulState))
+      const messagesAfterFirst = statefulAgent.messages.length
+      expect(messagesAfterFirst).toBeGreaterThan(0)
+
+      await collectGenerator(statefulNode.stream([new TextBlock('second')], statefulState))
+
+      expect(statefulAgent.messages.length).toBeGreaterThan(messagesAfterFirst)
```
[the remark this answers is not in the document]
notowen333: Added an `appState` retention test that uses a `BeforeInvocationEvent` hook to bump a counter on each invocation, and asserts the counter persists across two `stateful: true` runs.

strands-ts/src/multiagent/nodes.ts
```diff
@@ -173,13 +173,30 @@ export interface AgentNodeOptions {
...
+   * node is revisited and should build on its previous work (e.g., an
+   * analyst that accumulates findings, or iterative refinement).
+   *
+   * When `false` (default), the agent's state is snapshotted before each
+   * execution and restored in `finally`, so the node is stateless across
+   * visits.
+   *
+   * Only applies when the wrapped agent is an {@link Agent} instance;
+   * non-`Agent` `InvokableAgent`s are never snapshotted regardless of this
+   * flag.
```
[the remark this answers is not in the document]
notowen333: Done — TSDoc updated to reflect the constructor throw.

strands-ts/src/multiagent/__tests__/nodes.test.ts
```diff
@@ -171,6 +181,56 @@ describe('AgentNode', () => {
...
+      const statefulState = new MultiAgentState({ nodeIds: ['stateful-agent'] })
+
+      await collectGenerator(statefulNode.stream([new TextBlock('first')], statefulState))
+      expect(statefulAgent.appState.get<{ count: number }>('count')).toBe(1)
+
+      await collectGenerator(statefulNode.stream([new TextBlock('second')], statefulState))
+      expect(statefulAgent.appState.get<{ count: number }>('count')).toBe(2)
+    })
+
+    it('throws when stateful is set with a non-Agent InvokableAgent', () => {
```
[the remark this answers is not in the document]
notowen333: Done — moved into the `describe(constructor, ...)` block alongside the other `stateful` constructor tests.

strands-ts/src/multiagent/nodes.ts
```diff
@@ -190,9 +206,14 @@ export class AgentNode extends Node {
    * See {@link AgentNodeOptions.timeout}.
    */
   readonly timeout?: number
+  /**
+   * Whether the wrapped agent retains state across node executions.
+   * See {@link AgentNodeOptions.stateful}.
+   */
+  readonly stateful: boolean
```
zastrowm: Did we already discuss `stateful` vs `stateless`?

Or alternatively, matching the `agent.toTool` option as `preserveContext`?

notowen333: No we didn't have any such discussion on the naming.

I don't feel strongly except that I prefer stateful to stateless on a boolean to avoid double negations

notowen333: synced offline and we aligned on `preserveContext`

strands-ts/src/multiagent/nodes.ts
```diff
@@ -190,9 +206,14 @@ export class AgentNode extends Node {
    * See {@link AgentNodeOptions.timeout}.
    */
   readonly timeout?: number
+  /**
+   * Whether the wrapped agent retains state across node executions.
```
mkmeral, line 210: @lizradway @opieter-aws  how would this interact with context management, any issues? 

I'm thinking if this is a responsibility of the agent or the graph?

opieter-aws: I don't see problems with context management here. Accumulated context (`true`) would mean higher token usage and more frequent context window management. Offloading tool results should be in the same location for each agent right? @lizradway

opieter-aws (approved):

PR #1016, opened 2026-05-06
feat(mcp): add Symbol.asyncDispose to McpClient to enable await-using cleanup

### LINKED ISSUE #250: Issue 250: Consider exposing `using async` as a usage pattern for MCP clients and automatically disconnect when used this way.
cc: @zastrowm

Basically the following:

```typescript
using async mcpClient = new McpClient()

then...

class McpClient {
   [Symbol.disposeAsync](): Promise<void> { await this._client.close(); await this._transport.close() }
}

```

Right now users must call

```typescript
await mcpClient.disconnect()
```

gautamsirdeshmukh:
## Description

Currently, users must call `await mcpClient.disconnect()` manually to close the underlying MCP SDK client and transport channel.

Adding `Symbol.asyncDispose` enables the `await using` pattern:

```ts
await using client = new McpClient({ transport })
await client.connect()

// automatically disconnects when scope exits
```

No risk – it's safe to combine with manual `disconnect()` calls since calling `close()` multiple times on the same connection is a no-op.

## Related Issues

Closes #250 

## Documentation PR

N/A

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/strands-ts/src/mcp.ts
+++ b/strands-ts/src/mcp.ts
@@ -198,6 +198,14 @@ export class McpClient {
   }
 
   /**
+   * Enables the `await using` pattern for automatic resource cleanup.
+   * Delegates to {@link McpClient.disconnect}.
+   */
+  async [Symbol.asyncDispose](): Promise<void> {
+    await this.disconnect()
+  }
+
+  /**
    * Lists the tools available on the server and returns them as executable McpTool instances.
    *
    * @returns A promise that resolves with an array of McpTool instances.
--- a/strands-ts/src/__tests__/mcp.test.ts
+++ b/strands-ts/src/__tests__/mcp.test.ts
@@ -375,6 +375,12 @@ describe('MCP Integration', () => {
       expect(mockTransport.close).toHaveBeenCalled()
     })
 
+    it('supports Symbol.asyncDispose for await using pattern', async () => {
+      await client[Symbol.asyncDispose]()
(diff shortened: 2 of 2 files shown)


strands-ts/src/mcp.ts
```diff
@@ -197,6 +197,10 @@ export class McpClient {
     this._state = 'disconnected'
   }
 
+  async [Symbol.asyncDispose](): Promise<void> {
```
zastrowm, line 204: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncDispose

reports:

> This feature is not Baseline because it does not work in some of the most widely-used browsers.

Which concerns me a little

gautamsirdeshmukh: `Symbol.asyncDispose` was added in [Node 20.4.0](https://nodejs.org/en/blog/release/v20.4.0#:~:text=e50c3169af%5D%20%2D%20fs%2C%20stream%3A%20initial%20Symbol.dispose%20and%20Symbol.asyncDispose%20support%20(Moshe%20Atlow)%20%2348518), which is the same release that brought in `Symbol.dispose` – we already use the latter in [agent.ts](https://github.com/strands-agents/sdk-typescript/blob/main/strands-ts/src/agent/agent.ts#L407). So, I'm not too concerned about this introducing a new/heightened risk.

zastrowm (approved):

PR #1018, opened 2026-05-06
feat: refine sliding window coversation manager logic

notowen333:
## Description

Improves how `SlidingWindowConversationManager` handles context overflow by truncating oversized tool-result content instead of destroying it.

Previously, overflow recovery replaced any tool-result block with a fixed error message and forced its status to `error`. This misreported successful tool calls to the model on retry and threw away every byte of the result, including the small head/tail that's often the useful part.

This change aligns closer towards the python implementation.

The new behavior:

- Large text items in tool results keep their first and last 200 characters, with a `<truncated chars="N"/>` marker in the middle.
- Image, video, and binary/remote document blocks nested in tool results become textual placeholders that identify format and source (e.g. `bytes`, `url`, `s3`), with byte count for inline bytes.
- Document blocks with `text` or `content` (TextBlock[]) sources have their inner text partially truncated in place.
- Large JSON blocks are replaced with a size placeholder; mid-structure truncation would produce invalid JSON.
- Truncation targets the oldest tool result first so the most relevant recent context is preserved as long as possible.
- Original `status` and `error` fields on the tool-result block are preserved.
- Re-running truncation on already-truncated content is a no-op, enforced by the length threshold rather than a sentinel string — no coupling to the marker format.

When a tool-result is already small enough that truncation wouldn't help, the manager falls through to its existing message-trimming path, so context still shrinks.

### Python Comparison

Same core idea as Python: partial head/tail truncation (200 chars each side) with a middle marker, oldest-first targeting, image/video/document blocks replaced by text placeholders, tool-result `status` preserved. Differences: TS enforces idempotency via a length threshold (450 chars); Python does it via a sentinel string (`"... [truncated:"`).

TS image placeholders distinguish `bytes`/`url`/`s3` sources honestly; Python's only reports bytes and renders URL/S3 images as `0 bytes`. TS also covers `VideoBlock`, `DocumentBlock` (all four source kinds, including partial truncation for text/content sources), and `JsonBlock`, which the Python pass-through skips. Python has features TS still lacks: a `per_turn` proactive management mode for screenshot-heavy loops, and a fallback trim boundary at `assistant(toolUse) + user(toolResult)` pairs when no plain-user boundary exists — both are reasonable follow-ups but out of scope for this diff.

## Documentation PR

follow up

## Type of Change

Bug fix/Feature Update

## Testing

- [x] I ran `npm run check`

Unit tests in `strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts` cover:
- Partial truncation preserves head/tail and emits the expected marker
- Small results pass through unchanged
- Oldest tool-result message is the one truncated
- Already-truncated results are not re-truncated
- `status` and `error` fields survive truncation
- Placeholders for image / video / document (bytes, s3) sources
- Partial truncation for document text and content sources
- JSON blocks: size placeholder when large, pass-through when small

## Checklist

----

### DIFF
--- a/strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts
+++ b/strands-ts/src/conversation-manager/__tests__/sliding-window-conversation-manager.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, vi } from 'vitest'
 import { SlidingWindowConversationManager } from '../sliding-window-conversation-manager.js'
 import {
   ContextWindowOverflowError,
+  ImageBlock,
   Message,
   TextBlock,
   ToolUseBlock,
@@ -65,7 +66,7 @@ describe('SlidingWindowConversationManager', () => {
             new ToolResultBlock({
               toolUseId: 'tool-1',
               status: 'success',
-              content: [new TextBlock('Large tool result content')],
+              content: [new TextBlock('x'.repeat(500))],
             }),
           ],
         }),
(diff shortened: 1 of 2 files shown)


poshinchen (dismissed):
btw, existing diff between ts and python:

Python makes old tool results as "least relevant" and should be truncated first. But TS truncates the newest (last) tool result first ([code](https://github.com/strands-agents/sdk-typescript/blob/9d6ae1a310097815db085f4d3aec6ec8f0057c1b/strands-ts/src/conversation-manager/sliding-window-conversation-manager.ts#L284)), which may be less ideal.

poshinchen (approved):

PR #1032, opened 2026-05-08
feat: cache AccessDenied error for count tokens

opieter-aws:
## Description

Cache `AccessDeniedException` errors from Bedrock's CountTokens API to avoid repeated failing calls. When the caller's IAM role lacks `bedrock:CountTokens` permission, the first call now caches the failure and all subsequent calls skip the API, falling back to heuristic token estimation. The same "try only once" behavior already in place for models that don't support CountTokens (`ValidationException`).

Renames `UNSUPPORTED_COUNT_TOKENS_MODELS` to `SKIP_COUNT_TOKENS_MODELS` to better reflect that the cache covers multiple failure reasons (unsupported model, missing permission), not just one.

## Related Issues

Reported by users deploying with the AWS CDK L3 construct for Bedrock Agent Core, which does not include `bedrock:CountTokens` in the agent execution role policy.

## Documentation PR

N/A

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`
- Added a new unit test `'should cache model ID and skip API call on AccessDeniedException'` that verifies the first call hits the API, caches the error, and the second call skips the API entirely
- Existing test `'should not cache model ID for other errors'` confirms transient errors are still retried

## Checklist

----

### DIFF
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -97,10 +97,10 @@ const BEDROCK_CONTEXT_WINDOW_OVERFLOW_MESSAGES = [
 ]
 
 /**
- * Cache of model IDs that do not support the CountTokens API.
- * Prevents repeated failing API calls for models that will never support token counting.
+ * Cache of model IDs for which CountTokens API calls should be skipped.
+ * Prevents repeated failing API calls that will never succeed for the lifetime of the process.
  */
-const UNSUPPORTED_COUNT_TOKENS_MODELS = new Set<string>()
+const SKIP_COUNT_TOKENS_MODELS = new Set<string>()
 
 /**
  * Mapping of Bedrock stop reasons to SDK stop reasons.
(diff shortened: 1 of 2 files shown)


strands-ts/src/models/bedrock.ts
```diff
@@ -539,15 +539,21 @@ export class BedrockModel extends Model<BedrockModelConfig> {
       logger.debug(`total_tokens=<${response.inputTokens}> | native token count`)
       return response.inputTokens
     } catch (error) {
-      if (
+      if (error instanceof Error && error.name === 'AccessDeniedException') {
+        warnOnce(
```
lizradway, line 543: also wondering why this is warnOnce when the other error is debug?

opieter-aws: AccessDeniedException is an error that the user can fix (misconfiguration), so we want to notify the user actively. 

ValidationException is not actionable, so just introduces noise

lizradway: makes sense

lizradway (approved):

strands-ts/src/models/bedrock.ts
lizradway: might be worth including more explicit instructions but maybe not worth it?

PR #1034, opened 2026-05-08
feat: WIT-first SDK contract and strands-py 2.0.0a1 rewrite

chaynabors:
## Public API Changes

Python package is strands-agents 2.0.0a1. No TS SDK public API changes.

## Changes

- wit/agent.wit is split into twelve per-domain files. Every arm has documentation.
- _generated.py is regenerable from WIT via strandly/scripts/generate_types.py.
- strands-py/pyproject.toml is publishable and depends on upstream wasmtime.
- Python tooling lives at the workspace pyproject.toml.
- strands-dev is renamed to strandly.
- New py-check CI workflow. code-quality.yml and test.yml are renamed to ts-check.yml and ts-test.yml.
- Husky pre-commit no longer runs strands-wasm tests.
- docs/DIVERGENCES.md tracks open TS proposals and strands-py vs upstream differences.
- The old strands-py/strands/ module tree is replaced by src/strands/__init__.py.


strands-py/tests_integ/models/test_model_anthropic.py
mehtarac, line 1: General question regarding integration tests:

Don't we need a lot of these integration tests with the caveat that some of the tsts will be modified (given API changes to python implementation)?

wit/deps/io/io.wit
mehtarac, line 1: nit: I would prefer to call this something else. In bidi we have audioIO, with the vision of extending it with VideoIO, etc. The term would be overloaded or confused with bidi applications.

wit/agent.wit
```diff
@@ -1,290 +1,238 @@
-package strands:agent;
+package strands:agent@0.1.0;
```
mehtarac, line 1: is this 0.1 or 1.0?

strands-wasm/entry.ts
```diff
@@ -1,34 +1,37 @@
...
-/// <reference path="./generated/interfaces/strands-agent-types.d.ts" />
+/// <reference path="./generated/interfaces/strands-agent-api.d.ts" />
+/// <reference path="./generated/interfaces/strands-agent-messages.d.ts" />
+/// <reference path="./generated/interfaces/strands-agent-streaming.d.ts" />
+/// <reference path="./generated/interfaces/strands-agent-models.d.ts" />
+/// <reference path="./generated/interfaces/strands-agent-tools.d.ts" />
+/// <reference path="./generated/interfaces/strands-agent-sessions.d.ts" />
+/// <reference path="./generated/interfaces/strands-agent-conversation.d.ts" />
 /// <reference path="./generated/interfaces/strands-agent-host-log.d.ts" />
 /// <reference path="./generated/interfaces/strands-agent-tool-provider.d.ts" />
```
mehtarac, line 20: nit: can probably remove all these specific comments regarding the generated file path and just have a singular comment addressing that the generated files go in `./generated/interfaces/`

mehtarac (approved):

wit/streaming.wit
```diff
@@ -0,0 +1,389 @@
...
+  record metadata-event {
+    /// Cumulative token usage.
+    usage: option<usage>,
+    /// Cumulative performance metrics.
+    metrics: option<metrics>,
+  }
+
+  /// Single key-value pair attached to a trace. Values are string-typed
+  /// to keep traces compact; structured payloads belong on `message`.
+  record trace-metadata-entry {
```
pgrayy, line 52: I would recommend organizing trace and metric types under a telemetry.wit.

wit/sessions.wit
```diff
@@ -0,0 +1,309 @@
...
+  }
+
+  /// Sliding-window conversation manager state at snapshot time.
+  record sliding-window-state {
+    /// Number of messages dropped from the front of history this lifetime.
+    removed-message-count: s32,
+  }
+
+  /// Summarizing conversation manager state at snapshot time.
+  record summarizing-state {
```
pgrayy, line 91: Should these go in conversation.wit.

wit/models.wit
```diff
@@ -0,0 +1,166 @@
...
+    messages: list<message>,
+    /// Optional system prompt influencing the count.
+    system-prompt: option<prompt-input>,
+    /// Optional tool specs influencing the count.
+    tools: option<list<tool-spec>>,
+  }
+
+  /// Start a streaming generation. Return an async stream; cancellation
+  /// is signalled by the caller dropping the reader.
+  start-stream: func(args: start-stream-args) -> stream<stream-event>;
```
pgrayy, line 162: Should try to be consistent with naming. For example, here we have `start-stream` whereas for `Tool` we have `call-tool`. In my opinion, both should have `stream` for the verb.

wit/agent.wit
```diff
@@ -1,290 +1,238 @@
...
+    /// via `invoke-args.tools`.
     tools: option<list<tool-spec>>,
-    trace-context: option<string>,
+    /// Child agents exposed as tools to this agent's model, registered
+    /// alongside `tools`.
+    agent-tools: option<list<agent-as-tool-config>>,
+    /// Built-in tools to enable. Added to `tools`.
+    vended-tools: option<list<vended-tool>>,
+    /// Built-in plugins to enable.
+    vended-plugins: option<list<vended-plugin>>,
```
pgrayy, line 96: Are users able to pass in custom plugins or is that still to come?

strands-wasm/entry.ts
```diff
@@ -323,20 +312,25 @@ function createModel(config?: ModelConfig, params?: ModelParams): Model<BaseMode
       })
     }
     case 'gemini': {
```
pgrayy, line 314: Would recommend calling this google.

strands-wasm/entry.ts
```diff
@@ -487,160 +418,231 @@ function createSessionManager(config: AgentConfig): SessionManager | undefined {
...
+
+//
+// --- resources -----------------------------------------------------------
+//
+
+/**
+ * TS hook events that should surface on the WIT stream but aren't part
+ * of `agent.stream()`'s async-iterator output. Subscribed via the
+ * plugin's `initAgent` and pushed through a queue the stream drains.
+ */
```
pgrayy, line 482: Why do we do this? If an event doesn't come through agent.stream, how come we still want to stream it across the WIT boundary? Is it so customers can hook on them host side? They should be able to pass hooks across the WIT boundary for the guest (TS) to call right.

Whatever the case, these events are streamed through `agent.stream()`. Unlike Python, every `HookableEvent` in TS is a `StreamEvent` ([src](https://github.com/strands-agents/sdk-typescript/blob/main/strands-ts/src/hooks/events.ts#L89)).

Issue #850, opened 2026-04-21
[HARNESS] MCP Enhancements


gautamsirdeshmukh:
Bug fixes

  - listTools() doesn't paginate [TS only]Calls listTools() once, ignores nextCursor. Servers with >1 page silently lose tools. Python already paginates.
  - Missing tool description falls back to empty string [TS only]Uses empty string fallback instead of generating a default. Python already does this.

  Enhancements (missing fundamentals)

  - Graceful startup (failOpen) [both]One failing MCP server crashes the entire agent. Neither SDK has a fail-open mechanism.
  - Input validation errors as tool results [TS only]No handler for notifications/cancelled, no Symbol.asyncDispose for await-using cleanup. Python has context manager support.
  - Tool prefix and filters [TS only]No prefix or toolFilter on McpClientConfig. Global namespace collisions across servers. Python already has both.
  - Default MCP server log routing [TS only]No handler for notifications/message. Server logs silently dropped. Python has a message_handler.
  - Server instructions and Implementation description [TS only]connect() never captures initializeResult. No public getters. Python already exposes these.
  - Cancellation support and Symbol.asyncDispose [TS only]No handler for notifications/cancelled, no Symbol.asyncDispose for cleanup. Python has context manager support.
  - Consumer reference counting [TS only]No tracking of consumers per McpClient. Destroyed agents leave clients connected. Python has auto-disconnect.

  Agent integration (harness-crucial)

  - mcpClients on AgentConfig [both]MCP clients smuggled through tools list. No explicit field. Lifecycle contracts undefined in TS. Python has consumer tracking.
  - Config file loading with env passthrough [both]Neither SDK loads configs from JSON. No mcpServers format parser. No env passthrough. Manual transport construction required.
  - MCP hook event types [both]No MCP-specific hook events in either SDK. Missing McpLogEvent, McpProgressEvent, McpToolsChangedEvent.
  - tools/list_changed notification handling [both]No handler in either SDK. Tool registries populated once at init, never updated at runtime.
  - Progress callback passthrough [both]Neither SDK surfaces progress notifications through hooks. TS preserves progressToken but doesn't handle responses.

  Elicitation (harness-crucial)

  - Elicitation-as-interrupts bridge [both]Elicitation blocks until complete but interrupts exit the agent loop. No cross-invocation handling in either SDK.
  - Roots callback [both]Neither SDK handles roots/list. MCP servers can't discover accessible directories. Relevant to sandbox boundaries.

[...]

PR #1038, opened 2026-05-11
feat(mcp): handle toolsChanged notifications

gautamsirdeshmukh:
## Description

When an MCP server adds, removes, or changes its tools at runtime, the agent's tool registry now stays in sync automatically.

MCP servers emit notifications/tools/list_changed when their tool list changes. The MCP SDK already handles subscribing to this — it debounces (300ms) and calls a hook you provide at client construction time.

## Changes

- mcp.ts — Pass listChanged.tools.onChanged to the MCP SDK Client constructor; track registered tool names; expose onToolsChanged setter; implement _handleToolsChanged with re-entrance guard and error handling
- agent.ts — In initialize(), register an onToolsChanged callback on each McpClient that removes old tools from the registry and adds the new ones
- mcp.test.ts — Tests for: callback with correct old/new tools, error path (listTools rejects), re-entrance skip

## Related Issues

#850 

## Documentation PR

N/A

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -426,6 +426,10 @@ export class Agent implements LocalAgent, InvokableAgent {
       this._mcpClients.map(async (client) => {
         const tools = await client.listTools()
         this._toolRegistry.add(tools)
+        client.onToolsChanged = (oldTools, newTools): void => {
+          oldTools.forEach((name) => this._toolRegistry.remove(name))
+          this._toolRegistry.addOrReplace(newTools)
+        }
       })
     )
 
--- a/strands-ts/src/mcp.ts
+++ b/strands-ts/src/mcp.ts
@@ -103,6 +103,10 @@ export class McpClient {
   private _disableMcpInstrumentation: boolean
   private _tasksConfig: TasksConfig | undefined
   private _elicitationCallback: ElicitationCallback | undefined
+  private _registeredToolNames = new Set<string>()
+  private _onToolsChanged: ((oldTools: string[], newTools: McpTool[]) => void) | undefined
+  private _refreshingTools = false
+  private _pendingRefresh = false
 
   constructor(args: McpClientConfig) {
     this._clientName = args.applicationName || 'strands-agents-ts-sdk'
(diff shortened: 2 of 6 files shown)


strands-ts/src/registry/tool-registry.ts
```diff
@@ -27,11 +27,27 @@ export class ToolRegistry {
...
     }
   }
 
+  /**
+   * Registers one or more tools, replacing any existing tools with the same name.
+   *
+   * @param tools - Array of tools to register
+   * @throws ToolValidationError If a tool's properties are invalid
+   */
+  addOrReplace(tools: Tool[]): void {
```
notowen333: nit: `newTools` as the parameter name would make this more readable

strands-ts/src/agent/agent.ts
```diff
@@ -426,6 +426,10 @@ export class Agent implements LocalAgent, InvokableAgent {
       this._mcpClients.map(async (client) => {
         const tools = await client.listTools()
         this._toolRegistry.add(tools)
+        client.onToolsChanged = (oldTools, newTools): void => {
```
notowen333, line 429: The current shape doesn't quite work for this... but wondering if we would want to make this behavior configurable so the client passes in a function that also gets the toolRegistry if they want to define it (default same as we have).

I can't think of any great use case, so I'm OK with this current shape where `agent` owns the impl

strands-ts/src/mcp.ts
```diff
@@ -235,9 +250,43 @@ export class McpClient {
...
+  private async _handleToolsChanged(): Promise<void> {
+    if (this._refreshingTools) {
+      this._pendingRefresh = true
+      return
+    }
+    this._refreshingTools = true
+    try {
+      do {
+        this._pendingRefresh = false
+        const oldTools = [...this._registeredToolNames]
```
notowen333, line 277: Am I missing the bit where registeredToolNames is updated with `newTools`?

gautamsirdeshmukh: Handled on line 252

```
this._registeredToolNames = new Set(tools.map((t) => t.name))
```

notowen333 (approved):

PR #1040, opened 2026-05-11
chore: persist guardrails redaction

### LINKED ISSUE #903: [BUG] Guardrail redaction not persisted with default saveLatestOn='invocation' setting
### Steps to Reproduce

The guardrail redaction hook is only registered when `saveLatestOn='message'`. With the default `saveLatestOn='invocation'` setting, redactions are not immediately persisted.

Python always persists redactions immediately regardless of session strategy.

### Expected Behavior

Guardrail redactions should be persisted immediately regardless of the `saveLatestOn` setting. Redacted content (PII, sensitive data) should never remain in an unpersisted state where a crash could expose the pre-redaction content.

### Actual Behavior

With default settings, redacted messages are only persisted at the end of the invocation. If the process crashes mid-invocation, the session store may contain unredacted content.

### Possible Solution

Always register the redaction persistence hook regardless of `saveLatestOn` setting, or document that `saveLatestOn: 'message'` is required when using guardrail redaction.

### Additional Context

This was a finding from an agentic cross-SDK consistency scan between Python and TypeScript SDKs #789. The finding has been triaged, but verify that this needs an actual fix.

poshinchen:
## Description

Fixes #903, guardrail redactions were only persisted immediately when `SessionManager` was configured with `saveLatestOn: 'message'`. Under the default `'invocation'` strategy (and `'trigger'`), the `AfterModelCallEvent` redaction hook was never registered, so a crash mid-invocation could leave sensitive pre-redaction content in agent memory without ever being snapshotted at the reverse.

Python's SDK always persists redactions inline regardless of session strategy (`src/strands/agent/agent.py` → `redact_latest_message` on the session manager). This PR brings the TS SDK in line.

### Change

- `session-manager.ts`: hoist `AfterModelCallEvent` hook registration out of the `saveLatestOn === 'message'` branch so it always runs. Redaction persistence is a correctness concern, not a save-frequency tradeoff.
- `session-manager.ts`: updated `SaveLatestStrategy` JSDoc to note that redactions always persist immediately.
- `bedrock.ts`: removed the now-obsolete caveat recommending `saveLatestOn: 'message'` for sensitive content.
- `session-manager.test.ts`: replaced the test that codified the buggy behavior with a parameterized `it.each` verifying redaction saves under all three strategies (`invocation`, `trigger`, `message`), plus a narrower test confirming non-redaction `AfterModelCallEvent`s are still a no-op under `invocation`.

## Related Issues

Closes #903
Related to #789 (cross-SDK consistency scan)

## Type of Change

Bug fix

## Testing

- [x] I ran `npm run check`
- Session manager unit tests: 50 pass (including the 3 new parameterized cases)
- Full telemetry suite: 178 pass
- Full build: clean

## Checklist
- [x] I have updated the documentation accordingly (JSDoc)

----


notowen333 (commented):
The PR makes sense to me, but it is technically a breaking change. As long as we are OK with that, we can ship

strands-ts/src/session/session-manager.ts
```diff
@@ -119,12 +120,12 @@ export class SessionManager implements Plugin, MultiAgentPlugin {
...
-      // Also listen to AfterModelCallEvent when saving per-message to ensure
-      // message modifications (e.g., guardrail redactions) are persisted immediately
-      agent.addHook(AfterModelCallEvent, async (event) => {
-        await this._onAfterModelCall(event)
-      })
     }
+
+    // Always persist guardrail redactions immediately, regardless of saveLatestOn.
+    agent.addHook(AfterModelCallEvent, async (event) => {
+      await this._onAfterModelCall(event)
```
zastrowm: Won't this this result in a double-snapshot save for the `invocation` case?  Or rather what is _onAfterAgentInvocation doing that we're not catching this in there?

strands-ts/src/session/session-manager.ts
```diff
@@ -34,9 +34,10 @@ import type { Swarm } from '../multiagent/swarm.js'
  *
  * `SaveLatestStrategy` controls how frequently `snapshot_latest` is updated:
  * - `'invocation'`: after every agent invocation completes (default; balances durability and I/O)
- * - `'message'`: after every message added and after model calls with guardrail redactions
- *   (most durable, highest I/O)
+ * - `'message'`: after every message added (most durable, highest I/O)
  * - `'trigger'`: only when a `snapshotTrigger` fires (or manually via `saveSnapshot`)
+ *
+ * Guardrail redactions are always persisted immediately, regardless of strategy.
```
zastrowm: Why would we fire this in the trigger case - shouldn't the caller always be in control in that case?  Is there ever another case where `trigger` snapshots without the snapshotTrigger fires?

strands-ts/src/session/session-manager.ts
poshinchen: yes, but this cheap imo. The possible fixes:
1. Dedupe at the SessionManager level (skip the invocation-end save if a redaction-save already happened this turn).
2. Accept it as cheap, since `snapshot_latest` is one extra overwrite.

strands-ts/src/session/session-manager.ts
poshinchen: Makes sense, but users will get unredacted messages if something breaks in the middle. I'll make the change and add the doc saying that this is owner's responsibility for the risk.

zastrowm (approved):

PR #1043, opened 2026-05-11
chore: serialized interrupts and structuredOutput as JSON and citationsBlock

poshinchen:
## Description

Aligns `AgentResult.toString()` with the Python SDK's `__str__` priority order so parent agents and direct callers see interrupt and structured output content instead of falling back to (often empty) last-message text.

New priority order:
1. `interrupts` → `JSON.stringify(this.interrupts)` when any are present
2. `structuredOutput` → `JSON.stringify(this.structuredOutput)` when defined
3. Text walk over `lastMessage.content`, now including `citationsBlock` alongside `textBlock` and `reasoningBlock`

## Notes:
- `citationsBlock` uses `'text' in c` guard so the code stays safe if `CitationGeneratedContent` gains non-text variants.
- `reasoningBlock` with the `💭` prefix is preserved (TS-only enrichment, not in Python).
- Non-breaking on the common path: plain text results produce the same string as before.

## Type of Change

Bug fix

## Testing

Added unit tests in `src/types/__tests__/agent.test.ts` covering:
- Interrupts present → JSON output, takes priority over text
- Empty interrupts array → falls through to text
- `structuredOutput` present → JSON output, takes priority over text
- `citationsBlock` content → text appended to output

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/strands-ts/src/types/__tests__/agent.test.ts
+++ b/strands-ts/src/types/__tests__/agent.test.ts
@@ -4,6 +4,8 @@ import { AgentMetrics } from '../../telemetry/meter.js'
 import { AgentTrace } from '../../telemetry/tracer.js'
 import { Message } from '../messages.js'
 import { TextBlock, ReasoningBlock, ToolUseBlock, ToolResultBlock, CachePointBlock } from '../messages.js'
+import { CitationsBlock } from '../citations.js'
+import { Interrupt } from '../../interrupt.js'
 
 describe('AgentResult', () => {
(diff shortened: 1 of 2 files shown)


notowen333 (approved):
LGTM

Confirmed the __str__ in python has the same interrupt/structured output short circuit: https://github.com/strands-agents/sdk-python/blob/1847faec4fd37d2458156b6147c996826259377a/src/strands/agent/agent_result.py#L56

Issue #872, opened 2026-04-22
Add streaming event/hook for interrupts so that folks know that something has been interrupted


zastrowm:
> We should consider emitting an InterruptEvent. In Python, we have a ToolInterruptEvent, but maybe we can generalize. This could be considered for follow up though. 

 _Originally posted by @pgrayy in [#784](https://github.com/strands-agents/sdk-typescript/pull/784/changes#r3053565089)_

Issue #482, opened 2026-02-03
Interrupts - Graph


pgrayy:
## Summary

Support `BeforeNodeCallEvent`, agent node, and multi-agent node interrupts in Graph orchestration to enable human-in-the-loop workflows at any node.

## Usage

> ⚠️ **Agent-generated example** - This is a preliminary API sketch and has not been finalized.

```typescript
import { Graph, BeforeNodeCallEvent } from '@strands-agents/sdk'

const graph = new Graph({
  nodes: {
    analyze: analyzeAgent,
    decide: decisionAgent,
    execute: executionAgent,
  },
  edges: [
    ['analyze', 'decide'],
    ['decide', 'execute'],
  ],
  hooks: {
    beforeNodeCall: async (event: BeforeNodeCallEvent) => {
      if (event.node.name === 'execute') {
        event.interrupt({
          name: 'node_approval',
          reason: `Approve execution of node '${event.node.name}'?`,
          response: undefined,
        })
      }
    },
  },
})

const result = await graph.invoke('Analyze and execute the deployment')

if (result.status === 'INTERRUPTED') {
  console.log('Interrupt:', result.interrupt)
  
  // Resume with user approval
  const resumed = await graph.invoke({
    interruptResponse: {
      name: 'node_approval',
      response: { approved: true },
    },
  })
}
```

## Documentation

https://strandsagents.com/latest/documentation/docs/user-guide/concepts/interrupts/#graph

## Reference

https://github.com/strands-agents/sdk-python/blob/main/src/strands/multiagent/graph.py

Issue #481, opened 2026-02-03
Interrupts - Swarm


pgrayy:
## Summary

Support `BeforeNodeCallEvent` and agent node interrupts in Swarm orchestration to enable human-in-the-loop workflows for agent handoffs.

## Usage

> ⚠️ **Agent-generated example** - This is a preliminary API sketch and has not been finalized.

```typescript
import { Swarm, BeforeNodeCallEvent } from '@strands-agents/sdk'

const swarm = new Swarm({
  agents: [triageAgent, paymentAgent, supportAgent],
  hooks: {
    beforeNodeCall: async (event: BeforeNodeCallEvent) => {
      if (event.targetAgent.name === 'payment_agent') {
        event.interrupt({
          name: 'handoff_approval',
          reason: 'Approve handoff to payment agent?',
          response: undefined,
        })
      }
    },
  },
})

const result = await swarm.invoke('Process my refund request')

if (result.stopReason === 'interrupt') {
  console.log('Interrupt:', result.interrupt)
  
  // Resume with user approval
  const resumed = await swarm.invoke({
    interruptResponse: {
      name: 'handoff_approval',
      response: { approved: true },
    },
  })
}
```

## Documentation

https://strandsagents.com/latest/documentation/docs/user-guide/concepts/interrupts/#swarm

## Reference

https://github.com/strands-agents/sdk-python/blob/main/src/strands/multiagent/swarm.py

PR #1044, opened 2026-05-11
feat: support multi-agent interrupts; add InterruptEvent

notowen333:
## Description

Extends the single-agent interrupt feature to the `Graph` and `Swarm` multi-agent orchestrators. A tool or hook callback anywhere in the tree can now pause execution, surface the pending interrupt through the aggregate result, and resume on a fresh orchestrator invocation with `InterruptResponseContent[]`. Orchestrator-level hooks get their own interruption point via `BeforeNodeCallEvent.interrupt()`, so a node can be gated before it runs.

### Motivation

Interrupts were load-bearing for human-in-the-loop flows at the single-agent level but inaccessible to anyone using `Graph` or `Swarm`. A child agent's `stopReason: 'interrupt'` was ignored by the orchestrator aggregation, and `MultiAgentInput` explicitly rejected response inputs (the old code carried a comment "Graph interrupts will be handled separately"). This PR makes that a fully supported flow: interrupts bubble up through any orchestrator depth, are routable by id on resume, and survive process boundaries via the existing `SessionManager` snapshot machinery.

### Public API changes

**New `Status.INTERRUPTED` and `ResultStatus`.** `NodeResult.status` and `MultiAgentResult.status` can now be `'INTERRUPTED'`. Precedence in `MultiAgentResult._resolveStatus` is `FAILED > INTERRUPTED > CANCELLED > COMPLETED` — parallel-graph short-circuit aborts siblings as `CANCELLED`, but the actionable "resume me" signal is what surfaces at the top.

**`NodeResult.interrupts` / `MultiAgentResult.interrupts`.** Aggregated across the tree:

```typescript
const result = await graph.invoke('do a thing')
if (result.status === Status.INTERRUPTED) {
  for (const interrupt of result.interrupts!) {
    // interrupt.id, interrupt.name, interrupt.source, interrupt.reason
  }
}
```

`source` is a new union: `'tool' | 'hook' | 'multiagent-hook'`. It lets consumers filter by origin without subscribing to different event types.

**Resume via `invoke([InterruptResponseContent, ...])`** on `Graph`/`Swarm`:

```typescript
const final = await graph.invoke([
  { interruptResponse: { interruptId: interrupt.id, response: 'approved' } }
])
```

Responses route to whichever node holds an `Interrupt` with the matching id. Works same-instance (retained via an in-memory stash) or across a `SessionManager` restore. Unknown ids throw.

**`BeforeNodeCallEvent.interrupt()`** — orchestrator-level hooks can now pause a node before it runs, mirroring the tool/hook interrupt API on single-agent:

```typescript
graph.addHook(BeforeNodeCallEvent, (event) => {
  const { approved } = event.interrupt<{ approved: boolean }>({
    name: 'node_approval',
    reason: `approve ${event.nodeId}?`,
  })
  if (!approved) event.cancel = 'rejected'
})
```

On first pass the hook throws `InterruptError` and the node never runs. On resume with the matching response, `event.interrupt(...)` returns the stored response synchronously and execution continues. Handled in the orchestrator layer, not inside the wrapped agent, so nodes that never ran don't accumulate phantom agent state.

**New `InterruptEvent`** in the hook event stream. Emitted once per interrupt before the interrupt result materializes on `AgentResult`:

```typescript
agent.addHook(InterruptEvent, (event) => {
  // event.interrupt.source === 'tool' | 'hook'
})
```

A single event type discriminated by `interrupt.source` rather than per-origin event classes — consumers almost always observe "something interrupted" uniformly (logging, tracing, metrics, UI updates). The source is a filter, not a subscription boundary.

**`MultiAgentInvokeOptions.cancelSignal`.** `Graph` and `Swarm` now accept a `cancelSignal`, composed with internal signals via `AbortSignal.any`. Nested orchestrators inherit external cancellation. Closes a pre-existing gap where a nested `MultiAgentNode` could run past an outer timeout/abort.

### Design notes

**`_pendingInput` on `MultiAgentState`.** Hook-gated source nodes that never ran on the first invocation need their original content input replayed on resume — nothing else carries it. The node's `resumeSnapshot` is empty (nothing happened yet), there are no upstream results to derive input from, and the resume call only supplies `InterruptResponseContent[]`. The input belongs to the invocation, not to any single node — parallel source nodes share it, downstream nodes that get scheduled on resume after an upstream completes can also need it. Lives on `MultiAgentState` (invocation-scoped), underscore-prefixed and `@internal`-tagged. Serialization rides the existing `stateToJSONSymbol` path, so SessionManager round-trip is automatic.

**Nested multi-agent interrupt resume — follow-up.** This PR does not solve cross-orchestrator interrupt resume when a nested orchestrator (a `MultiAgentNode` wrapping an inner `Graph` or `Swarm`) has no `SessionManager` of its own. The interrupt bubbles up correctly on the first run; on resume the outer routes the response to the nested `MultiAgentNode`, which spins up a fresh inner `MultiAgentState` whose `NodeState.interrupts` don't contain the original id — `groupInterruptResponsesByNode` throws and the nested node ends `FAILED`. Pinned by a test in `src/multiagent/__tests__/interrupts.test.ts`. The fix requires making nested orchestrator state part of the outer snapshot, or propagating the `SessionManager` down — larger shape changes than this PR should carry. The single-SessionManager workflow (outer with `sessionManager`, nested without) works today; the gap is real but narrow.

## Related Issues

Resolves:

- strands-agents/sdk-typescript#481
- strands-agents/sdk-typescript#482 
- strands-agents/sdk-typescript#872

## Documentation PR

Follow Up

## Type of Change

New feature

## Testing

236 multi-agent unit tests and 16 multi-agent integration tests cover: tool-callback interrupts inside a child agent, orchestrator-hook interrupts via `BeforeNodeCallEvent.interrupt()`, parallel short-circuit with cancelled siblings, same-instance resume, `SessionManager` `FileStorage` round-trip (including `ContentBlock[]` input), and nested-orchestrator interrupt propagation.

One Bedrock integ test (`Swarm + hook-gate + session round-trip`) was dropped because it triggered a pre-existing `Agent` bug in the structured-output forced-retry path (not introduced by this PR). Writeup attached as a separate follow-up issue. Coverage for that path is retained at the unit level via `MockSnapshotStorage`; `FileStorage` round-trip is covered by the Graph + tool + session test, which exercises `ContentBlock[]` input through the JSON serialization boundary.

- [x] I ran `npm run check`

## Checklist

---

### DIFF
--- a/strands-ts/src/__tests__/interrupt.test.ts
+++ b/strands-ts/src/__tests__/interrupt.test.ts
@@ -16,6 +16,7 @@ describe('Interrupt', () => {
       name: 'confirm_action',
       reason: 'Please confirm',
       response: 'approved',
+      source: 'hook',
     })
 
     // response is mutable after construction
@@ -41,7 +42,7 @@ describe('Interrupt', () => {
     const interrupt = new Interrupt({ id: 'int-1', name: 'test' })
 
     const json = interrupt.toJSON()
-    expect(json).toStrictEqual({ id: 'int-1', name: 'test' })
+    expect(json).toStrictEqual({ id: 'int-1', name: 'test', source: 'hook' })
     expect('reason' in json).toBe(false)
     expect('response' in json).toBe(false)
   })
@@ -81,7 +82,7 @@ describe('InterruptState', () => {
 
       const interrupt = state.getOrCreateInterrupt('int-1', 'test', 'reason')
 
-      expect(interrupt).toEqual({ id: 'int-1', name: 'test', reason: 'reason' })
+      expect(interrupt).toEqual({ id: 'int-1', name: 'test', reason: 'reason', source: 'hook' })
       expect(state.interrupts['int-1']).toBe(interrupt)
       expect(state.getInterruptsList()).toStrictEqual([interrupt])
     })
@@ -119,6 +120,7 @@ describe('InterruptState', () => {
         name: 'confirm',
         reason: 'reason',
         response: 'pre-approved',
+        source: 'hook',
       })
     })
 
@@ -135,6 +137,7 @@ describe('InterruptState', () => {
         name: 'confirm',
         reason: 'reason',
         response: 'user response',
+        source: 'hook',
       })
     })
   })
--- a/strands-ts/src/__tests__/interrupt.test.ts
+++ b/strands-ts/src/__tests__/interrupt.test.ts
@@ -258,18 +258,24 @@ describe('interruptFromAgent', () => {
     const state = new InterruptState()
     const agent = createMockAgent(state)
 
-    const result = interruptFromAgent(agent, 'test-id', {
-      name: 'confirm',
-      reason: 'need approval',
-      response: 'pre-approved',
-    })
+    const result = interruptFromAgent(
+      agent,
+      'test-id',
+      {
+        name: 'confirm',
+        reason: 'need approval',
+        response: 'pre-approved',
+      },
+      'tool'
+    )
 
     expect(result).toBe('pre-approved')
-    expect(state.interrupts['test-id']).toEqual({
+    expect(state.interrupts['test-id']).toMatchObject({
       id: 'test-id',
       name: 'confirm',
       reason: 'need approval',
       response: 'pre-approved',
+      source: 'tool',
     })
   })
 
@@ -280,14 +286,19 @@ describe('interruptFromAgent', () => {
 
     const agent = createMockAgent(state)
 
-    const result = interruptFromAgent(agent, 'test-id', {
-      name: 'confirm',
-      reason: 'need approval',
-      response: 'preemptive',
-    })
+    const result = interruptFromAgent(
+      agent,
+      'test-id',
+      {
+        name: 'confirm',
+        reason: 'need approval',
+        response: 'preemptive',
+      },
+      'tool'
+    )
 
     expect(result).toBe('user-provided')
-    expect(state.interrupts['test-id']).toEqual({
+    expect(state.interrupts['test-id']).toMatchObject({
       id: 'test-id',
       name: 'confirm',
       reason: 'need approval',
@@ -299,16 +310,22 @@ describe('interruptFromAgent', () => {
     const state = new InterruptState()
     const agent = createMockAgent(state)
 
-    const result = interruptFromAgent(agent, 'test-id', {
-      name: 'confirm',
-      response: null,
-    })
+    const result = interruptFromAgent(
+      agent,
+      'test-id',
+      {
+        name: 'confirm',
+        response: null,
+      },
+      'tool'
+    )
 
     expect(result).toBeNull()
-    expect(state.interrupts['test-id']).toEqual({
+    expect(state.interrupts['test-id']).toMatchObject({
       id: 'test-id',
       name: 'confirm',
       response: null,
+      source: 'tool',
     })
   })
 })
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -60,6 +60,7 @@ import {
   ToolResultEvent,
   AgentResultEvent,
   ToolStreamUpdateEvent,
+  InterruptEvent,
   type ModelStopData,
 } from '../hooks/events.js'
 import { StructuredOutputTool, STRUCTURED_OUTPUT_TOOL_NAME } from '../tools/structured-output-tool.js'
@@ -1070,6 +1071,12 @@ export class Agent implements LocalAgent, InvokableAgent {
         return result
       }
       if (error instanceof InterruptError) {
+        // Fan out one event per interrupt. Each event exposes `interrupt.source` so
+        // consumers can filter by origin (tool callback vs hook callback) without
+        // subscribing to separate event types.
+        for (const interrupt of error.interrupts) {
+          yield new InterruptEvent({ agent: this, interrupt, invocationState })
+        }
         result = this._createInterruptResult(invocationState)
         return result
       }
@@ -1870,7 +1877,7 @@ export class Agent implements LocalAgent, InvokableAgent {
           agent: this,
           invocationState,
           interrupt: <T = JSONValue>(params: InterruptParams): T => {
-            return interruptFromAgent<T>(this, `tool:${toolUseBlock.toolUseId}:${params.name}`, params)
+            return interruptFromAgent<T>(this, `tool:${toolUseBlock.toolUseId}:${params.name}`, params, 'tool')
           },
         }
 
--- a/strands-ts/src/hooks/events.ts
+++ b/strands-ts/src/hooks/events.ts
@@ -4,7 +4,7 @@ import { type Tool, ToolStreamEvent } from '../tools/tool.js'
 import type { JSONValue } from '../types/json.js'
 import type { ModelStreamEvent } from '../models/streaming.js'
 import type { Model } from '../models/model.js'
-import { interruptFromAgent, type Interruptible } from '../interrupt.js'
+import { interruptFromAgent, type Interrupt, type Interruptible } from '../interrupt.js'
 import type { InterruptParams } from '../types/interrupt.js'
 
 /**
@@ -288,7 +288,12 @@ export class BeforeToolCallEvent extends HookableEvent implements Interruptible 
    * @returns The user's response when resuming from an interrupt
    */
   interrupt<T = JSONValue>(params: InterruptParams): T {
-    return interruptFromAgent<T>(this.agent, `hook:beforeToolCall:${this.toolUse.toolUseId}:${params.name}`, params)
+    return interruptFromAgent<T>(
+      this.agent,
+      `hook:beforeToolCall:${this.toolUse.toolUseId}:${params.name}`,
+      params,
+      'hook'
+    )
   }
 
   /**
@@ -695,6 +700,30 @@ export class AgentResultEvent extends HookableEvent {
 }
 
 /**
+ * Event emitted when an interrupt is raised during agent execution. The `interrupt.source`
+ * field discriminates between tool-callback and hook-callback origins. One event fires
+ * per unanswered interrupt at the moment the agent stops to wait for responses.
+ */
+export class InterruptEvent extends HookableEvent {
+  readonly type = 'interruptEvent' as const
+  readonly agent: LocalAgent
+  readonly interrupt: Interrupt
+  readonly invocationState: InvocationState
+
+  constructor(data: { agent: LocalAgent; interrupt: Interrupt; invocationState: InvocationState }) {
+    super()
+    this.agent = data.agent
+    this.interrupt = data.interrupt
+    this.invocationState = data.invocationState
+  }
+
+  /** Serializes for wire transport, excluding agent and invocationState. */
+  toJSON(): Pick<InterruptEvent, 'type'> & { interrupt: ReturnType<Interrupt['toJSON']> } {
+    return { type: this.type, interrupt: this.interrupt.toJSON() }
+  }
+}
+
+/**
  * Event triggered before executing tools.
  * Fired when the model returns tool use blocks that need to be executed.
  * Hook callbacks can set {@link cancel} to prevent all tools from executing.
@@ -728,7 +757,7 @@ export class BeforeToolsEvent extends HookableEvent implements Interruptible {
    * @returns The user's response when resuming from an interrupt
    */
   interrupt<T = JSONValue>(params: InterruptParams): T {
-    return interruptFromAgent<T>(this.agent, `hook:beforeTools:${params.name}`, params)
+    return interruptFromAgent<T>(this.agent, `hook:beforeTools:${params.name}`, params, 'hook')
   }
 
   /**
--- a/strands-ts/src/interrupt.ts
+++ b/strands-ts/src/interrupt.ts
@@ -48,7 +48,8 @@ export class Interrupt {
 
   /**
    * Where this interrupt was raised from — a tool callback or a hook callback.
-   * Optional to preserve backward compatibility with older serialized snapshots.
+   * Always populated on interrupts created by current code; only `undefined` when
+   * deserializing a snapshot produced by a pre-interrupt-source SDK version.
    */
   readonly source?: InterruptSource
 
@@ -442,10 +443,11 @@ export function interruptFromMultiAgentNode<T>(
       source,
     })
 
-  if (!existing) interrupts.push(interrupt)
-
-  if (interrupt.response !== undefined) {
-    return interrupt.response as T
+  if (!existing) {
+    interrupts.push(interrupt)
+    if (interrupt.response !== undefined) {
+      return interrupt.response as T
+    }
   }
 
   throw new InterruptError(interrupt)
--- a/strands-ts/src/interrupt.ts
+++ b/strands-ts/src/interrupt.ts
@@ -15,6 +15,14 @@ import type { LocalAgent } from './types/agent.js'
 import { Message, ToolResultBlock, type MessageData, type ToolResultBlockData } from './types/messages.js'
 
 /**
+ * Origin of an interrupt:
+ * - `'tool'` — raised by a tool callback via `ToolContext.interrupt()`.
+ * - `'hook'` — raised by an agent-level hook (e.g. `BeforeToolCallEvent.interrupt()`).
+ * - `'multiagent-hook'` — raised by a multi-agent hook (e.g. `BeforeNodeCallEvent.interrupt()`).
+ */
+export type InterruptSource = 'tool' | 'hook' | 'multiagent-hook'
+
+/**
  * Represents an interrupt that can pause agent execution for human-in-the-loop workflows.
  */
 export class Interrupt {
@@ -38,7 +46,13 @@ export class Interrupt {
    */
   response?: JSONValue
 
-  constructor(data: { id: string; name: string; reason?: JSONValue; response?: JSONValue }) {
+  /**
+   * Where this interrupt was raised from — a tool callback or a hook callback.
+   * Optional to preserve backward compatibility with older serialized snapshots.
+   */
+  readonly source?: InterruptSource
+
+  constructor(data: { id: string; name: string; reason?: JSONValue; response?: JSONValue; source?: InterruptSource }) {
     this.id = data.id
     this.name = data.name
     if (data.reason !== undefined) {
@@ -47,17 +61,21 @@ export class Interrupt {
     if (data.response !== undefined) {
       this.response = data.response
     }
+    if (data.source !== undefined) {
+      this.source = data.source
+    }
   }
 
   /**
    * Serializes the interrupt to a JSON-compatible object.
    */
-  toJSON(): { id: string; name: string; reason?: JSONValue; response?: JSONValue } {
+  toJSON(): { id: string; name: string; reason?: JSONValue; response?: JSONValue; source?: InterruptSource } {
     return {
       id: this.id,
       name: this.name,
       ...(this.reason !== undefined && { reason: this.reason }),
       ...(this.response !== undefined && { response: this.response }),
+      ...(this.source !== undefined && { source: this.source }),
     }
   }
 
@@ -67,7 +85,13 @@ export class Interrupt {
    * @param data - JSON data to deserialize
    * @returns Interrupt instance
    */
-  static fromJSON(data: { id: string; name: string; reason?: JSONValue; response?: JSONValue }): Interrupt {
+  static fromJSON(data: {
+    id: string
+    name: string
+    reason?: JSONValue
+    response?: JSONValue
+    source?: InterruptSource
+  }): Interrupt {
     return new Interrupt(data)
   }
 }
@@ -274,9 +298,16 @@ export class InterruptState implements InterruptStateData {
    * @param name - User-defined name for the interrupt
    * @param reason - Optional reason for the interrupt
    * @param response - Optional preemptive response to skip the interrupt
+   * @param source - Where the interrupt was raised from (tool or hook callback)
    * @returns The interrupt (may have a response if resuming or preemptive)
    */
-  getOrCreateInterrupt(id: string, name: string, reason?: JSONValue, response?: JSONValue): Interrupt {
+  getOrCreateInterrupt(
+    id: string,
+    name: string,
+    reason?: JSONValue,
+    response?: JSONValue,
+    source?: InterruptSource
+  ): Interrupt {
     const existing = this.interrupts[id]
     if (existing) {
       return existing
@@ -287,6 +318,7 @@ export class InterruptState implements InterruptStateData {
       name,
       ...(reason !== undefined && { reason }),
       ...(response !== undefined && { response }),
+      ...(source !== undefined && { source }),
     })
     this.interrupts[id] = interrupt
     return interrupt
@@ -349,18 +381,68 @@ export interface Interruptible {
  * @param agent - The agent whose interrupt state to access
  * @param interruptId - Unique identifier for this interrupt instance
  * @param params - Interrupt parameters including name and optional reason
+ * @param source - Where the interrupt was raised from (tool callback vs hook callback)
  * @returns The user's response when resuming from an interrupt
  * @throws InterruptError when no response is available (first invocation)
  *
  * @internal
  */
-export function interruptFromAgent<T>(agent: LocalAgent, interruptId: string, params: InterruptParams): T {
+export function interruptFromAgent<T>(
+  agent: LocalAgent,
+  interruptId: string,
+  params: InterruptParams,
+  source: InterruptSource
+): T {
   const interruptState = (agent as unknown as { _interruptState?: InterruptState })._interruptState
   if (!interruptState) {
     throw new Error('Interrupt state not available')
   }
 
-  const interrupt = interruptState.getOrCreateInterrupt(interruptId, params.name, params.reason, params.response)
+  const interrupt = interruptState.getOrCreateInterrupt(
+    interruptId,
+    params.name,
+    params.reason,
+    params.response,
+    source
+  )
+
+  if (interrupt.response !== undefined) {
+    return interrupt.response as T
+  }
+
+  throw new InterruptError(interrupt)
+}
+
+/**
+ * Interrupt-or-resume helper for multi-agent hooks where interrupts live on a per-node
+ * `Interrupt[]` list rather than on an agent's `InterruptState`. Mirrors the
+ * {@link interruptFromAgent} contract: returns the response if the interrupt already
+ * has one (resume path), otherwise records a new interrupt and throws `InterruptError`.
+ *
+ * @internal
+ */
+export function interruptFromMultiAgentNode<T>(
+  interrupts: Interrupt[],
+  interruptId: string,
+  params: InterruptParams,
+  source: InterruptSource
+): T {
+  const existing = interrupts.find((i) => i.id === interruptId)
+  if (existing?.response !== undefined) {
+    return existing.response as T
+  }
+
+  const interrupt =
+    existing ??
+    new Interrupt({
+      id: interruptId,
+      name: params.name,
+      ...(params.reason !== undefined && { reason: params.reason }),
+      ...(params.response !== undefined && { response: params.response }),
+      source,
+    })
+
+  if (!existing) interrupts.push(interrupt)
 
   if (interrupt.response !== undefined) {
     return interrupt.response as T
--- a/strands-ts/src/interrupt.ts
+++ b/strands-ts/src/interrupt.ts
@@ -47,11 +47,11 @@ export class Interrupt {
   response?: JSONValue
 
   /**
-   * Where this interrupt was raised from — a tool callback or a hook callback.
-   * Always populated on interrupts created by current code; only `undefined` when
-   * deserializing a snapshot produced by a pre-interrupt-source SDK version.
+   * Where this interrupt was raised from — a tool callback, an agent-level hook, or
+   * a multi-agent orchestrator hook. Always populated. When deserializing a snapshot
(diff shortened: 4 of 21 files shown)


strands-ts/src/multiagent/state.ts
```diff
@@ -307,5 +369,10 @@ export class MultiAgentState implements StateSerializable {
         this._nodes.set(id, nodeState)
       }
     }
+    if (data._pendingInput !== undefined) {
+      this._pendingInput = data._pendingInput as unknown as MultiAgentInput
```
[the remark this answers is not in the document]
notowen333: Addressed in incoming commit: `_pendingInput` load in `state.ts` now runs array entries through `contentBlockFromData` via a `rehydratePendingInput` helper, so restored `ContentBlock[]` values are real instances rather than raw data. Two new unit tests in `state.test.ts` pin the round-trip for `string` and `ContentBlock[]`.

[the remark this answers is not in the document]
notowen333: Same concern as the other `_pendingInput` thread — resolved. `rehydratePendingInput` in `state.ts` now maps array entries through `contentBlockFromData` on load, so `TextBlock` etc. come back as instances. Covered by a new unit test in `state.test.ts` that asserts `instanceof TextBlock` after round-trip.

strands-ts/src/multiagent/__tests__/state.test.ts
```diff
@@ -247,6 +250,44 @@ describe('NodeState', () => {
...
+        data: { messages: [], interrupts: { activated: true, interrupts: {} } },
+        appData: {},
+      }
+
+      const restored = new NodeState()
+      restored[loadStateFromJSONSymbol](original[stateToJSONSymbol]())
+
+      expect(restored.status).toBe(Status.INTERRUPTED)
+      expect(restored.interrupts).toHaveLength(1)
+      expect(restored.interrupts[0]).toMatchObject({ id: 'tool:1:confirm', name: 'confirm', reason: 'need it' })
```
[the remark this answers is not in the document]
notowen333: Addressed: swapped the one `toMatchObject` I added (the Interrupt round-trip assertion) for `toEqual(original.interrupts[0])` so the full object shape is checked.

strands-ts/src/multiagent/swarm.ts
```diff
@@ -244,40 +265,73 @@ export class Swarm implements MultiAgent {
...
+
+    let node: AgentNode
+    let handoff: HandoffResult | undefined
+    let nextInput: MultiAgentInput = input
+    if (interruptResponsesByNode) {
+      // Swarm is sequential, so at most one node is INTERRUPTED per run.
+      const entry = interruptResponsesByNode.entries().next().value
+      if (!entry) throw new Error(`swarm_id=<${this.id}> | no interrupt responses to route`)
+      const [nodeId, responses] = entry
+      node = this.nodes.get(nodeId)!
```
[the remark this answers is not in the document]
notowen333: Addressed: `this.nodes.get(nodeId)` and `state.node(nodeId)` in the swarm resume path now do explicit existence checks and throw descriptive errors (`topology changed between save and resume?`) instead of relying on non-null assertions.

strands-ts/src/multiagent/graph.ts
```diff
@@ -255,41 +275,83 @@ export class Graph implements MultiAgent {
...
+      contentInput = state._pendingInput as MultiAgentContentInput | undefined
+    } else {
+      contentInput = input as MultiAgentContentInput
+      state._pendingInput = contentInput
+    }
 
-    // Resume: if state was restored, find nodes that are ready but haven't completed otherwise start from source nodes
-    const targets = (await this._findResumeTargets(state)) ?? [...this._sources]
+    const targets = interruptResponsesByNode
+      ? [...interruptResponsesByNode.keys()].map((id) => this.nodes.get(id)!)
```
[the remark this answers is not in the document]
notowen333: Addressed: `this.nodes.get(id)` in the graph resume target construction now explicitly checks for undefined and throws `node_id=<...>, graph_id=<...> | resume response targets a node missing from the graph; topology changed between save and resume?`. Similar fix applied to `_resolveInputForScheduling`. A new unit test in `interrupts.test.ts` pins the error when the rebuilt graph has a different node id than the persisted state.

strands-ts/src/interrupt.ts
```diff
@@ -349,18 +381,68 @@ export interface Interruptible {
...
+      id: interruptId,
+      name: params.name,
+      ...(params.reason !== undefined && { reason: params.reason }),
+      ...(params.response !== undefined && { response: params.response }),
+      source,
+    })
+
+  if (!existing) interrupts.push(interrupt)
 
   if (interrupt.response !== undefined) {
```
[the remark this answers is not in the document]
notowen333: Addressed: collapsed the second check into the `if (!existing)` branch where it actually matters (a newly-constructed interrupt carrying a pre-set response). Single early-return path for the existing/answered case, single throw for the not-answered case.

strands-ts/src/interrupt.ts
```diff
@@ -38,7 +46,13 @@ export class Interrupt {
    */
   response?: JSONValue
 
-  constructor(data: { id: string; name: string; reason?: JSONValue; response?: JSONValue }) {
+  /**
+   * Where this interrupt was raised from — a tool callback or a hook callback.
+   * Optional to preserve backward compatibility with older serialized snapshots.
+   */
+  readonly source?: InterruptSource
```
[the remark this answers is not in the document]
notowen333: Addressed: TSDoc on `Interrupt.source` now states it is always populated on interrupts created by current code and is only `undefined` when deserializing a snapshot produced by a pre-interrupt-source SDK version.

strands-ts/src/interrupt.ts
```diff
@@ -38,7 +46,13 @@ export class Interrupt {
...
+
+  constructor(data: { id: string; name: string; reason?: JSONValue; response?: JSONValue; source?: InterruptSource }) {
```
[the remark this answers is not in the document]
notowen333: Addressed: TSDoc on `Interrupt.source` clarified — always populated on new interrupts, only undefined for pre-this-PR snapshots.

strands-ts/src/multiagent/multiagent.ts
```diff
@@ -2,19 +2,32 @@ import type { InvocationState, InvokeArgs } from '../types/agent.js'
...
- * and `InterruptResponseContent[]` from {@link InvokeArgs} since orchestrators route
- * content blocks between nodes. Graph interrupts will be handled separately.
+ * The non-resume subset of {@link MultiAgentInput}. Internal orchestrator helpers that
+ * participate in dependency resolution / handoff routing accept this narrower type so
+ * they don't need to re-check for {@link InterruptResponseContent} entries at each call.
  */
-export type MultiAgentInput = Exclude<
-  InvokeArgs,
-  Message[] | MessageData[] | InterruptResponseContent[] | InterruptResponseContentData[]
+export type MultiAgentContentInput = Exclude<
```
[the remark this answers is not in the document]
notowen333: Addressed: `MultiAgentContentInput` now carries an explicit `@internal` TSDoc tag. Still exported from the module file so `graph.ts` can import it, but the tag signals that external consumers should not depend on it.

strands-ts/src/multiagent/multiagent.ts
```diff
@@ -2,19 +2,32 @@ import type { InvocationState, InvokeArgs } from '../types/agent.js'
...
+  MultiAgentInput,
+  InterruptResponseContent[] | InterruptResponseContentData[]
 >
```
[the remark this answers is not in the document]
notowen333: Agreed — keeping it internal for now, marked `@internal` in TSDoc. If custom node implementations become a documented pattern, we can revisit the export surface.

strands-ts/src/multiagent/graph.ts
```diff
@@ -255,41 +275,83 @@ export class Graph implements MultiAgent {
...
-    // composed cancel signal. In-flight nodes return `stopReason: 'cancelled'`; the queue
-    // drains and the outer loop throws below when it sees the signal aborted.
-    const execController = Number.isFinite(this.config.timeout) ? new AbortController() : undefined
-    const execTimeoutHandle = execController ? setTimeout(() => execController.abort(), this.config.timeout) : undefined
+    // execController fires on both timeout and interrupt short-circuit; external
+    // cancellation stays on its own signal so the loop can distinguish the two causes.
+    const execController = new AbortController()
+    const execTimeoutHandle = Number.isFinite(this.config.timeout)
+      ? setTimeout(() => execController.abort(), this.config.timeout)
+      : undefined
```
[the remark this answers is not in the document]
notowen333: Addressed: expanded the comment on `execController` to name both abort sources (timeout vs interrupt short-circuit), call out the `!interrupted` discriminator explicitly, and warn future editors to keep it intact. The two sources and the guard are now self-documenting in the source.

strands-ts/src/multiagent/swarm.ts
```diff
@@ -244,40 +265,73 @@ export class Swarm implements MultiAgent {
...
+      state._pendingInput = input
+    }
+
+    let node: AgentNode
+    let handoff: HandoffResult | undefined
+    let nextInput: MultiAgentInput = input
+    if (interruptResponsesByNode) {
+      // Swarm is sequential, so at most one node is INTERRUPTED per run.
+      const entry = interruptResponsesByNode.entries().next().value
+      if (!entry) throw new Error(`swarm_id=<${this.id}> | no interrupt responses to route`)
```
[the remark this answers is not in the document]
notowen333: Addressed: added an explicit invariant check that throws if `interruptResponsesByNode.size > 1` — a future change that accidentally produces multiple interrupted nodes simultaneously will surface loudly rather than silently taking the first entry from the Map. Comment above the check explains the invariant.

strands-ts/src/multiagent/multiagent.ts
```diff
@@ -66,3 +92,114 @@ export interface MultiAgent {
...
+ */
+export function groupInterruptResponsesByNode(
+  responses: InterruptResponseContent[],
+  state: MultiAgentState
+): Map<string, InterruptResponseContent[]> {
+  const grouped = new Map<string, InterruptResponseContent[]>()
+  for (const response of responses) {
+    const id = response.interruptResponse.interruptId
+    let target: string | undefined
+    for (const [nodeId, nodeState] of state.nodes) {
```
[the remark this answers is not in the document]
notowen333: Accepting the trade-off as-is. For typical graph/swarm sizes (handful of nodes, handful of concurrent interrupts) the nested scan is imperceptible. Converting to a reverse index would add a few lines for no observable gain today. Happy to revisit if interrupt volume grows — filing a mental note rather than changing code preemptively.

strands-ts/src/multiagent/graph.ts
```diff
@@ -578,10 +685,38 @@ export class Graph implements MultiAgent {
...
+   * (dependency-merged). Falls back to an empty input with a warning if a custom
+   * SessionManager dropped `_pendingInput`.
+   */
+  private _resolveInputForScheduling(
+    node: Node,
+    routed: InterruptResponseContent[] | undefined,
+    contentInput: MultiAgentContentInput | undefined,
+    state: MultiAgentState
+  ): MultiAgentInput {
+    if (routed) {
```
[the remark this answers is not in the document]
notowen333: Deliberate choice to warn-and-degrade rather than throw here. The condition is: a custom `SessionManager` persisted state without `_pendingInput` (e.g., a partial-persistence implementation or a legacy snapshot format). Throwing on resume in that case would be worse than running with empty input — the node already has upstream results to work from via dependency merging, and a hard error blocks recovery entirely. Mirrors the existing pattern in `_findResumeNode` (swarm) which logs and falls back to `this.start` when the resume target is missing. Warning ensures the misconfiguration is logged; degradation keeps the run productive.

strands-ts/src/hooks/events.ts
```diff
@@ -680,6 +685,30 @@ export class AgentResultEvent extends HookableEvent {
   }
 }
 
+/**
+ * Event emitted when an interrupt is raised during agent execution. The `interrupt.source`
+ * field discriminates between tool-callback and hook-callback origins. One event fires
+ * per unanswered interrupt at the moment the agent stops to wait for responses.
+ */
+export class InterruptEvent extends HookableEvent {
```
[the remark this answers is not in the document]
notowen333: Confirming: this matches how every other `HookableEvent` behaves. `agent.invoke()` drains the generator and routes events through the hook registry; `agent.stream()` consumers that manage their own event loop are responsible for running hooks themselves if they have any registered. Not specific to `InterruptEvent`. The TSDoc doesn't call this out for `ModelMessageEvent`, `ToolResultEvent`, etc. either — consistent with the framework convention.

strands-ts/src/__tests__/interrupt.test.ts
zastrowm, line 1: Do we have an issue reference for the follow-up issue? (This PR does not solve cross-orchestrator interrupt resume when a nested orchestrator (a MultiAgentNode wrapping an inner Graph or Swarm) has no SessionManager)

strands-ts/src/hooks/events.ts
```diff
@@ -728,7 +757,7 @@ export class BeforeToolsEvent extends HookableEvent implements Interruptible {
    * @returns The user's response when resuming from an interrupt
    */
   interrupt<T = JSONValue>(params: InterruptParams): T {
-    return interruptFromAgent<T>(this.agent, `hook:beforeTools:${params.name}`, params)
+    return interruptFromAgent<T>(this.agent, `hook:beforeTools:${params.name}`, params, 'hook')
```
zastrowm, line 760: Is this a breaking change at all?

strands-ts/src/agent/agent.ts
```diff
@@ -1000,6 +1001,12 @@ export class Agent implements LocalAgent, InvokableAgent {
         return result
       }
       if (error instanceof InterruptError) {
+        // Fan out one event per interrupt. Each event exposes `interrupt.source` so
+        // consumers can filter by origin (tool callback vs hook callback) without
+        // subscribing to separate event types.
+        for (const interrupt of error.interrupts) {
+          yield new InterruptEvent({ agent: this, interrupt, invocationState })
```
zastrowm, line 1078: Can type ever be 'multiagent-hook' here?

strands-ts/src/multiagent/__tests__/state.test.ts
```diff
@@ -247,6 +250,44 @@ describe('NodeState', () => {
...
+      original.resumeSnapshot = {
+        scope: 'agent',
+        schemaVersion: '1.0',
+        createdAt: '2026-01-01T00:00:00Z',
+        data: { messages: [], interrupts: { activated: true, interrupts: {} } },
+        appData: {},
+      }
+
+      const restored = new NodeState()
+      restored[loadStateFromJSONSymbol](original[stateToJSONSymbol]())
```
zastrowm: Use `loadStateSerializable` and `serializeStateSerializable` instead; it's more readable

strands-ts/src/multiagent/__tests__/state.test.ts
```diff
@@ -247,6 +250,44 @@ describe('NodeState', () => {
...
+
+      expect(restored.status).toBe(Status.INTERRUPTED)
```
zastrowm: low nit: expect the entire object here and throughout

strands-ts/src/multiagent/graph.ts
```diff
@@ -255,41 +275,97 @@ export class Graph implements MultiAgent {
     })
 
     // SessionManager (or plugins) may restore state.results here via the hook
-    yield new BeforeMultiAgentInvocationEvent({ orchestrator: this, state, invocationState })
+    yield* this._emit(new BeforeMultiAgentInvocationEvent({ orchestrator: this, state, invocationState }))
```
zastrowm, line 278: Oh that's kind of clever

strands-ts/src/multiagent/graph.ts
```diff
@@ -255,41 +275,97 @@ export class Graph implements MultiAgent {
...
+
+    // Resume input bypasses dependency resolution (routed by interrupt id). On fresh
+    // runs, stash the input so resume can replay it to hook-gated nodes that never ran.
```
zastrowm, line 281: What's an example of this?:

 1. Node A starts, and interrupts
 2. When the graph resumes later, it needs the original input for hook-gated nodes?

strands-ts/src/multiagent/graph.ts
```diff
@@ -307,14 +383,26 @@ export class Graph implements MultiAgent {
 
           state.results.push(nodeResult)
 
+          if (interrupted) continue
+
+          // Short-circuit: first node to interrupt aborts in-flight siblings so they
+          // don't keep burning work past the point we're going to pause.
+          if (nodeResult.status === Status.INTERRUPTED) {
```
zastrowm, line 392: This matches Python behavior?

strands-ts/src/multiagent/graph.ts
```diff
@@ -307,14 +383,26 @@ export class Graph implements MultiAgent {
...
+
           const ready = await this._findReady(node, state, streams, targets)
           if (ready.length > 0) {
-            yield new MultiAgentHandoffEvent({
-              source: node.id,
-              targets: ready.map((n) => n.id),
-              state,
-              invocationState,
-            })
+            yield* this._emit(
```
zastrowm, line 399: Emit is clever, I like it

strands-ts/src/multiagent/graph.ts
```diff
@@ -339,58 +434,86 @@ export class Graph implements MultiAgent {
...
-    await queue.send({ type: 'event', node, event: beforeEvent })
+    try {
+      await this._hookRegistry.invokeCallbacks(beforeEvent)
+    } catch (error) {
+      if (error instanceof InterruptError) {
+        const result = recordHookInterrupt(node.id, nodeState)
+        yield beforeEvent
+        yield* this._emit(new NodeResultEvent({ nodeId: node.id, nodeType: node.type, state, result, invocationState }))
+        yield* this._emit(new AfterNodeCallEvent({ orchestrator: this, state, nodeId: node.id, invocationState }))
+        this._tracer.endNodeSpan(nodeSpan, { status: Status.INTERRUPTED, duration: result.duration })
```
zastrowm, line 479: Do we have tests verifying the spans?

strands-ts/src/multiagent/multiagent.ts
```diff
@@ -66,3 +94,114 @@ export interface MultiAgent {
...
+/**
+ * Detects whether a {@link MultiAgentInput} is a resume from an interrupted run and
+ * normalizes its entries to {@link InterruptResponseContent} instances.
+ *
+ * Returns `undefined` for fresh input (string / content blocks / empty array).
+ */
+export function extractResumeResponses(input: MultiAgentInput): InterruptResponseContent[] | undefined {
+  if (!Array.isArray(input) || input.length === 0) return undefined
+  if (!isInterruptResponseContent(input[0])) return undefined
+  return (input as readonly (InterruptResponseContent | InterruptResponseContentData)[]).map((entry) =>
```
zastrowm: nit; this cast should be on the return value, not the initial array

strands-ts/src/multiagent/multiagent.ts
```diff
@@ -66,3 +94,114 @@ export interface MultiAgent {
...
+    entry instanceof InterruptResponseContent ? entry : InterruptResponseContent.fromJSON(entry)
```
zastrowm: I think in the single agent case we throw when we have non-interrupt content; if so, we should do that here

strands-ts/src/multiagent/multiagent.ts
```diff
@@ -66,3 +94,114 @@ export interface MultiAgent {
...
+/**
+ * Removes a stale INTERRUPTED result for the given node from both per-node history
+ * and the orchestrator-level aggregate so a fresh result (from resume or cancel)
+ * replaces it cleanly. No-op if the node isn't in an INTERRUPTED state.
+ */
+export function dropStaleInterruptedResult(nodeId: string, nodeState: NodeState, state: MultiAgentState): void {
+  if (nodeState.status !== Status.INTERRUPTED) return
+  if (nodeState.results[nodeState.results.length - 1]?.status === Status.INTERRUPTED) {
+    nodeState.results.pop()
+  }
```
zastrowm, line 166: This does the same as the next lines, right?  If so, I'd remove these three lines to simplify the code

```suggestion
```

strands-ts/src/multiagent/nodes.ts
```diff
@@ -213,11 +230,16 @@ export class AgentNode extends Node {
...
     const invocationState: InvocationState = options?.invocationState ?? {}
 
-    // Only Agent instances support snapshot/restore for state isolation
-    const snapshot =
-      this._agent instanceof Agent
-        ? takeSnapshot(this._agent, { include: ['messages', 'state', 'modelState'] })
-        : undefined
+    // Only Agent instances support snapshot/restore for state isolation.
+    const isAgent = this._agent instanceof Agent
+    const preRunSnapshot = isAgent ? takeSnapshot(this._agent as Agent, { preset: 'session' }) : undefined
```
zastrowm: You should now be able to do this._agent.takeSnapshot() I think?

Same for load below

strands-ts/src/multiagent/nodes.ts
```diff
@@ -232,23 +254,34 @@ export class AgentNode extends Node {
...
+      const interrupted =
+        agentResult.stopReason === 'interrupt' && agentResult.interrupts && agentResult.interrupts.length > 0
+
+      // Capture post-interrupt state for the next resume cycle. Only Agent instances
+      // are snapshottable.
+      if (interrupted && isAgent && nodeState) {
+        nodeState.resumeSnapshot = takeSnapshot(this._agent as Agent, { preset: 'session' })
+      }
+
       return {
```
zastrowm: Can `omitUndefined` be used here to simplify?

strands-ts/src/multiagent/state.ts
```diff
@@ -122,12 +139,22 @@ export class NodeState implements StateSerializable {
...
   readonly results: NodeResult[]
+  /** Unanswered interrupts raised during this node's most recent run. Populated when `status === 'INTERRUPTED'`. */
+  interrupts: Interrupt[]
+  /**
+   * Snapshot of the node's underlying runnable (Agent or nested orchestrator) captured
+   * when the node returned INTERRUPTED. Loaded back into the runnable on resume so it
+   * can pick up mid-execution without losing its interrupt bookkeeping. Cleared when
+   * the node completes.
+   */
+  resumeSnapshot?: Snapshot
```
zastrowm: nit: interruptedSnapshot or something to make it clear that it's interrupt specific?

strands-ts/src/multiagent/swarm.ts
```diff
@@ -244,40 +265,92 @@ export class Swarm implements MultiAgent {
     })
 
     // SessionManager (or plugins) may restore state.results here via the hook
-    yield new BeforeMultiAgentInvocationEvent({ orchestrator: this, state, invocationState })
+    yield* this._emit(new BeforeMultiAgentInvocationEvent({ orchestrator: this, state, invocationState }))
```
zastrowm, line 268: I'm surprised Graph & Swarm can't share more here

strands-ts/src/interrupt.ts
```diff
@@ -38,7 +46,14 @@ export class Interrupt {
    */
   response?: JSONValue
 
-  constructor(data: { id: string; name: string; reason?: JSONValue; response?: JSONValue }) {
+  /**
+   * Where this interrupt was raised from — a tool callback or a hook callback.
+   * Always populated on interrupts created by current code; only `undefined` when
+   * deserializing a snapshot produced by a pre-interrupt-source SDK version.
+   */
+  readonly source?: InterruptSource
```
zastrowm: For 2.0, would we want to make this required?  If so, cut an issue

strands-ts/src/__tests__/interrupt.test.ts
notowen333: https://github.com/strands-agents/sdk-typescript/issues/1050

strands-ts/src/hooks/events.ts
notowen333: `interruptFromAgent` is marked as `@internal` so only if someone did a deep import on the dist which we we are not making any guarentees about I believe

strands-ts/src/agent/agent.ts
notowen333: No, only 'tool' or 'hook' reach this emission site. 'multiagent-hook' interrupts live on NodeState.interrupts and never flow through agent._stream's catch block."

strands-ts/src/multiagent/multiagent.ts
notowen333: Different arrays: `nodeState.results` is per-node history, `state.results` is the orchestrator-level aggregate. Both need the stale INTERRUPTED result dropped — `NodeState.content` reads the last entry of `nodeState.results` and is consumed by `_resolveContent` (terminus nodes) and dependency merging, so leaving a stale entry there would return empty content to downstream nodes.

strands-ts/src/multiagent/multiagent.ts
notowen333: Moved the cast to the return value.

strands-ts/src/multiagent/state.ts
notowen333: Renamed to `interruptedSnapshot` everywhere (field, serialization key, tests). The new name makes the interrupt-specific capture timing explicit.

strands-ts/src/multiagent/__tests__/state.test.ts
notowen333: Switched the tests I added (interrupt round-trip + `_pendingInput` round-trips) to use `loadStateSerializable` / `serializeStateSerializable`. Pre-existing tests in the file still use the symbol form for now — happy to sweep those in a follow-up if desired.

strands-ts/src/multiagent/__tests__/state.test.ts
notowen333: Tightened to full-object equality for the tests I added (`expect(restored).toEqual(original)`). Same caveat as the helper sweep — pre-existing tests retain their style.

strands-ts/src/multiagent/multiagent.ts
notowen333: Done — `extractResumeResponses` now throws `TypeError` if any entry is not an interrupt response, matching `Agent._extractInterruptResponses`. New unit test pins the throw.

strands-ts/src/multiagent/nodes.ts
notowen333: Done — rebased onto main and switched to `this._agent.takeSnapshot()` / `this._agent.loadSnapshot()` at all three sites in `AgentNode.handle`.

strands-ts/src/multiagent/nodes.ts
notowen333: Done — applied `omitUndefined` to both `AgentNode.handle` and `MultiAgentNode.handle` return objects. Reads cleaner without the spread-conditional pattern.

strands-ts/src/multiagent/graph.ts
notowen333: Span lifecycle for COMPLETED, FAILED, and CANCELLED node terminations was already covered. Added an INTERRUPTED test in `graph.tracer.test.ts` to pin the hook-interrupt path that ends `nodeSpan` from `_runBeforeNodeCall`.

strands-ts/src/multiagent/graph.ts
notowen333: Yep exactly what you said. We don't write the outer invocation input onto the node that didn't run. Without this property we can't execute the original input and we'd only have the follow-up after the interrupt.

1. Graph has a source node A with a `BeforeNodeCallEvent` hook that calls `event.interrupt(...)`.
2. User calls `graph.invoke('original task')`.
3. Hook fires before A runs, throws `InterruptError`, graph short-circuits with status INTERRUPTED. A never executed.
4. User collects the interrupt, returns later with `graph.invoke([response])`.
5. Hook re-fires for A, reads the stored response, returns it (no throw).
6. A is now ready to run — but the resume call only carried `[response]`, not `'original task'`. A's snapshot is empty (it never ran). No upstream results to merge.

strands-ts/src/multiagent/graph.ts
notowen333: It does not... I'm not sure what is the ideal behavior here. Short circuiting can avoid unnecessary work. Letting the other nodes work allows independent work to continue. Maybe we should bias for parity here?

notowen333: Synced with @pgrayy and we'll match python. We WON'T short-circuit

notowen333: We can support short-circuiting with configuration later on

strands-ts/src/interrupt.ts
notowen333: https://github.com/strands-agents/sdk-typescript/issues/1054

strands-ts/src/multiagent/graph.ts
zastrowm: As a follow-up, consider expanding the code comment a bit more to include this example

strands-ts/src/multiagent/graph.ts
zastrowm: +1 I think not short circuiting by default makes more sense and matches how tool interrupts work

strands-ts/src/interrupt.ts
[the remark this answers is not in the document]
zastrowm: Is deserialization the only case in which it's not populated?  If so, could we populate it with `agent`  (previously the only valid interrupt type) and ditch the optionality?

zastrowm (dismissed):

zastrowm (approved):

PR #1055, opened 2026-05-13
fix: warn on legacy Bedrock cache options

asim48-ctrl:
## Summary
This adds a small runtime warning when `BedrockModel` receives the stale `cachePrompt` or `cacheTools` fields documented on the older indexed API page. Those fields are not honored by the current provider, so the warning points users to `cacheConfig` instead of silently leaving Bedrock prompt caching disabled.

Fixes #1027.

## Testing
- npm ci --ignore-scripts
- npm run test -w strands-ts -- src/models/__tests__/bedrock.test.ts -t "legacy cache"
- npx prettier --check strands-ts/src/models/bedrock.ts strands-ts/src/models/__tests__/bedrock.test.ts
- git diff --check

### DIFF
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -102,6 +102,11 @@ const BEDROCK_CONTEXT_WINDOW_OVERFLOW_MESSAGES = [
  */
 const SKIP_COUNT_TOKENS_MODELS = new Set<string>()
 
+type LegacyBedrockCacheOptions = {
+  cachePrompt?: unknown
+  cacheTools?: unknown
+}
+
 /**
  * Mapping of Bedrock stop reasons to SDK stop reasons.
  */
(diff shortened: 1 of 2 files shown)


poshinchen (approved):

strands-ts/src/models/bedrock.ts
```diff
@@ -388,6 +393,14 @@ export class BedrockModel extends Model<BedrockModelConfig> {
   constructor(options?: BedrockModelOptions) {
     super()
 
+    const legacyCacheOptions = options as (BedrockModelOptions & LegacyBedrockCacheOptions) | undefined
+    if (legacyCacheOptions?.cachePrompt !== undefined || legacyCacheOptions?.cacheTools !== undefined) {
+      warnOnce(
+        logger,
+        'unsupported_fields=<cachePrompt,cacheTools> | cachePrompt and cacheTools are not supported by BedrockModel, use cacheConfig instead'
```
notowen333, line 400:
```
  1. (Functional gap) updateConfig is not covered. The same legacy fields can
  land via updateConfig({ cachePrompt: 'default' } as any) (bedrock.ts:513) —
  that path silently drops them just like the constructor used to. If the goal
  is "users shouldn't believe caching is enabled when it isn't," updateConfig
  should warn too. Minor refactor: extract a warnIfLegacyCacheOptions(options)
  helper and call it in both places.

  2. (Test placement) The new test lives in the wrong describe block. Per the
  diff context (@@ -448,6 +448,18 @@), the it('warns when unsupported legacy
  cache options...') is inserted at the tail of describe('getConfig')
  (bedrock.test.ts:437), not describe('constructor') (line 183), where the
  analogous it('warns when modelId is not explicitly set') test already lives
  (line 190). Move it next to the other constructor-warning tests for
  discoverability.

  3. (Message accuracy) Warning text claims both fields when only one may be
  present. If a caller passes only cachePrompt, the warning still says
  unsupported_fields=<cachePrompt,cacheTools>. Either:
  - List only the keys actually present (more accurate, costs warnOnce dedup
  since the key set varies), or
  - Keep the static message but drop the redundancy (the structured prefix and
  prose say the same thing twice).
```

notowen333 (changes requested):

Issue #904, opened 2026-04-24
[BUG] SDK consistency papercuts (minor differences vs Python SDK)


opieter-aws:
### Steps to Reproduce

Collection of minor consistency differences found during cross-SDK scan. None are critical individually, but together they create friction for users working with both SDKs.

**No null check on sub-agent result**
`agent-as-tool.ts:178-193`: No undefined check on `result` after generator completes. If generator returns undefined, accessing `result.structuredOutput` throws TypeError. Python explicitly checks `if result is None`.

**AgentResult.toString() missing citations, structured output, interrupts**
TypeScript only includes text and reasoning blocks. Python includes interrupts (stringified list), structured output (JSON), and citationsContent blocks. Affects multi-agent scenarios where parent agent receives sub-agent result as text.

**Snapshot resets systemPrompt, Python does not**
TypeScript's `takeSnapshot` with `preset: 'session'` captures and restores `systemPrompt`. Python only snapshots `messages` and `state`. Question: is this intentional?

**No hot-reload overwrite for duplicate tool registration**
TypeScript always throws on duplicate tool name. Python allows overwrite when `supports_hot_reload=True`.

**No normalized name conflict check (hyphen vs underscore)**
Python treats `tool-name` and `tool_name` as conflicts. TypeScript only checks exact name match.

**MessageAddedEvent documentation says it doesn't fire for initial input, but code does**
`events.ts:155` docs say it doesn't fire for initial user input. `agent.ts:651` fires it. Documentation bug.

**CitationLocation search result key differs**
Python uses `"searchResultLocation"` as outer key. TypeScript uses `"searchResult"` as type discriminator. User code checking citations needs different keys per SDK.

**cancel() no-ops when idle**
Python always sets cancel signal (next invocation exits early). TypeScript no-ops when not invoking.

**Structured output forced mode sends all tools**
Python sends only the structured output tool spec. TypeScript sends all registered tools but constrains via `toolChoice`. Minor token waste.

### Expected Behavior

Consistent behavior across SDKs for these edge cases.

### Actual Behavior

See individual items above.

### Additional context 

This was a finding from an agentic cross-SDK consistency scan between Python and TypeScript SDKs #789. The finding has been triaged, but verify that this needs an actual fix.

dpruessner:
>MessageAddedEvent documentation says it doesn't fire for initial input, but code does
events.ts:155 docs say it doesn't fire for initial user input. agent.ts:651 fires it. Documentation bug.

[...]

PR #1066, opened 2026-05-13
feat: added bedrock-mantle support

poshinchen:
## Summary
Mirrors [feat: enable openai provider use aws profile](https://github.com/strands-agents/sdk-python/pull/2230)

Amazon Bedrock's OpenAI-compatible endpoint (Mantle) at `https://bedrock-mantle.<region>.api.aws/v1` serves the Chat Completions and Responses APIs plus Responses-only features like stateful conversations and reasoning controls, capabilities the native Converse API doesn't expose.

Reaching Mantle from `OpenAIModel` today means hand-rolling the plumbing: mint a bearer token via SigV4 (or an AWS helper), build a custom `fetch`/`httpx` pipeline, and pass a bespoke `baseURL` and `apiKey` through `clientConfig`.

This PR adds a first-class `bedrockMantleConfig` option to `OpenAIModel` so the Bedrock routing stops leaking into application code.

This is the last paper-cut of https://github.com/strands-agents/sdk-typescript/issues/904

## Public API Changes

A new optional field on `OpenAIModelOptions` routes requests through Mantle.
A fresh bearer token is minted per request via `@aws/bedrock-token-generator` (new optional peer dependency).

```typescript
import { OpenAIModel } from '@strands-agents/sdk/models/openai'

// Chat Completions via Mantle
new OpenAIModel({
  api: 'chat',
  modelId: 'openai.gpt-oss-120b',
  bedrockMantleConfig: { region: 'us-east-1' },
})

// Responses API with stateful conversations and reasoning
new OpenAIModel({
  modelId: 'openai.gpt-oss-120b',
  stateful: true,
  params: { reasoning: { effort: 'medium' } },
  bedrockMantleConfig: { region: 'us-east-1' },
})
```
### Region
Region resolution prefers explicit config, then falls back to `AWS_REGION` / `AWS_DEFAULT_REGION`. Credentials come from the standard AWS credential chain unless an explicit `credentials` provider is passed in the config. The pathway cannot be combined with a pre-built `client` or a top-level `apiKey`, and `clientConfig` must not set `baseURL` or `apiKey` (those are derived).

Unlike the Python port, no per-call client reconstruction is needed: the OpenAI Node SDK's async `apiKey` setter is invoked before each request, so tokens are minted per request by the SDK itself.

## Use Cases

- Talk to OpenAI-family models from a Bedrock account using AWS IAM rather   than OpenAI-issued API keys.
- Use Responses-only features (server-side stateful conversations, reasoning effort) without giving up Bedrock's billing, networking, and audit posture.

## Testing

- [x] I ran `npm run check`

Also manually verified end-to-end against the live Mantle endpoint in `us-east-1` (Chat Completions, Responses stateless, Responses stateful multi-turn, Responses with `reasoning.effort` multi-turn).

### DIFF
--- a/strands-ts/src/models/openai/mantle.ts
+++ b/strands-ts/src/models/openai/mantle.ts
@@ -1,0 +1,142 @@
+/**
+ * Internal helpers for routing an {@link OpenAIModel} through Amazon Bedrock's
+ * OpenAI-compatible "Mantle" endpoint.
+ *
+ * Converts a {@link BedrockMantleConfig} into the `baseURL` and `apiKey` the
+ * OpenAI SDK consumes. Tokens are minted on demand via
+ * `@aws/bedrock-token-generator` so long-running agents survive the bearer
+ * token's maximum lifetime.
+ *
+ * `@aws/bedrock-token-generator` is declared as an optional peer dependency, so
+ * the import is lazy: it happens the first time the OpenAI client's async
+ * `apiKey` setter is invoked.
+ */
+
+import type { AwsCredentialIdentity, AwsCredentialIdentityProvider } from '@smithy/types'
+
+const MANTLE_DOCS_URL = 'https://docs.aws.amazon.com/bedrock/latest/userguide/inference-openai.html'
+
+/**
+ * Async function that returns a freshly minted Bedrock Mantle bearer token.
+ * Matches the shape returned by `@aws/bedrock-token-generator`'s
+ * `getTokenProvider`.
+ *
+ * @internal
+ */
+export type TokenProvider = () => Promise<string>
+
+/**
+ * Config for routing an OpenAI-compatible client through Amazon Bedrock's
+ * Mantle endpoint.
+ *
+ * When supplied to `OpenAIModel`, this config derives the OpenAI client's
+ * `baseURL` and `apiKey`. It cannot be combined with a pre-built `client`,
+ * a top-level `apiKey`, or `clientConfig.baseURL` / `clientConfig.apiKey`,
+ * since those are derived from this config.
+ */
+export interface BedrockMantleConfig {
+  /**
+   * AWS region hosting the Bedrock Mantle endpoint. If omitted, resolved from
+   * the `AWS_REGION` or `AWS_DEFAULT_REGION` environment variable. An error is
+   * thrown if none resolve.
+   */
+  region?: string
+
+  /**
+   * AWS credentials forwarded to the bearer token generator. Accepts either a
+   * static credential identity or a credential provider function (e.g. the
+   * result of `fromNodeProviderChain()` from `@aws-sdk/credential-providers`).
+   * When omitted, the token generator resolves credentials from the standard
+   * AWS credential chain.
+   */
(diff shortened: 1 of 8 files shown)


strands-ts/src/models/openai/bedrock-mantle.ts
```diff
@@ -0,0 +1,129 @@
...
+  )
+}
+
+/**
+ * Builds the Mantle base URL for a region.
+ *
+ * @internal
+ */
+export function bedrockMantleBaseUrl(region: string): string {
+  return MANTLE_BASE_URL_TEMPLATE.replace('{region}', region)
```
notowen333: clearer to just define the template in the same code block instead of up top IMO

strands-ts/src/models/openai/bedrock-mantle.ts
```diff
@@ -0,0 +1,129 @@
...
+ * Builds an async `apiKey` setter (matching the OpenAI SDK's `ApiKeySetter`
+ * signature) that mints a fresh bearer token on every request.
+ *
+ * The `@aws/bedrock-token-generator` package is loaded lazily on first use so
+ * applications that never touch the Mantle pathway don't need it installed.
+ *
+ * @internal
+ */
+export function createMantleApiKeySetter(config: BedrockMantleConfig, region: string): () => Promise<string> {
+  let provideTokenPromise: Promise<() => Promise<string>> | null = null
```
notowen333: This is functionally sound, but hard to read because `provideTokenPromise` is
a cached promise for the token provider function, not for a token. Consider
introducing a `TokenProvider` type alias and renaming this to
`tokenProviderPromise` so the lazy initialization behavior is clearer.

strands-ts/src/models/openai/model.ts
```diff
@@ -258,3 +265,34 @@ export class OpenAIModel extends Model<OpenAIModelConfig> {
...
 }
+
+function buildMantleClient(
+  bedrockMantleConfig: NonNullable<OpenAIModelOptions['bedrockMantleConfig']>,
+  apiKey: OpenAIModelOptions['apiKey'],
+  clientConfig: OpenAIModelOptions['clientConfig']
+): OpenAI {
+  if (apiKey !== undefined) {
+    throw new Error(
+      "'apiKey' cannot be combined with 'bedrockMantleConfig'; the API key is derived from the Mantle config automatically."
```
notowen333, line 276: is this also on the public doc string?

strands-ts/src/models/openai/mantle.ts
```diff
@@ -0,0 +1,129 @@
+/**
```
notowen333, line 1: Checking in that we're sure that `/openai` is the right directory for this code. 

Is there any openAI specific code here?

I could also see us placing it in `/bedrock` (which we'd need to create a folder for)

poshinchen: moved it to `bedrock/`

strands-ts/src/models/openai/model.ts
poshinchen: added

notowen333 (approved):

PR #1067, opened 2026-05-13
fix: cap Meter invocation history to prevent unbounded memory growth

pgrayy:
## Description

`Meter._agentInvocations` grows by one entry on every `agent.invoke()` call with no eviction. For long-lived Agent instances serving many requests (stateless API servers, batch processors), this causes silent, unbounded heap growth over time.

This PR caps the retained invocation history at 100 entries. When a new invocation begins and the array is at capacity, the oldest entry is shifted out. Lifetime-accumulated scalars (`cycleCount`, `totalDuration`, `accumulatedUsage`, `toolMetrics`) remain accurate regardless of eviction since they are tracked independently.

Users who need full invocation history can collect `result.metrics.latestAgentInvocation` from each `AgentResult` and store it themselves.

### Design rationale

I considered exposing a `MeterConfig` (with `maxInvocationHistory`) on `AgentConfig`, giving users explicit control over the cap. I opted for a hardcoded limit instead because:

- The unbounded growth is a bug, not a feature. No user intentionally opts into a memory leak. A reasonable default fixes the problem without requiring awareness of it.
- A `MeterConfig` type with a single field adds API surface for a knob most users will never touch. If configurable history retention becomes a real need, it can be added later as a clear feature request.
- The Meter is private on the Agent. Users access per-invocation metrics through `AgentResult`, so the SDK does not need to be a metrics database.
- I suspect most users who care about historical invocation data would prefer looking at it through OTEL exports on a dashboard rather than inspecting the in-process array. The same data (cycle durations, token usage, latency) is already emitted via OTEL counters and histograms as it happens.

The cap of 100 is a reasonable default for conversational agents while keeping memory negligible for server workloads.

## Related Issues

Closes #785

## Type of Change

Bug fix

## Testing

- [x] I ran `npm run check`
- All 2589 unit tests pass
- New test verifies eviction at the 100-entry cap

## Checklist

----

### DIFF
--- a/strands-ts/src/telemetry/__tests__/meter.test.ts
+++ b/strands-ts/src/telemetry/__tests__/meter.test.ts
@@ -144,16 +144,16 @@ describe('Meter', () => {
     })
 
     it('evicts oldest entries when exceeding max history cap', () => {
-      for (let i = 0; i < 110; i++) {
+      for (let i = 0; i < 60; i++) {
         meter.startNewInvocation()
         meter.startCycle()
       }
 
       const invocations = meter.metrics.agentInvocations
-      expect(invocations).toHaveLength(100)
+      expect(invocations).toHaveLength(50)
       // First retained entry is the 11th invocation (oldest 10 were evicted)
       expect(invocations[0]!.cycles[0]!.cycleId).toBe('cycle-11')
-      expect(invocations[99]!.cycles[0]!.cycleId).toBe('cycle-110')
+      expect(invocations[49]!.cycles[0]!.cycleId).toBe('cycle-60')
     })
   })
 
--- a/strands-ts/src/telemetry/__tests__/meter.test.ts
+++ b/strands-ts/src/telemetry/__tests__/meter.test.ts
@@ -142,6 +142,14 @@ describe('Meter', () => {
       expect(snapshot.agentInvocations).toHaveLength(2)
       expect(snapshot.latestAgentInvocation).toBe(snapshot.agentInvocations[1])
     })
+
+    it('evicts oldest entries when exceeding max history cap', () => {
+      for (let i = 0; i < 110; i++) {
+        meter.startNewInvocation()
+      }
+
+      expect(meter.metrics.agentInvocations).toHaveLength(100)
+    })
   })
 
   describe('startCycle', () => {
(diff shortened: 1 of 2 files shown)


strands-ts/src/telemetry/meter.ts
```diff
@@ -274,6 +274,14 @@ export class AgentMetrics implements JSONSerializable<AgentMetricsData> {
   }
 }
 
+/**
+ * Maximum number of invocation history entries retained by the Meter.
+ * Prevents unbounded memory growth on long-lived Agent instances.
+ * Users who need full history can collect per-invocation metrics
+ * from successive AgentResult objects.
+ */
+const MAX_INVOCATION_HISTORY = 100
```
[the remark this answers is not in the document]
pgrayy: This was intentional as I am viewing this as a bug. Also, not certain how useful agentInvocations is to users and so don't want to add unnecessary configs if we don't have to.

poshinchen: We could let user configure it, or we could lower it down to 20 / 50? 100 seems large to me.

poshinchen (dismissed):

strands-ts/src/telemetry/__tests__/meter.test.ts
zastrowm, line 1: > No user intentionally opts into a memory leak. A reasonable default fixes the problem without requiring awareness of it.

To clarify - we're talking about a memory bloat not leak, right? Nothing is leaked if you throw the agent is thrown away?

> I suspect most users who care about historical invocation data would prefer looking at it through OTEL exports on a dashboard rather than inspecting the in-process array. The same data (cycle durations, token usage, latency) is already emitted via OTEL counters and histograms as it happens.

Have we validated this?  The fact that we have invocations at all implies that there was a usage/need of it, so I want to ensure we're understanding the use case.

pgrayy: Right more specifically bloat. Regarding validation, this is something I want to confirm as well and so not quick to merge here.

strands-ts/src/telemetry/meter.ts
lizradway: lowering to 50

poshinchen (approved):
LGTM

PR #1059, opened 2026-05-13
feat(mcp): support url + auth fields on McpClientConfig

gautamsirdeshmukh:
## Description

`McpClient` now supports declarative connection config — `url`, `auth`, and `headers` fields — enabling authenticated MCP connections without manually setting transports and providers.
- When `url` is provided, a StreamableHTTP transport is constructed automatically — no manual transport setup required. 
- When `auth` is provided, a ClientCredentialsProvider handles OAuth token acquisition and refresh automatically via OIDC discovery. A raw `authProvider` field supports custom OAuth flows beyond client credentials (i.e. pre-acquired tokens, etc). 
- The `headers` field allows custom request headers without dropping back to manual transport construction.

```typescript
// Before: manual transport + provider wiring
const provider = new ClientCredentialsProvider({ clientId: '...', clientSecret: '...' })
const transport = new StreamableHTTPClientTransport(url, { authProvider: provider })
new McpClient({ transport })
```

```typescript
// After: OAuth with client credentials
new McpClient({ url: 'https://mcp.example.com', auth: { clientId: '...', clientSecret: '...' } })

// After: custom headers (e.g. static API keys)
new McpClient({ url: 'https://mcp.example.com', headers: { 'X-Api-Key': '...' } })

// After: custom auth provider (advanced flows)
new McpClient({ url: 'https://mcp.example.com', authProvider: myOAuthProvider })
```

## Related Issues

#850 

## Documentation PR

Will have a sweeping docs PR for the MCP enhancements once complete

## Type of Change

New feature (enhancement)

## Testing

How have you tested the change?

- [x] I ran `npm run check`

## Checklist
- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [ ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1070, opened 2026-05-13
feat(mcp): load MCP servers from JSON

gautamsirdeshmukh:
## Description

Adds `McpClient.loadServers()` — a static factory that parses an MCP servers config (file path or inline object) and returns ready-to-use `McpClient[]` without any manual transport construction.

**Before** (original manual way):
```ts
import { Agent, McpClient } from '@strands-agents/sdk'
import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js'
import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/streamableHttp.js'

const agent = new Agent({
  tools: [
    new McpClient({
      transport: new StdioClientTransport({
        command: 'uvx',
        args: ['awslabs.aws-documentation-mcp-server@latest'],
        env: { ...process.env, AWS_PROFILE: process.env.AWS_PROFILE! },
      }),
    }),
    new McpClient({
      transport: new StreamableHTTPClientTransport(
        new URL('https://api.githubcopilot.com/mcp/'),
        { requestInit: { headers: { Authorization: `Bearer ${process.env.GITHUB_TOKEN}` } } }
      ),
    }),
  ],
})
```

**After:**
```ts
import { Agent, McpClient } from '@strands-agents/sdk'

// Using template literals (standard TypeScript):
const agent = new Agent({
  tools: await McpClient.loadServers({
    'aws-docs': {
      command: 'uvx',
      args: ['awslabs.aws-documentation-mcp-server@latest'],
      env: { AWS_PROFILE: `${process.env.AWS_PROFILE}` },
    },
    'github': {
      url: 'https://api.githubcopilot.com/mcp/',
      headers: { Authorization: `Bearer ${process.env.GITHUB_TOKEN}` },
    },
  }),
})
```

```ts
// Using ${VAR} interpolation (same syntax works in JSON config files):
const agent = new Agent({
  tools: await McpClient.loadServers({
    'aws-docs': {
      command: 'uvx',
      args: ['awslabs.aws-documentation-mcp-server@latest'],
      env: { AWS_PROFILE: '${AWS_PROFILE}' },
    },
    'github': {
      url: 'https://api.githubcopilot.com/mcp/',
      headers: { Authorization: 'Bearer ${GITHUB_TOKEN}' },
    },
  }),
})
```

From a file:
```ts
const agent = new Agent({
  tools: await McpClient.loadServers('~/mcp-config.json'),
})
```

Mixed (load from file + manual):
```ts
const agent = new Agent({
  tools: [
    ...(await McpClient.loadServers('~/mcp-config.json')),
    new McpClient({
      url: 'https://my-custom-server.com/mcp',
      headers: { Authorization: `Bearer ${process.env.MY_TOKEN}` },
    }),
  ],
})
```

### Transport auto-detection

When `transport` is not explicitly set, it is inferred from the config fields:
- `command` present → `stdio`
- `url` present → `streamable-http`
- Neither → error

You can override this by setting `transport` explicitly (e.g. `transport: 'sse'` for a legacy server with a `url`).

### Additional changes

- Renames `failOpen` to `continueOnError` for clarity (GitHub Actions precedent). Not a breaking change — `failOpen` was merged days ago with no docs or adoption.
- Adds `auth` field to `McpServerConfig` for OAuth client credentials on streamable-http servers.
- Adds `clientName` and `continueOnError` public getters to `McpClient`.
- Extracts `McpClientOptions` from `McpClientConfig` for clean type reuse.

## Related Issues

strands-agents/sdk-typescript#850, strands-agents/sdk-python#2492

Builds on strands-agents/sdk-typescript#1059 which added url + auth fields to McpClientConfig.

## Type of Change

New feature / enhancement

## Testing

- [x] I ran `npm run check` (lint, typecheck, 2717 tests pass)
- [x] Unit tests cover: transport detection, env interpolation, file loading, defaults/overrides, error cases, continueOnError skip behavior, auth+sse rejection, invalid config shape

## Checklist

----

### DIFF
--- a/strands-ts/src/__tests__/mcp-config.test.node.ts
+++ b/strands-ts/src/__tests__/mcp-config.test.node.ts
@@ -1,0 +1,404 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js'
+import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/streamableHttp.js'
+import { SSEClientTransport } from '@modelcontextprotocol/sdk/client/sse.js'
+import { McpClient } from '../mcp.js'
+
+vi.mock('node:fs/promises', () => ({
+  readFile: vi.fn(),
+}))
+
+vi.mock('node:os', () => ({
+  homedir: vi.fn(() => '/home/user'),
+}))
+
+vi.mock('node:path', () => ({
+  join: (...segments: string[]) => segments.join('/'),
+}))
+
+vi.mock('@modelcontextprotocol/sdk/client/stdio.js', () => ({
+  StdioClientTransport: vi.fn(function () {}),
+  getDefaultEnvironment: vi.fn(() => ({ PATH: '/usr/bin', HOME: '/home/user' })),
+}))
+
+vi.mock('@modelcontextprotocol/sdk/client/streamableHttp.js', () => ({
+  StreamableHTTPClientTransport: vi.fn(function () {}),
+}))
+
+vi.mock('@modelcontextprotocol/sdk/client/sse.js', () => ({
+  SSEClientTransport: vi.fn(function () {}),
+}))
+
+vi.mock('@modelcontextprotocol/sdk/client/index.js', () => ({
+  Client: vi.fn(function (this: Record<string, unknown>) {
+    this.connect = vi.fn()
+    this.close = vi.fn()
+    this.listTools = vi.fn()
+    this.callTool = vi.fn()
+    this.setRequestHandler = vi.fn()
+    this.setNotificationHandler = vi.fn()
+    this.getServerCapabilities = vi.fn()
+    this.getServerVersion = vi.fn()
+    this.getInstructions = vi.fn()
+    this.experimental = { tasks: { callToolStream: vi.fn() } }
+  }),
+}))
+
+describe('McpClient.loadServers', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  describe('transport detection', () => {
+    it('creates StdioClientTransport when command is present', async () => {
+      const clients = await McpClient.loadServers({
+        'my-server': { command: 'node', args: ['server.js'] },
+      })
+
+      expect(clients).toHaveLength(1)
+      expect(StdioClientTransport).toHaveBeenCalledWith({
+        command: 'node',
+        args: ['server.js'],
+      })
+    })
+
+    it('creates McpClient with url when url is present', async () => {
+      const clients = await McpClient.loadServers({
+        'remote-server': { url: 'https://example.com/mcp' },
+      })
+
+      expect(clients).toHaveLength(1)
+      expect(StdioClientTransport).not.toHaveBeenCalled()
+      expect(SSEClientTransport).not.toHaveBeenCalled()
+    })
+
+    it('creates SSEClientTransport when transport is "sse"', async () => {
+      const clients = await McpClient.loadServers({
+        'sse-server': { url: 'https://example.com/sse', transport: 'sse' },
+      })
+
+      expect(clients).toHaveLength(1)
+      expect(SSEClientTransport).toHaveBeenCalledWith(new URL('https://example.com/sse'), undefined)
+    })
+
+    it('explicit transport overrides auto-detection', async () => {
+      const clients = await McpClient.loadServers({
+        server: { url: 'https://example.com/mcp', transport: 'sse' },
+      })
+
+      expect(clients).toHaveLength(1)
+      expect(SSEClientTransport).toHaveBeenCalled()
+      expect(StreamableHTTPClientTransport).not.toHaveBeenCalled()
+    })
+
+    it('interpolates auth credentials for streamable-http', async () => {
+      vi.stubEnv('CLIENT_ID', 'my-id')
+      vi.stubEnv('CLIENT_SECRET', 'my-secret')
+
+      const clients = await McpClient.loadServers({
+        server: {
+          url: 'https://example.com/mcp',
+          auth: { clientId: '${CLIENT_ID}', clientSecret: '${CLIENT_SECRET}' },
+        },
+      })
+
+      expect(clients).toHaveLength(1)
+    })
+
+    it('throws when auth credential references missing env var', async () => {
+      vi.unstubAllEnvs()
+
+      await expect(
+        McpClient.loadServers({
+          server: {
+            url: 'https://example.com/mcp',
+            auth: { clientId: '${MISSING_ID}', clientSecret: 'literal' },
+          },
+        })
+      ).rejects.toThrow('Environment variable "MISSING_ID" is not set')
+    })
+  })
+
+  describe('env interpolation', () => {
+    it('interpolates ${VAR} in env values', async () => {
+      vi.stubEnv('MY_SECRET', 'secret123')
+
+      await McpClient.loadServers({
+        server: { command: 'node', env: { SECRET: '${MY_SECRET}' } },
+      })
+
+      expect(StdioClientTransport).toHaveBeenCalledWith({
+        command: 'node',
+        env: { PATH: '/usr/bin', HOME: '/home/user', SECRET: 'secret123' },
+      })
+    })
+
+    it('interpolates ${VAR} in headers', async () => {
+      vi.stubEnv('TOKEN', 'abc')
+
+      await McpClient.loadServers({
+        server: { url: 'https://example.com/sse', transport: 'sse', headers: { Authorization: 'Bearer ${TOKEN}' } },
+      })
+
+      expect(SSEClientTransport).toHaveBeenCalledWith(new URL('https://example.com/sse'), {
+        requestInit: { headers: { Authorization: 'Bearer abc' } },
+      })
+    })
+
+    it('interpolates ${VAR} in url', async () => {
+      vi.stubEnv('HOST', 'myhost.com')
+
+      const clients = await McpClient.loadServers({
+        server: { url: 'https://${HOST}/mcp', transport: 'sse' },
+      })
+
+      expect(clients).toHaveLength(1)
+      expect(SSEClientTransport).toHaveBeenCalledWith(new URL('https://myhost.com/mcp'), undefined)
+    })
+
+    it('throws when env var is not set', async () => {
+      vi.unstubAllEnvs()
+
+      await expect(
+        McpClient.loadServers({
+          server: { command: 'node', env: { VAL: '${NONEXISTENT_VAR}' } },
+        })
+      ).rejects.toThrow('Environment variable "NONEXISTENT_VAR" is not set')
+    })
+
+    it('skips server with missing env var when continueOnError is true', async () => {
+      vi.unstubAllEnvs()
+
+      const clients = await McpClient.loadServers({
+        broken: { command: 'node', env: { VAL: '${NONEXISTENT_VAR}' }, continueOnError: true },
+        working: { command: 'node' },
+      })
+
+      expect(clients).toHaveLength(1)
+      expect(clients[0]!.clientName).toBe('working')
+    })
+
+    it('interpolates ${VAR} in cwd', async () => {
+      vi.stubEnv('PROJECT_DIR', '/home/user/projects')
+
+      await McpClient.loadServers({
+        server: { command: 'node', cwd: '${PROJECT_DIR}/my-server' },
+      })
+
+      expect(StdioClientTransport).toHaveBeenCalledWith({
+        command: 'node',
+        cwd: '/home/user/projects/my-server',
+      })
+    })
+
+    it('merges env with default environment', async () => {
+      await McpClient.loadServers({
+        server: { command: 'node', env: { CUSTOM: 'value' } },
+      })
+
+      expect(StdioClientTransport).toHaveBeenCalledWith({
+        command: 'node',
+        env: { PATH: '/usr/bin', HOME: '/home/user', CUSTOM: 'value' },
+      })
+    })
+  })
+
+  describe('file config loading', () => {
+    it('reads and parses a JSON file', async () => {
+      const { readFile } = await import('node:fs/promises')
+      vi.mocked(readFile).mockResolvedValue(
+        JSON.stringify({
+          'my-server': { command: 'node', args: ['server.js'] },
+        })
+      )
+
+      const clients = await McpClient.loadServers('/path/to/config.json')
+
+      expect(readFile).toHaveBeenCalledWith('/path/to/config.json', 'utf-8')
+      expect(clients).toHaveLength(1)
+    })
+
+    it('extracts mcpServers key when present', async () => {
+      const { readFile } = await import('node:fs/promises')
+      vi.mocked(readFile).mockResolvedValue(
+        JSON.stringify({
+          mcpServers: {
+            'server-a': { command: 'node' },
+            'server-b': { url: 'https://example.com' },
+          },
+        })
+      )
+
+      const clients = await McpClient.loadServers('/path/to/config.json')
+
+      expect(clients).toHaveLength(2)
+    })
+
+    it('uses whole object when mcpServers key is absent', async () => {
+      const { readFile } = await import('node:fs/promises')
+      vi.mocked(readFile).mockResolvedValue(
+        JSON.stringify({
+          'server-a': { command: 'node' },
+        })
+      )
+
+      const clients = await McpClient.loadServers('/path/to/config.json')
+
+      expect(clients).toHaveLength(1)
+    })
+
+    it('expands ~ to home directory', async () => {
+      const { readFile } = await import('node:fs/promises')
+      vi.mocked(readFile).mockResolvedValue(
+        JSON.stringify({
+          server: { command: 'node' },
+        })
+      )
+
+      await McpClient.loadServers('~/config/mcp.json')
+
+      expect(readFile).toHaveBeenCalledWith('/home/user/config/mcp.json', 'utf-8')
+    })
+  })
+
+  describe('defaults and per-server overrides', () => {
+    it('per-server continueOnError overrides defaults', async () => {
+      const clients = await McpClient.loadServers(
+        {
+          'strict-server': { command: 'node', continueOnError: false },
+          'lenient-server': { command: 'node', continueOnError: true },
+        },
+        { continueOnError: true }
+      )
+
+      expect(clients[0]!.continueOnError).toBe(false)
+      expect(clients[1]!.continueOnError).toBe(true)
+    })
+
+    it('applies default continueOnError when server does not override', async () => {
+      const clients = await McpClient.loadServers({ server: { command: 'node' } }, { continueOnError: true })
+
+      expect(clients[0]!.continueOnError).toBe(true)
+    })
+
+    it('uses server name as applicationName when not in defaults', async () => {
+      const clients = await McpClient.loadServers({
+        'my-named-server': { command: 'node' },
+      })
+
+      expect(clients[0]!.clientName).toBe('my-named-server')
+    })
+
+    it('uses defaults applicationName over server name', async () => {
+      const clients = await McpClient.loadServers({ server: { command: 'node' } }, { applicationName: 'my-app' })
+
+      expect(clients[0]!.clientName).toBe('my-app')
+    })
+  })
+
+  describe('error cases', () => {
+    it('throws when server has neither command nor url', async () => {
+      await expect(McpClient.loadServers({ bad: {} })).rejects.toThrow(
+        'Server config must include either "command" (stdio) or "url" (http)'
+      )
+    })
+
+    it('throws when stdio transport specified without command', async () => {
+      await expect(McpClient.loadServers({ bad: { transport: 'stdio' } })).rejects.toThrow(
+        'Stdio transport requires "command" field'
+      )
+    })
+
+    it('throws when streamable-http transport specified without url', async () => {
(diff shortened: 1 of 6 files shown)


strands-ts/src/mcp.ts
```diff
@@ -109,12 +94,30 @@ export type McpClientConfig = RuntimeConfig & {
   elicitationCallback?: ElicitationCallback
 
   /** When true, connection failures are logged as warnings instead of throwing. */
-  failOpen?: boolean
+  continueOnError?: boolean
```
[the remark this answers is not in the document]
gautamsirdeshmukh: `failOpen` was merged days ago, we haven't put out any docs about the new MCP enhancements yet. Renaming to `continueOnError` (both are used commonly across the industry but the former is quite confusing and open to interpretation IMO, unnecessary work for the user)

strands-ts/src/mcp-config.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+function interpolateRecord(record: Record<string, string>): Record<string, string> {
+  return Object.fromEntries(Object.entries(record).map(([k, v]) => [k, interpolateEnv(v)]))
+}
+
+async function resolveConfig(config: string | McpServersConfig): Promise<McpServersConfig> {
+  if (typeof config !== 'string') return config
+
+  const { readFile } = await import('node:fs/promises')
```
mkmeral, line 173: why are we doing inline imports? and why are we reading file as part of config resolution? would this work in a website/frontend where we don't have node env?

strands-ts/src/mcp-config.ts
```diff
@@ -0,0 +1,146 @@
+import type { McpClientCredentials, McpClientOptions, TasksConfig } from './mcp.js'
+import { McpClient } from './mcp.js'
+import { logger } from './logging/index.js'
+
+/** Configuration for a single MCP server entry in a config file or object. */
+export interface McpServerConfig {
```
mkmeral, line 10: Do we want a single config?

Can we do StdioConfig StreamableHttpConfig etc (and then combine them like (McpServerConfig = StdioConfig & StreamableHttpConfig) ? or would it make more sense to have it together? I was thinking we can just defer validation to data types essentially?

strands-ts/src/mcp.ts
```diff
@@ -109,12 +94,30 @@ export type McpClientConfig = RuntimeConfig & {
   elicitationCallback?: ElicitationCallback
 
   /** When true, connection failures are logged as warnings instead of throwing. */
-  failOpen?: boolean
```
mkmeral, line 112: why are we making a backwards incompatible change?

mkmeral (changes requested):
the most critical issue: why are we making backwards incompatible change?

strands-ts/src/mcp-config.ts
zastrowm, line 1: Why when you switch to config-mode, do you need to introduce await?

Without:

```ts
tools: [
    new McpClient({
      transport: new StdioClientTransport({
        command: 'uvx',
        args: ['awslabs.aws-documentation-mcp-server@latest'],
        env: { ...process.env, AWS_PROFILE: process.env.AWS_PROFILE! },
      }),
    }),
    new McpClient({
      transport: new StreamableHTTPClientTransport(
        new URL('https://api.githubcopilot.com/mcp/'),
        { requestInit: { headers: { Authorization: `Bearer ${process.env.GITHUB_TOKEN}` } } }
      ),
    }),
  ],
```

With:

```ts
 tools: await loadMcpServers({
    'aws-docs': {
      command: 'uvx',
      args: ['awslabs.aws-documentation-mcp-server@latest'],
      env: { AWS_PROFILE: '${AWS_PROFILE}' },
    },
    'github': {
      url: 'https://api.githubcopilot.com/mcp/',
      headers: { Authorization: 'Bearer ${GITHUB_TOKEN}' },
    },
  }),
```

zastrowm: Is `Bearer ${GITHUB_TOKEN}` interpolates the environment variable via a normal string?  In this case, why wouldn't we guide them to use template strings?

strands-ts/src/mcp-config.ts
```diff
@@ -0,0 +1,146 @@
...
+  return Object.fromEntries(Object.entries(record).map(([k, v]) => [k, interpolateEnv(v)]))
+}
+
+async function resolveConfig(config: string | McpServersConfig): Promise<McpServersConfig> {
+  if (typeof config !== 'string') return config
+
+  const { readFile } = await import('node:fs/promises')
+  const { homedir } = await import('node:os')
+
+  const filePath = config.startsWith('~') ? homedir() + config.slice(1) : config
```
zastrowm: Wow I was surprised that this is not part of node.

But then again, I guess that's normal

strands-ts/src/mcp-config.ts
zastrowm, line 10: How does it decide which transport it's using if it's not included?  That should be in the description and the PR description

strands-ts/src/mcp.ts
zastrowm, line 1: Did we explore static apis like `McpClient.fromConfig` or `McpClient.loadFrom` or something?

strands-ts/src/__tests__/mcp-config.test.node.ts
zastrowm, line 1: Ensure this is api bar-raised as well

strands-ts/src/mcp.ts
gautamsirdeshmukh: `failOpen` was merged a few days before this PR with no docs, no samples, and no adoption. It's not in any published docs yet, I'd rather not be stuck with a confusing name forever. Happy to add a deprecated alias if you feel strongly but I highly doubt anyone is using this yet.

strands-ts/src/mcp-config.ts
gautamsirdeshmukh: The dynamic imports are what keep a few of these node imports out of the static dependency graph — without them, the SDK's `check:browser-bundle` CI check fails. TLDR, we need this for browser compatibility

The file read piece is there to support loading config files from desktop (~/config.json). If you pass an object instead of a string, the file-reading code never runs.

strands-ts/src/mcp-config.ts
gautamsirdeshmukh: Yeah, Node treats ~ like an actual directory name

strands-ts/src/mcp-config.ts
gautamsirdeshmukh: `command` present → stdio, `url` present → streamable-http. I'll add this the TSDoc and PR description

strands-ts/src/mcp.ts
gautamsirdeshmukh: I did... `loadMcpServers` handles file parsing, env interpolation and `continueOnError` across the set, which feels like a more involved process than what `McpClient` should be responsible for. More importantly, `loadMcpServers` takes a multi-server config and returns `McpClient[]` — a static factory on `McpClient` would imply constructing one client, not many

strands-ts/src/mcp-config.ts
gautamsirdeshmukh: The "without" example uses top-level static imports that the user writes themselves (they import StdioClientTransport, StreamableHTTPClientTransport directly). In `loadMcpServers`, we use dynamic await import() to lazy-load those transports so they don't end up in the browser bundle (prevent CI breakages). In other words, we only load and construct what we need for the given runtime environment.

In practice, this doesn't change DX since MCP clients need an async `.connect()` before they're usable anyway, so the caller is already in an async context.

gautamsirdeshmukh: Regarding the other question, a user totally can just write `Bearer ${process.env.GITHUB_TOKEN}`. The `${VAR}` interpolation exists for when users pass a path to a JSON config file (e.g. loadMcpServers('~/config.json')). When their file contains something like:

```
{
    "github": {
      "headers": { "Authorization": "Bearer ${GITHUB_TOKEN}" }
    }
}
```


...we need our own mechanism to resolve env vars from those files. We essentially just regex and swap in process.env under the hood.

strands-ts/src/mcp-config.ts
gautamsirdeshmukh: McpClientConfig already does this same pattern, `transport` and `url` are both optional fields with runtime validation in `_resolveTransport`. Don't think we get any added value from splitting

strands-ts/src/mcp-config.ts
zastrowm: For inline-config, does it make sense to *not* do a replacement?  It seems odd not to leverage the language in that case I think?

zastrowm: > In loadMcpServers, we use dynamic await import() to lazy-load those transports so they don't end up in the browser bundle (prevent CI breakages). In other words, we only load and construct what we need for the given runtime environment.

That's needed for loading from file only or for any config?

> In practice, this doesn't change DX since MCP clients need an async .connect() before they're usable anyway, so the caller is already in an async context.

Is that true?  In the first example, there's no `connect` call?

zastrowm: > to lazy-load those transports so they don't end up in the browser bundle (prevent CI breakages)

We should validate what happens - is this a node only feature overall?  What's the CI breakage?

strands-ts/src/mcp.ts
mkmeral: but what about the discoverability of the feature? I think `McpClient.fromConfig` makes more sense to me, as it feels more discoverable. it also aligns with `Skills.fromDirectory` pattern

strands-ts/src/mcp.ts
mkmeral: I don't like this. I won't block it (probably), but I could use second opinion here cc @zastrowm 

TS is v1, so overall this is not a good pattern for us to release a feature, and break it because no one published code to github using it

> but I highly doubt anyone is using this yet.

probably you are right, but my main problem is we cannot scale "probably you are right" to introduce breaking changes :)

strands-ts/src/mcp-config.ts
gautamsirdeshmukh: > For inline-config, does it make sense to not do a replacement? It seems odd not to leverage the language in that case I think?

Even in the inline config case users can write ``Bearer ${process.env.GITHUB_TOKEN}`` if they want. The `${VAR}` interpolation is just there to make it easy (in case they want to reuse the same config shape they'd use in a JSON file)

> That's needed for loading from file only or for any config?

Any config that uses stdio. `StdioClientTransport` imports `node:process` and `node:stream` at the module level, so even if you pass an inline object with `command: 'node'`, we need the dynamic import to avoid pulling those Node-only dependencies into the static graph (which will cause the `check:browser-bundle` CI test to fail).

> Is that true? In the first example, there's no `connect` call?

I should be more clear – `.connect` is called [async](https://github.com/strands-agents/sdk-typescript/blob/main/strands-ts/src/mcp.ts#L295) inside `McpClient.listTools()`, which the Agent invokes when initializing its tool registry. So, awaiting the load-from-json call doesn't introduce new behavior.

> We should validate what happens - is this a node only feature overall? What's the CI breakage?

The stdio path is Node-only (spawns child processes). The streamable-http path is browser-compatible (fetch-based). The SDK runs `check:browser-bundle` in CI which does `esbuild --platform=browser` on `src/index.ts`. Since `mcp-config.ts` is exported from `src/index.ts`, a static import of `StdioClientTransport` would fail that check because esbuild can't resolve `node:stream` nor `node:process` for browser platform.

strands-ts/src/mcp.ts
gautamsirdeshmukh: I'm aligned with `McpClient.loadServers(...)`

strands-ts/src/mcp.ts
gautamsirdeshmukh: Agreed, initial PR could have used an API bar raise before it went out. I just discussed this with Mackenzie, plan will be to swap to `continueOnError` and add a line in the release notes. These are all useful perspectives to keep in mind going forward, especially in the post 1.0 world :)

strands-ts/src/mcp-config.ts
```diff
@@ -0,0 +1,146 @@
...
+  const parsed = JSON.parse(await readFile(filePath, 'utf-8'))
+
+  const servers = parsed?.mcpServers ?? parsed
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Unnecessarily defensive

strands-ts/src/mcp-config.ts
```diff
@@ -0,0 +1,28 @@
+import type { McpClientCredentials, TasksConfig } from './mcp.js'
+
+/** Configuration for a single MCP server entry in a config file or object. */
+export interface McpServerConfig {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Already covered by the doc on the transport field

agent-of-gsird (commented):
**Assessment**: 💬 Comment — Overall strong implementation, a few items to address before final merge

The PR has evolved substantially through multiple review rounds and is in good shape. The architecture is clean (config resolution separated from McpClient), the API is discoverable (`McpClient.loadServers`), and all prior review feedback has been properly addressed. Below are the remaining items I found after a thorough line-by-line review.

<details>
<summary>Review Details</summary>

### ✅ What's been properly addressed
- `failOpen` → `continueOnError` rename (agreed upon with maintainers)
- API moved to `McpClient.loadServers(...)` per discoverability discussion
- Tilde expansion fixed (`~/` only)
- Tests use public getters (`clientName`, `continueOnError`) instead of private properties
- TSDoc on all interface properties with interpolation documentation
- Env var interpolation throws on undefined (not silent empty string)
- JSON shape validation in `loadServersObject`
- Auth support for streamable-http, rejection for SSE
- Dynamic imports for browser bundle compatibility
- `disabled` field support
- `${env:VAR}` namespaced syntax support
- Ambiguity guard (command + url without explicit transport → error)
- `vi.stubEnv()` usage in tests

### Items to address

**1. `resolveServerConfigs` return type has redundant fields (Minor/Suggestion)**

The function returns `{ name, config, continueOnError }[]` but `McpClient.loadServers` only uses `entry.config`. Since `baseOptions` already bakes `applicationName` and `continueOnError` into the config, the extra fields in the tuple are technically redundant for the current consumer. This is fine if you anticipate external consumers of `resolveServerConfigs` needing structured access, but since it's not re-exported from `index.ts`, consider simplifying to just return `McpClientConfig[]` directly — less indirection, fewer allocations.

**2. Missing test: `${VAR}` interpolation in `command` field (Minor)**

The implementation correctly interpolates `command` (line 75: `command: interpolateEnv(server.command)`), and `args` (line 77: `server.args.map(interpolateEnv)`). Tests cover `${VAR}` in env, headers, url, and cwd — but not in `command` or `args` themselves. A test like `{ command: '${MY_CMD}', args: ['--flag=${MY_ARG}'] }` would close this gap.

**3. PR description needs `await` in examples (Already flagged, still unfixed)**

The PR description's "After" examples now show `await McpClient.loadServers(...)` correctly in some places, but the very first "After" block still shows `tools: await McpClient.loadServers(...)` which requires top-level await context. The description should note this requires an `async` IIFE or top-level await (ESM). This is a documentation-only concern.

**4. Documentation PR missing**

The checklist shows `[ ] I have updated the documentation accordingly`. Per repository guidelines, a documentation PR link or justification should be provided. Given this introduces a new public API, documentation should be tracked.

**5. `McpClientOptions` not exported from `index.ts` type-only? (Verify)**

The second parameter of `McpClient.loadServers` is typed as `McpClientOptions`. Checking exports: `McpClientOptions` IS exported from `index.ts` ✓. Users can reference it for the `defaults` parameter. Good.

### Architecture Assessment

The separation of concerns is well-executed:
- `mcp-config.ts` — Pure config resolution (parsing, interpolation, transport construction)
- `mcp.ts` — Owns `McpClient` class, exposes `loadServers` as a static factory
- `index.ts` — Exports only `McpServerConfig` type (not `resolveServerConfigs`)

The `resolveServerConfigs` function being `export` but not re-exported from index is a deliberate internal API boundary. `McpClient.loadServers` is the public entry point.

### Open Discussion Threads

Several threads from `mkmeral` and `zastrowm` are marked unresolved on GitHub but appear to have reached consensus in the comments (e.g., `McpClient.loadServers` naming, `continueOnError` rename, inline imports rationale). These should be explicitly resolved by the participants.

</details>

Overall this is well-crafted — clean code, comprehensive tests, good error messages, and the before/after DX improvement is substantial.

---

agent-of-gsird (approved):
**Assessment**: ✅ Approve

All feedback from the previous review round has been addressed. The implementation is clean, well-tested (26 test cases), and properly documented.

<details>
<summary>Review Details</summary>

- **Simplification**: `resolveServerConfigs` now returns `McpClientConfig[]` directly — previous redundant wrapper removed
- **Test gap closed**: New test covers `${VAR}` interpolation in `command` and `args` fields
- **TSDoc**: Interface-level doc now explains mutual exclusivity and auto-detection
- **Architecture**: Clean separation between config resolution (`mcp-config.ts`) and client management (`mcp.ts`)
- **Exports**: Correct API boundary — `McpClientOptions`, `McpServerConfig` public; `resolveServerConfigs` internal
- **No regressions**: All `failOpen` → `continueOnError` rename is complete, no stale references

**Non-blocking note**: Documentation PR should be tracked as a follow-up since this introduces a new public API.

</details>

---

strands-ts/src/mcp-config.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    try {
+      if (server.command && server.url && !server.transport) {
+        throw new Error('Server config has both "command" and "url" — set "transport" explicitly or remove one')
+      }
+
+      const type = server.transport ?? (server.command ? 'stdio' : server.url ? 'streamable-http' : undefined)
+      let clientConfig: McpClientConfig
+
+      if (type === 'stdio') clientConfig = await buildStdioConfig(server)
```
notowen333: could refactor this to switch + default: never so we have an exhaustiveness check at compile time.

strands-ts/src/mcp-config.ts
```diff
@@ -0,0 +1,154 @@
...
+
+async function loadServersObject(
+  config: string | Record<string, McpServerConfig>
+): Promise<Record<string, McpServerConfig>> {
+  if (typeof config !== 'string') return config
+
+  const { readFile } = await import('node:fs/promises')
+  const { homedir } = await import('node:os')
+
+  const filePath = config.startsWith('~/') ? homedir() + config.slice(1) : config
```
notowen333: There may be windows support issue on this pathing. Can pull in also `path` in dynamic import and use

`join(homedir(), config.slice(2))`

strands-ts/src/mcp.ts
```diff
@@ -123,12 +127,26 @@ export class McpClient {
...
   public static readonly DEFAULT_POLL_TIMEOUT = 300000
 
+  /**
+   * Parses an MCP servers config (file path or object) and returns McpClient instances.
+   *
+   * @param config - A file path to a JSON config, or a flat server map object.
+   * @param defaults - Options applied to all clients unless overridden per-server.
+   * @returns An array of McpClient instances ready to be passed to an Agent.
+   */
+  public static async loadServers(
```
notowen333, line 137: Are we happy with `resolveServerConfigs` only being exposed as a static method on the `McpClient` or might there also be use cases to export it independently too?

strands-ts/src/mcp-config.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+function baseOptions(name: string, server: McpServerConfig, defaults?: McpClientOptions): McpClientOptions {
+  const opts: McpClientOptions = { ...defaults, applicationName: defaults?.applicationName ?? name }
+  if (server.continueOnError != null) opts.continueOnError = server.continueOnError
+  if (server.tasksConfig != null) opts.tasksConfig = server.tasksConfig
+  return opts
+}
+
+function interpolateEnv(value: string): string {
```
notowen333, line 155: can we add a docstring with an example transformation here

strands-ts/src/mcp-config.ts
gautamsirdeshmukh: Will update to a switch

strands-ts/src/mcp.ts
gautamsirdeshmukh: We can keep it internal. The use case for exporting it independently would be resolving configs without constructing clients, but `McpClient.loadServers()` already covers this because the returned clients don't connect until first use. So, a user can just call `loadServers()` and inspect the clients without triggering any kind of I/O.

strands-ts/src/mcp-config.ts
gautamsirdeshmukh: Adding

strands-ts/src/mcp-config.ts
```diff
@@ -0,0 +1,188 @@
...
+  /** HTTP headers sent with every request (supports `${VAR}` or `${env:VAR}` interpolation). */
+  headers?: Record<string, string>
+  /** Explicit transport type. When omitted, auto-detected: `command` → stdio, `url` → streamable-http. */
+  transport?: 'stdio' | 'sse' | 'streamable-http'
+  /** Client credentials for OAuth machine-to-machine auth (streamable-http only). */
+  auth?: McpClientCredentials
+  /** When true, this server is skipped during loadServers. */
+  disabled?: boolean
+  /** When true, config or connection failures skip this server instead of throwing. */
+  continueOnError?: boolean
```
mkmeral, line 30: not a blocker: but we should document these are not convention, strands specific items (at least for the strands specific ones)

mkmeral (approved):

strands-ts/src/mcp-config.ts
gautamsirdeshmukh: Neither one of these are strands specific

PR #1073, opened 2026-05-15
feat(interventions): add human in the loop intervention handler

lizradway:
## Description

Adds a ready-to-use `HumanInTheLoop` intervention handler that gates tool execution behind human approval, using the `confirm` intervention action.

This is the first vended intervention — a concrete, configurable `InterventionHandler` shipped with the SDK, following the same pattern as vended plugins (`ContextOffloader`, `Skills`).

### Usage

```typescript
import { Agent, InterruptResponseContent } from '@strands-agents/sdk'
import { HumanInTheLoop } from '@strands-agents/sdk/vended-interventions/hitl'

// Default: all tools require approval via interrupt/resume
const agent = new Agent({
  tools: [deleteTool, readTool],
  interventions: [new HumanInTheLoop({ allowedTools: ['read_file'] })],
})

const result = await agent.invoke('Delete the temp files')
// result.stopReason === 'interrupt'
// result.interrupts[0].reason === 'Tool "delete_file" requires human approval. Input: {...}'

// Present the interrupt to your user (web UI, Slack, push notification, etc.)
// Then resume with their response:
const finalResult = await agent.invoke([
  new InterruptResponseContent({
    interruptId: result.interrupts[0].id,
    response: 'yes',  // 'y', 'yes', or true → approved. Anything else → denied.
  }),
])
```

```typescript
// CLI mode: prompts in terminal inline, no interrupt handling needed
const agent = new Agent({
  tools: [deleteTool, readTool],
  interventions: [new HumanInTheLoop({ allowedTools: ['read_file'], ask: 'stdio' })],
})

await agent.invoke('Delete the temp files')
// Terminal: "Tool "delete_file" requires human approval. Input: {...} (y/n): "
```

### Configuration

- **Default** (no `ask`): all tools require approval, agent pauses via interrupt/resume
- **`allowedTools`**: listed tools run freely, everything else requires approval
- **`allowedTools: ['*']`**: all tools run freely (disables HITL)
- **`allowedTools: ['*', '!deleteFile']`**: all tools run freely except `deleteFile`
- **`enableTrust: true`** (opt-in): trust responses approve the tool AND remember it in `agent.appState` for the session. Customizable via `evaluateTrust`.

### `ask` option

Controls how the human's response is collected.

```typescript
// Default (omitted): interrupt/resume — agent pauses, caller resumes with response
// Use this for cloud, API servers, web apps — anywhere the agent can't block inline
new HumanInTheLoop()

// CLI readline (Node.js only) — agent blocks inline until human responds
new HumanInTheLoop({ ask: 'stdio' })

// Custom UI (Slack, web socket, etc.) — agent blocks inline on your callback
new HumanInTheLoop({
  ask: async (prompt) => await slackDM(userId, prompt),
})
```

### Cloud / API server example (interrupt/resume)

```typescript
import { Agent, InterruptResponseContent } from '@strands-agents/sdk'
import { HumanInTheLoop } from '@strands-agents/sdk/vended-interventions/hitl'

const agent = new Agent({
  tools: [deleteTool, readTool],
  interventions: [new HumanInTheLoop({ allowedTools: ['read_file'] })],
  sessionManager,
})

// Request 1: Agent tries to delete → pauses
app.post('/invoke', async (req, res) => {
  const result = await agent.invoke(req.body.prompt)
  if (result.stopReason === 'interrupt') {
    res.json({
      status: 'pending_approval',
      interruptId: result.interrupts[0].id,
      reason: result.interrupts[0].reason,
    })
  } else {
    res.json({ status: 'complete', message: result.message })
  }
})

// Request 2: Client sends back the approval/denial
app.post('/approve', async (req, res) => {
  const result = await agent.invoke([
    new InterruptResponseContent({
      interruptId: req.body.interruptId,
      response: req.body.approved ? 'yes' : 'no',
    }),
  ])
  res.json({ status: 'complete', message: result.message })
})
```

### Approval response format

Defaults to accepting `true`, `'y'`/`'yes'` (case-insensitive). Customizable via `evaluate` option.

## Related Issues

- #883
- #1072

## Documentation PR

N/A

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`
- 17 unit tests covering: default interrupt mode, allowedTools, wildcard, negation pattern, inline approve/deny, trust mode (t/trust/T), negation-overrides-trust, custom evaluate/evaluateTrust, prompt metadata
- Interactive examples run locally

## Checklist

----

### DIFF
--- a/strands-ts/src/vended-interventions/hitl/__tests__/hitl.test.ts
+++ b/strands-ts/src/vended-interventions/hitl/__tests__/hitl.test.ts
@@ -360,15 +360,49 @@ describe('HumanInTheLoop', () => {
       expect(askCount).toBe(1)
     })
 
+    it('supports custom evaluateTrust function', async () => {
+      let askCount = 0
+
+      const model = new MockMessageModel()
+        .addTurn({ type: 'toolUseBlock', name: 'myTool', toolUseId: 'tool-1', input: {} })
+        .addTurn({ type: 'toolUseBlock', name: 'myTool', toolUseId: 'tool-2', input: {} })
+        .addTurn({ type: 'textBlock', text: 'Done' })
+
+      const tool = createMockTool('myTool', () => 'executed')
+
+      const agent = new Agent({
+        model,
+        tools: [tool],
+        interventions: [
+          new HumanInTheLoop({
+            enableTrust: true,
+            evaluateTrust: (r) => r === 'approve-and-remember',
+            ask: async () => {
+              askCount++
+              return 'approve-and-remember'
+            },
+          }),
+        ],
+        printer: false,
+      })
+
+      await agent.invoke('Run tool twice')
+      expect(askCount).toBe(1)
+    })
+
     it('negated tools cannot be trusted even with "t" response', async () => {
       let askCount = 0
+      let toolExecuted = false
 
       const model = new MockMessageModel()
         .addTurn({ type: 'toolUseBlock', name: 'dangerTool', toolUseId: 'tool-1', input: {} })
         .addTurn({ type: 'toolUseBlock', name: 'dangerTool', toolUseId: 'tool-2', input: {} })
         .addTurn({ type: 'textBlock', text: 'Done' })
 
-      const tool = createMockTool('dangerTool', () => 'ran')
+      const tool = createMockTool('dangerTool', () => {
+        toolExecuted = true
+        return 'ran'
+      })
 
       const agent = new Agent({
         model,
@@ -388,7 +422,7 @@ describe('HumanInTheLoop', () => {
 
       await agent.invoke('Run danger twice')
       expect(askCount).toBe(2)
+      expect(toolExecuted).toBe(false)
     })
   })
-
 })
--- a/strands-ts/src/vended-interventions/hitl/__tests__/hitl.test.ts
+++ b/strands-ts/src/vended-interventions/hitl/__tests__/hitl.test.ts
@@ -216,37 +216,6 @@ describe('HumanInTheLoop', () => {
       expect(prompts[0]).toContain('bob@example.com')
     })
 
-    it('receives the BeforeToolCallEvent for context-aware routing', async () => {
-      const model = new MockMessageModel()
-        .addTurn({ type: 'toolUseBlock', name: 'shell', toolUseId: 'tool-1', input: { command: 'ls -la' } })
-        .addTurn({ type: 'textBlock', text: 'Done' })
-
-      let toolExecuted = false
-      const tool = createMockTool('shell', () => {
-        toolExecuted = true
-        return 'output'
-      })
-
-      const agent = new Agent({
-        model,
-        tools: [tool],
-        interventions: [
-          new HumanInTheLoop({
-            ask: async (_prompt, event) => {
-              const input = event.toolUse.input as { command: string }
-              if (input.command.startsWith('ls')) return 'yes'
-              return 'no'
-            },
-          }),
-        ],
-        printer: false,
-      })
-
-      const result = await agent.invoke('List files')
-      expect(result.stopReason).toBe('endTurn')
-      expect(toolExecuted).toBe(true)
-    })
-
     it('supports custom evaluate function', async () => {
       const model = new MockMessageModel()
         .addTurn({ type: 'toolUseBlock', name: 'myTool', toolUseId: 'tool-1', input: {} })
--- a/strands-ts/src/vended-interventions/hitl/__tests__/hitl.test.ts
+++ b/strands-ts/src/vended-interventions/hitl/__tests__/hitl.test.ts
@@ -1,0 +1,415 @@
+import { describe, expect, it } from 'vitest'
+import { HumanInTheLoop } from '../hitl.js'
+import { Agent } from '../../../agent/agent.js'
+import { MockMessageModel } from '../../../__fixtures__/mock-message-model.js'
+import { createMockTool } from '../../../__fixtures__/tool-helpers.js'
+
+describe('HumanInTheLoop', () => {
+  describe('default config (interrupt/resume)', () => {
+    it('pauses agent with interrupt on any tool call', async () => {
+      const model = new MockMessageModel()
+        .addTurn({ type: 'toolUseBlock', name: 'anyTool', toolUseId: 'tool-1', input: { x: 1 } })
+        .addTurn({ type: 'textBlock', text: 'Done' })
+
+      let toolExecuted = false
+      const tool = createMockTool('anyTool', () => {
+        toolExecuted = true
+        return 'result'
+      })
+
+      const agent = new Agent({
+        model,
+        tools: [tool],
+        interventions: [new HumanInTheLoop()],
+        printer: false,
+      })
+
+      const result = await agent.invoke('Do something')
+
+      expect(result.stopReason).toBe('interrupt')
+      expect(result.interrupts).toEqual([
+        expect.objectContaining({
+          name: 'human-in-the-loop',
+          reason: expect.stringContaining('anyTool'),
+        }),
+      ])
+      expect(toolExecuted).toBe(false)
+    })
+  })
+
+  describe('inline mode (with ask callback)', () => {
+    it('allows tool execution when approved', async () => {
+      const model = new MockMessageModel()
+        .addTurn({ type: 'toolUseBlock', name: 'myTool', toolUseId: 'tool-1', input: {} })
+        .addTurn({ type: 'textBlock', text: 'Done' })
+
+      let toolExecuted = false
+      const tool = createMockTool('myTool', () => {
+        toolExecuted = true
+        return 'executed'
+      })
+
+      const agent = new Agent({
+        model,
+        tools: [tool],
+        interventions: [new HumanInTheLoop({ ask: async () => 'yes' })],
+        printer: false,
+      })
+
+      const result = await agent.invoke('Run tool')
+      expect(result.stopReason).toBe('endTurn')
+      expect(toolExecuted).toBe(true)
+    })
+
+    it('denies tool execution when rejected', async () => {
+      const model = new MockMessageModel()
+        .addTurn({ type: 'toolUseBlock', name: 'myTool', toolUseId: 'tool-1', input: {} })
+        .addTurn({ type: 'textBlock', text: 'Understood' })
+
+      let toolExecuted = false
+      const tool = createMockTool('myTool', () => {
+        toolExecuted = true
+        return 'executed'
+      })
+
+      const agent = new Agent({
+        model,
+        tools: [tool],
+        interventions: [new HumanInTheLoop({ ask: async () => 'no' })],
+        printer: false,
+      })
+
+      const result = await agent.invoke('Run tool')
+      expect(result.stopReason).toBe('endTurn')
+      expect(toolExecuted).toBe(false)
+    })
+  })
+
+  describe('allowedTools config', () => {
+    it('does not prompt for tools in allowedTools', async () => {
+      const model = new MockMessageModel()
+        .addTurn({ type: 'toolUseBlock', name: 'readFile', toolUseId: 'tool-1', input: {} })
+        .addTurn({ type: 'textBlock', text: 'Done' })
+
+      let toolExecuted = false
+      const tool = createMockTool('readFile', () => {
+        toolExecuted = true
+        return 'content'
+      })
+
+      const agent = new Agent({
+        model,
+        tools: [tool],
+        interventions: [new HumanInTheLoop({ allowedTools: ['readFile'], ask: async () => 'no' })],
+        printer: false,
+      })
+
+      const result = await agent.invoke('Read it')
+      expect(result.stopReason).toBe('endTurn')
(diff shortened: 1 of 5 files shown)


strands-ts/src/vended-interventions/hitl/hitl.ts
```diff
@@ -0,0 +1,163 @@
...
+ * // Enable trust — user can type 't' to always allow a tool for the session
+ * const agent = new Agent({
+ *   interventions: [new HumanInTheLoop({ enableTrust: true })],
+ * })
+ * ```
+ */
+export class HumanInTheLoop extends InterventionHandler {
+  readonly name = 'human-in-the-loop'
+
+  private readonly _allowedTools: Set<string>
```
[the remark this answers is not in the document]
lizradway: i kinda like separate `_trustedTools`

strands-ts/src/vended-interventions/hitl/hitl.ts
```diff
@@ -0,0 +1,163 @@
...
+   * // All tools run freely EXCEPT deleteFile and sendEmail
+   * { allowedTools: ['*', '!deleteFile', '!sendEmail'] }
+   * ```
+   */
+  allowedTools?: string[]
+
+  /**
+   * When true, the default CLI prompt includes a "trust" option (t).
+   * If the human responds with 't', the tool is approved AND added to
+   * the allowed set for the rest of the session (won't ask again).
```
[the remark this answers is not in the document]
lizradway: thought about it and happy to change if we think that's the right strategy, but i think that creates too many weird edge cases for if a user specifies both/defaults/etc.


I think that the 80% case is that the user will _not_ be explicitly allowing all tools except a certain few (but that the user will be trusting tools as they land) and tbh I find this `!` very intuitive? 

would be interested in hearing other's thoughts tho

strands-ts/src/vended-interventions/hitl/hitl.ts
```diff
@@ -0,0 +1,163 @@
...
+      })
+    }
+
+    return confirm(prompt, {
+      ...(this._evaluate && { evaluate: this._evaluate }),
+    })
+  }
+
+  private _requiresApproval(toolName: string): boolean {
+    if (this._allowedTools.has(`!${toolName}`)) return true
```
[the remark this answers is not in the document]
lizradway: !foo should always take priority. i think this is intuitive but can document.

strands-ts/examples/hitl/src/index.ts
```diff
@@ -0,0 +1,65 @@
+import { Agent, BedrockModel, tool } from '@strands-agents/sdk'
```
mkmeral: I'm not sure if this example is helpful enough. The only additional value it gives me over any boilerplate is `interventions: [new HumanInTheLoop({ allowedTools: ['read_file'] })],` line

We should add some info around how to respond, how to continue and so on. what happens when we deny?

strands-ts/src/vended-interventions/hitl/hitl.ts
```diff
@@ -0,0 +1,176 @@
...
+
+/**
+ * Default CLI prompt that reads from stdin.
+ * Serializes prompts so concurrent tool calls don't collide on stdin.
+ */
+function createDefaultAsk(includeTrust: boolean): (prompt: string) => Promise<JSONValue> {
+  const options = includeTrust ? '(y/n/t)' : '(y/n)'
+  let queue: Promise<unknown> = Promise.resolve()
+
+  return (prompt: string) => {
```
mkmeral, line 18: are we defaulting to in-process conform "I print question, use input and wait for customer response"? Was that a request from someone? 

Most of our use cases are remote or in cloud. interrupts should be default imo

strands-ts/src/vended-interventions/hitl/hitl.ts
```diff
@@ -0,0 +1,176 @@
...
+   * ```
+   */
+  allowedTools?: string[]
+
+  /**
+   * When true, the default CLI prompt includes a "trust" option (t).
+   * If the human responds with 't', the tool is approved AND added to
+   * the allowed set for the rest of the session (won't ask again).
+   *
+   * Has no effect when `ask` is `'interrupt'` (stateless mode has no session to remember).
```
mkmeral: why do we think `session = process`, we have a concept of session manager with its own session. why don't we just save these things in agent app state, and use that as session?

strands-ts/examples/hitl/src/index.ts
mkmeral: on this note, i'd rather have a local agent that loops in `while True` and shows me how to manage interrupts and continue on the loop

lizradway: oops i did not actually mean to commit example. removed

strands-ts/src/vended-interventions/hitl/hitl.ts
lizradway: interrupt is default on the [`confirm` intervention action](https://github.com/strands-agents/sdk-typescript/pull/1072/changes) so that users can implement a custom `InterventionHandler` for remote/cloud confirmation. however this `interventionhandler` defaults to CLI user response kiro/claude code style. Users can override `evaluate` for custom cloud/remote impl

lizradway: user would likely write their own thin intervention handler for cloud use case or be overriding a lot of our specifications since there is no default case we can provide for that.

strands-ts/src/vended-interventions/hitl/hitl.ts
lizradway: makes sense! can update

strands-ts/src/vended-interventions/hitl/hitl.ts
mkmeral: > however this interventionhandler defaults to CLI user response kiro/claude code style

why did we decide to scope it to CLI and local? Also this wouldn't work out of the box for browser, right? how does stdio work in browser? 🤔 

my 2c: we should default to interrupts. that's our HITL mechanism that supports cross-session interrupts by default. if someone wants to do in-process stdio HITL, we can just add a separate param for `ask`, like `ask: "stdio"` or sth

strands-ts/src/agent/__tests__/printer.test.ts
```diff
@@ -132,71 +131,7 @@ describe('AgentPrinter', () => {
       await collectGenerator(agent.stream('Test'))
 
       const allOutput = outputs.join('')
-      expect(allOutput).toBe('\n  ⏳ bad_tool\n\n🔧 Tool #1: bad_tool\n✗ Tool failed\nError handled\n')
```
mkmeral, line 135: why are we deleting all these tests?

strands-ts/src/vended-interventions/hitl/hitl.ts
lizradway: if we are defaulting to interrupt, we do not need a human in the loop intervention handler as it would be defined like this:
```
override beforeToolCall(event) {
  if (!this._requiresApproval(event)) return proceed()
  return confirm(prompt)
}
```
that is just the `confirm` intervention action and can be implemented by a user:  see https://github.com/strands-agents/sdk-typescript/pull/1072/

this has 0 value add imo


Maybe what you are looking for is more specific naming of this then? open to suggestions if you're cool with that

lizradway: > why did we decide to scope it to CLI and local?

This is not scoped for CLI/local, that is just the fallback if `ask` is not configured. If I were building a CLI, i would want it to _just work_ whereas if I were building a cloud/production environment, I would need to configure the confirmation `ask` strategy and would never hit the fallback strategy anyways.

> Also this wouldn't work out of the box for browser, right? how does stdio work in browser? 🤔

No the fallback wouldn't work in the browser. The user would need to specify the `ask` strategy, but that is implied imo for both cloud + browser use case as there is not default used for that.

mkmeral: > that is just the confirm intervention action and can be implemented by a user

that's where I disagree. you implicitly got rid of all config and trust logic by putting `this._requiresApproval` in your example 🙃 

I'd expect people to just be able to configure allowed tools, trust, etc. simply like we are doing it here.

> but that is implied imo for both cloud + browser use case as there is not default used for that.

but why have a broken "default" experience in half of the places, and only have it working on CLI? if I have to configure, as I move my locally tested agent to cloud, everyone will run into an issue where their HITL handler makes their agent stuck in cloud. That's what I am worried about


---

Overall I think this is a DevX question not necessarily architecture. Would we want simpler experience (like strands tool in python) today for CLI, while we want explicit configuration for cloud? or would we want a bit more complicated getting started experience (i.e. you need to handle interrupt output), but you can move the logic around seamlessly? 

Given that current default adds friction while moving to browser, cloud, etc + it doesn't actually work with session manager, I'd like to default to interrupt. That said I think I can be convinced here 🤔

strands-ts/src/agent/__tests__/printer.test.ts
mkmeral: it seems like we are updating printer. is this the best solution? 🤔 

can't we just call before tool call hooks for printer at the end? now that we have hook order :)

strands-ts/src/vended-interventions/hitl/hitl.ts
lizradway: yeah id be interested to raise this devx question to the team to see what they think.

i think the usefulness of an HITL interventions handler if it defaults to explicit interrupt would be nil since the handler becomes a simple passthrough to `confirm/interrupt`.

however, i understand that opinionated local default that is overridden for production/cloud can make the production migration non-obvious, but in this case in particular I would assume that a user setting up HITL in cloud would expect to specify how they will be verifying HITL (ie. will define `ask`) since that is an inherently bespoke use case.

part of me is convinced that this devx could be resolved by naming specification or obvious documentation.

strands-ts/src/agent/__tests__/printer.test.ts
lizradway: stale / pulled in rebase

lizradway: > can't we just call before tool call hooks for printer at the end? now that we have hook order :) 


down to do that. already merged in the printer fix, but can update this later

strands-ts/src/vended-interventions/hitl/hitl.ts
zastrowm: The overall experience of:

> using the confirm intervention action with inline resolution

is non-intuitive to me, esp considering that we're using interrupts as the underlying mechanism.   If we're using interrupts, I think the handling of accept should be *above* the agent; our stance on interrupts/user-input should center around async input and not necessarily holding onto CPU waiting for an in-process response.

Defaulting to CLI experience is even weirder to me at this point IMHO

lizradway: defaulting to interrupts 👍

strands-ts/src/vended-interventions/hitl/hitl.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  override async beforeToolCall(event: BeforeToolCallEvent): Promise<InterventionAction> {
+    const toolName = event.toolUse.name
+    if (!this._requiresApproval(event)) {
+      return proceed()
+    }
+
+    const prompt = `Tool "${toolName}" requires human approval. Input: ${JSON.stringify(event.toolUse.input)}`
+
+    if (!this._ask) {
```
mkmeral, line 164: > kiro: So enableTrust: true with no ask is effectively a no-op. It won't error, but it won't do anything either

Any reason interrupts don't work with trust?

lizradway: originally, trust was only here for the cli default use case really (y/n/t). but now that default is interrupt for cloud/browser use case, i think it makes sense to extend support since we still need the top level param to enable trust.

mkmeral (dismissed):

strands-ts/src/vended-interventions/hitl/hitl.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    if (!this._ask) {
+      return confirm(prompt, { evaluate })
+    }
+
+    const response = await this._ask(prompt)
+
+    if (this._enableTrust && this._evaluateTrust(response)) {
+      this._trustTool(event, toolName)
```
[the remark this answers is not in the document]
mkmeral: +1 on this suggestion, not a blocker tho

mkmeral (dismissed):

mkmeral (dismissed):

strands-ts/src/vended-interventions/hitl/__tests__/hitl.test.ts
zastrowm, line 1: Can you update the PR description (and eventually docs) with the cloud permission example?  E.g. what happens when you don't have an inline `ask`

mkmeral (dismissed):

mkmeral:
I'm okay with that ^ we can add later

mkmeral (approved):

PR #1082, opened 2026-05-18
feat: add goal vended plugin

notowen333:
### Motivation

Real agent workflows often need iterative refinement — write code that passes tests, draft prose under constraints, generate output that satisfies a domain check. Today users hand-roll this with `BeforeInvocationEvent` / `AfterInvocationEvent` hooks; the result is repetitive and easy to get wrong (budget enforcement, judge transcript isolation, resume-prompt construction, transcript rewinds for fresh-context retries).

`GoalLoop` is the SDK analogue of Claude Code's `/goal` command: validate the agent's response after each invocation, feed validator feedback back as a fresh user message via `AfterInvocationEvent.resume`, and loop until validation passes or a budget is exhausted.

### Two retry modes

The plugin supports two retry semantics, picked with a single flag.

**Standard mode (default)** preserves the conversation across attempts — the agent sees its own prior responses and the validator's feedback messages. Use when prior attempts inform the next one (iterative prose refinement, conversational tightening).

**Ralph Wiggum mode (`preserveContext: false`)** restores the agent's transcript to its initial post-input state on every retry. The agent never sees its own prior attempts — only the original input plus the latest validator feedback. Use when prior context misleads more than it helps (agent doubles down on a bad approach) and when flat inference cost across attempts matters. Pairs naturally with toolchain-driven validators (e.g. `npm test`) where the agent has tools to fix what the validator reports.

Critically, Ralph mode rewinds the full model-visible session — `messages`, `systemPrompt`, `modelState`, and `interrupts` all restore to the pre-attempt-1 snapshot, so the model (including any server-side conversation chaining) sees only the original input. Only `appState` accumulates across attempts, since that's plugin scratch space invisible to the model — other plugins (rate-limiters, cost-trackers) are unaffected.

### Public API Changes

```typescript
import { Agent } from '@strands-agents/sdk'
import { GoalLoop } from '@strands-agents/sdk/vended-plugins/goal'

// Natural-language goal — judged by an auto-built Agent against the host's transcript.
const concise = new GoalLoop({
  goal: 'At most 3 sentences, accessible to a 10-year-old, no jargon.',
  maxAttempts: 3,
})

// Programmatic predicate — pass a function as `goal` to run your own check.
const wordCount = new GoalLoop({
  goal: (response) => {
    const text = response.content
      .flatMap((b) => (b.type === 'textBlock' ? [b.text] : []))
      .join(' ')
    const words = text.trim().split(/\s+/).length
    return words <= 50 || { passed: false, feedback: `Too long (${words}). Cap is 50.` }
  },
  maxAttempts: 5,
})

// Toolchain-driven Ralph loop — restart from clean state every retry.
const npmTestLoop = new GoalLoop({
  goal: async () => {
    try { await execAsync('npm test'); return true }
    catch (err) {
      const e = err as { stdout?: string; stderr?: string }
      return { passed: false, feedback: `${e.stdout ?? ''}${e.stderr ?? ''}`.slice(-4000) }
    }
  },
  preserveContext: false,
  maxAttempts: 10,
})

const agent = new Agent({ model, plugins: [concise] })
await agent.invoke('Explain how rainbows form.')

console.log(concise.lastResult(agent))
// { passed: true, stopReason: 'satisfied', attempts: [{ attempt: 1, passed: true }] }
```

`lastResult(agent)` exposes the terminal outcome for that agent — `stopReason` is one of `'satisfied' | 'maxAttempts' | 'timeout'`, with the per-attempt feedback preserved on `attempts`.

### Customization

The defaults handle the common cases, but several extensibility surfaces are exposed for power users:

- **`judge`** — tuning for the auto-built NL judge, grouped into a single `JudgeConfig` object so it's clear the knobs apply only when `goal` is a string (harmlessly ignored for function validators):
  - **`judge.model`** — swap the judge to a cheaper/faster model (e.g. Haiku) while keeping the heavyweight model on the host. Mirrors Claude Code's fast-evaluator default.
  - **`judge.systemPrompt`** — override the judge's system prompt (defaults to `JUDGE_SYSTEM_PROMPT`) to retune the rubric, localize the instructions, or tighten/relax the pass criteria.
- **`resumePromptTemplate`** — override the canned English retry message. Useful for non-English agents, terser framing, or embedding feedback in a domain-specific structure.
- **Judge primitives** — `JUDGE_OUTCOME_SCHEMA`, `JUDGE_SYSTEM_PROMPT`, and `buildJudgePrompt` are exported, so users wanting a fully custom NL judge can wire one through a function validator without forking.

```typescript
new GoalLoop({
  goal: 'be concise',
  judge: {
    model: new BedrockModel({
      modelId: 'anthropic.claude-haiku-4-5-20251001-v1:0',
      maxTokens: 256,
    }),
    systemPrompt: '...custom judging rubric...',
  },
  resumePromptTemplate: (feedback) =>
    feedback ? `Last attempt was wrong: ${feedback}. Fix and retry.` : 'Wrong, retry.',
  preserveContext: false,
  maxAttempts: 5,
  timeout: 60_000,
})
```

## Related Issues

N/A — new feature

## Documentation PR

TBD — agent-docs PR to follow.

## Type of Change

New feature

## Testing

- Unit tests cover all `goal` shapes (string / function / async), lifecycle (mid-run, post-throw, multi-invoke), multi-instance stacking, multi-agent rejection, Ralph mode (transcript rewind, appState survival), the `resumePromptTemplate` override, and the `judge.model` / `judge.systemPrompt` overrides.
- Package-compatibility suites (ESM, CJS, npm-pack) verify the `./vended-plugins/goal` subpath export and `GoalLoop`'s presence in the `vended-plugins` barrel, alongside `AgentSkills` and `ContextOffloader`.
- Integration tests against Bedrock cover both standard and Ralph modes end-to-end. Assertions are deterministic by design — the validator drives loop length, model output never gates pass/fail.
- [ ] I ran `npm run check`

## Checklist

----

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3747,6 +3747,8 @@
     },
     "node_modules/@types/chai": {
       "version": "5.2.3",
+      "resolved": "https://registry.npmjs.org/@types/chai/-/chai-5.2.3.tgz",
+      "integrity": "sha512-Mw558oeA9fFbv65/y4mHtXDs9bPnFMZAL/jxdPFUpOHHIXX91mcgEHbS5Lahr+pwZFR8A7GQleRWeI6cGFC2UA==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -3764,6 +3766,8 @@
     },
     "node_modules/@types/deep-eql": {
       "version": "4.0.2",
+      "resolved": "https://registry.npmjs.org/@types/deep-eql/-/deep-eql-4.0.2.tgz",
+      "integrity": "sha512-c9h9dVVMigMPc4bwTvC5dxqtqJZwQPePsWjPlpSOnojbor6pGqdk541lfA7AqFQr5pB1BRdq0juY9db81BwyFw==",
       "dev": true,
       "license": "MIT"
     },
@@ -4067,87 +4071,17 @@
         "url": "https://opencollective.com/eslint"
       }
     },
-    "node_modules/@vitest/browser": {
-      "version": "4.1.4",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@blazediff/core": "1.9.1",
-        "@vitest/mocker": "4.1.4",
-        "@vitest/utils": "4.1.4",
-        "magic-string": "^0.30.21",
-        "pngjs": "^7.0.0",
-        "sirv": "^3.0.2",
-        "tinyrainbow": "^3.1.0",
-        "ws": "^8.19.0"
-      },
-      "funding": {
-        "url": "https://opencollective.com/vitest"
-      },
-      "peerDependencies": {
-        "vitest": "4.1.4"
-      }
-    },
-    "node_modules/@vitest/browser-playwright": {
-      "version": "4.1.4",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@vitest/browser": "4.1.4",
-        "@vitest/mocker": "4.1.4",
-        "tinyrainbow": "^3.1.0"
-      },
-      "funding": {
-        "url": "https://opencollective.com/vitest"
-      },
-      "peerDependencies": {
-        "playwright": "*",
-        "vitest": "4.1.4"
-      },
-      "peerDependenciesMeta": {
-        "playwright": {
-          "optional": false
-        }
-      }
-    },
-    "node_modules/@vitest/coverage-v8": {
-      "version": "4.1.4",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@bcoe/v8-coverage": "^1.0.2",
-        "@vitest/utils": "4.1.4",
-        "ast-v8-to-istanbul": "^1.0.0",
-        "istanbul-lib-coverage": "^3.2.2",
-        "istanbul-lib-report": "^3.0.1",
-        "istanbul-reports": "^3.2.0",
-        "magicast": "^0.5.2",
-        "obug": "^2.1.1",
-        "std-env": "^4.0.0-rc.1",
-        "tinyrainbow": "^3.1.0"
-      },
-      "funding": {
-        "url": "https://opencollective.com/vitest"
-      },
-      "peerDependencies": {
-        "@vitest/browser": "4.1.4",
-        "vitest": "4.1.4"
-      },
-      "peerDependenciesMeta": {
-        "@vitest/browser": {
-          "optional": true
-        }
-      }
-    },
     "node_modules/@vitest/expect": {
-      "version": "4.1.4",
+      "version": "4.1.8",
+      "resolved": "https://registry.npmjs.org/@vitest/expect/-/expect-4.1.8.tgz",
+      "integrity": "sha512-h3nDO677RDLEGlBxyQ5CW8RlMThSKSRLUePLOx09gNIWRL40edgA1GCZSZgf1W55MFAG6/Sw14KeaAnqv0NKdQ==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
         "@standard-schema/spec": "^1.1.0",
         "@types/chai": "^5.2.2",
-        "@vitest/spy": "4.1.4",
-        "@vitest/utils": "4.1.4",
+        "@vitest/spy": "4.1.8",
+        "@vitest/utils": "4.1.8",
         "chai": "^6.2.2",
         "tinyrainbow": "^3.1.0"
       },
@@ -4156,11 +4090,13 @@
       }
     },
     "node_modules/@vitest/mocker": {
-      "version": "4.1.4",
+      "version": "4.1.8",
+      "resolved": "https://registry.npmjs.org/@vitest/mocker/-/mocker-4.1.8.tgz",
+      "integrity": "sha512-LEiN/xe4OSIbKe9HQIp5OC24agGD9J5CnmMgsLohVVoOPWL9a2sBoR6VBx43jQZb7Kr1l4RCuyCJzcAa0+dojw==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
-        "@vitest/spy": "4.1.4",
+        "@vitest/spy": "4.1.8",
         "estree-walker": "^3.0.3",
         "magic-string": "^0.30.21"
       },
@@ -4181,7 +4117,9 @@
       }
     },
     "node_modules/@vitest/pretty-format": {
-      "version": "4.1.4",
+      "version": "4.1.8",
+      "resolved": "https://registry.npmjs.org/@vitest/pretty-format/-/pretty-format-4.1.8.tgz",
+      "integrity": "sha512-9GasEBxpZ1VYIpqHf/0+YGg121uSNwCKOJqIrTwWP/TB7DmFCiaBpNl3aPZzoLWfWkuqhbH8vJIVobZkvdo2cA==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -4192,11 +4130,13 @@
       }
     },
     "node_modules/@vitest/runner": {
-      "version": "4.1.4",
+      "version": "4.1.8",
+      "resolved": "https://registry.npmjs.org/@vitest/runner/-/runner-4.1.8.tgz",
+      "integrity": "sha512-EmVxeBAfMJvycdjd6Hm+RbFBbA9fKvo0Kx37hNpBYoYeavH3RNsBXWDooR1mgD52dCrxIIuP7UotpfiwOikvcg==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
-        "@vitest/utils": "4.1.4",
+        "@vitest/utils": "4.1.8",
         "pathe": "^2.0.3"
       },
       "funding": {
@@ -4204,12 +4144,14 @@
       }
     },
     "node_modules/@vitest/snapshot": {
-      "version": "4.1.4",
+      "version": "4.1.8",
+      "resolved": "https://registry.npmjs.org/@vitest/snapshot/-/snapshot-4.1.8.tgz",
+      "integrity": "sha512-acfZboRmAIf05DEKcBQy33VXojFJjtUdLyo7oOmV9kebb2xdU01UknNiPuPZoJZQyO7DF0gZdTGTpeAzET9QPQ==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
-        "@vitest/pretty-format": "4.1.4",
-        "@vitest/utils": "4.1.4",
+        "@vitest/pretty-format": "4.1.8",
+        "@vitest/utils": "4.1.8",
         "magic-string": "^0.30.21",
         "pathe": "^2.0.3"
       },
@@ -4218,7 +4160,9 @@
       }
     },
     "node_modules/@vitest/spy": {
-      "version": "4.1.4",
+      "version": "4.1.8",
+      "resolved": "https://registry.npmjs.org/@vitest/spy/-/spy-4.1.8.tgz",
+      "integrity": "sha512-6EevtBp6OZOPF7bmz36HrGMeP3txgVSrgebWxHOafDXGkhIzfXK14f8KF6MuFfgXXUeHxmpD3BQxkV00/3s5mA==",
       "dev": true,
       "license": "MIT",
       "funding": {
@@ -4226,11 +4170,13 @@
       }
     },
     "node_modules/@vitest/utils": {
-      "version": "4.1.4",
+      "version": "4.1.8",
+      "resolved": "https://registry.npmjs.org/@vitest/utils/-/utils-4.1.8.tgz",
+      "integrity": "sha512-uOJamYALNhfJ6iolExyQM40yIQwDqYnkKtQ5VCiSe17E33H0aQ/u+1GlRuz4LZBk6Mm3sg90G9hEbmEt37C1Zg==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
-        "@vitest/pretty-format": "4.1.4",
+        "@vitest/pretty-format": "4.1.8",
         "convert-source-map": "^2.0.0",
         "tinyrainbow": "^3.1.0"
       },
@@ -4319,6 +4265,8 @@
     },
     "node_modules/assertion-error": {
       "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/assertion-error/-/assertion-error-2.0.1.tgz",
+      "integrity": "sha512-Izi8RQcffqCeNVgFigKli1ssklIbpHnCYc6AknXGYoB6grJqyeby7jv12JUQgmTAnIDnbck1uxksT4dzN3PWBA==",
       "dev": true,
       "license": "MIT",
       "engines": {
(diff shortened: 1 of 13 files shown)


package-lock.json
```diff
@@ -8653,6 +8653,7 @@
         "@opentelemetry/sdk-metrics": "^2.6.1",
         "@opentelemetry/sdk-trace-base": "^2.6.1",
         "@opentelemetry/sdk-trace-node": "^2.6.1",
+        "@smithy/types": "^4.0.0",
```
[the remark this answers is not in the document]
notowen333: The `@smithy/types` entries pre-date this branch — they were added by the bedrock-mantle PR (#1066) that landed on main on 2026-05-14. Confirmed via `git log strands-ts/package.json` showing those lines unchanged in this PR's diff. Resolving as not-introduced-here.

agent-of-mkmeral (commented):
## Review — sharing learnings from a sibling plugin

Great work on this — the snapshot/restore mechanism is clean and the test discipline (validator-driven structural assertions, never on model output) is exactly right for non-deterministic LLM tests.

I'm leaving this as a `COMMENT` review (not `REQUEST_CHANGES`) because the existing bot review has the structural items covered. What I want to add is **field-tested learnings from a near-identical plugin we shipped recently** in a downstream agent — same `AfterInvocationEvent`-driven retry loop, same critic/judge concept, same failure shape. We hit several issues in production that are reproducible in this code; flagging them while the design is still mutable.

<details>
<summary><b>🔴 Empty/errored actor response is silently dropped — we shipped this exact bug, fixed last week</b></summary>

In `plugin.ts:259-261`:

```ts
// Cancelled or model-threw before emitting an assistant message.
const response = lastAssistantMessage(agent.messages)
if (!response) return
```

When the actor throws mid-invocation, errors out, or never emits an assistant message, the After hook bails — no retry, no `lastResult` updated, no `stopReason`. The whole goal loop silently disengages exactly when retry would be most valuable.

**Empirical data from our plugin:** across 45 runs in the 2.5 days after our equivalent shipped, **12 of 27 critic-eligible runs hit the silent-skip path and the critic issued zero RETRYs** — the highest-value failure mode never reached the validator. This is *especially* costly in Ralph mode, which is sold for long-running autonomous loops where transient failures are expected and recovery is the point.

**How we fixed it ([commit](https://github.com/agent-of-mkmeral/strands-coder-private/commit/1bf0d61)):** stop hardcoding the skip in plugin code; instead annotate the empty response (`[NO FINAL RESPONSE — actor produced no message text]` or `[ACTOR ERRORED OR TIMED OUT BEFORE COMPLETING]`) and pass it through to the judge/critic. The decision (RETRY vs PASS-on-intentional-no-op) lives in the judge's system prompt rather than plugin logic.

**Suggestion for #1082:**
- When `response` is missing, still increment `attempts`, push a `GoalAttempt` with `passed: false, feedback: '[ACTOR ERRORED OR PRODUCED NO MESSAGE]'`, and arm `event.resume` (subject to budgets).
- For NL-judge validators: detect this case before calling the judge (it can't grade an empty response anyway) and short-circuit to a structured fail-with-feedback.
- The `cross-invocation lastResult lifecycle` describe block at `plugin.test.ts:507-527` covers `lastResult` clearing after a thrown invoke — but doesn't cover "actor errored mid-loop, plugin should retry". That gap is what masked our bug for 2.5 days; consider adding it.

</details>

<details>
<summary><b>🟡 NL-judge sees text-only — tool calls invisible, makes systematically worse verdicts on tool-using agents</b></summary>

In `judge.ts:30-36`, `buildJudgePrompt` only serialises `textBlock` content from the transcript:

```ts
const text = m.content.flatMap((b) => (b.type === 'textBlock' ? [b.text] : [])).join('\n')
```

For agents using tools — which is the headline use case in the PR's `npm test`-driven Ralph example — the judge sees no `toolUse` / `toolResult` blocks. It can't tell whether the agent successfully ran a tool or hallucinated about it.

**What we learned:** our first version of the critic only saw the final response. Verdicts were noisy and missed concrete failures. We expanded context [in this commit](https://github.com/agent-of-mkmeral/strands-coder-private/commit/64b4fd9) to include a tool-call summary section, and verdict quality improved measurably. The current shape:

```
## Original Task
...
## Actor's Tool Calls
- shell: ls -la
  → success: total 24 ...
## Actor's Final Response
...
```

**Suggestion:** add a one-line summary of `toolUse`/`toolResult` blocks to `buildJudgePrompt`'s transcript serialisation. Otherwise the NL-judge path is anti-Ralph (model grading model on incomplete data) on the very use case the PR sells.

In the function-validator path this is moot (the user knows what they're checking) — only matters for the NL-judge.

</details>

<details>
<summary><b>🟡 Binary `passed: boolean` over-retries on unverifiable goals — we needed a third verdict</b></summary>

We started with `passed: boolean` too. We found that for genuinely conversational tasks ("explain X", research questions, design proposals), forcing the judge into PASS/RETRY led to noisy retries: the response was fine, the judge nitpicked, the loop burned `maxAttempts` chasing a non-issue.

We added a third verdict — `BYPASS` (task is not concretely verifiable) — encoded directly in the judge's system prompt. Three verdicts, two outcomes:
- `BYPASS` → terminate with `stopReason: 'unverifiable'`, signal to caller that the goal couldn't be meaningfully validated
- `PASS` → terminate, satisfied
- `RETRY` → resume with feedback

**For #1082 specifically**, the function-validator path is fine as binary (user knows their predicate). For the NL-judge path, two options:
1. Add an `abstain` outcome with its own `stopReason: 'unverifiable'`.
2. Document NL-judge mode as best for *measurable soft constraints* (length, style, format) and explicitly steer users to function validators for hard checks.

Our experience: option 1 is worth the small API surface cost. The signal to callers ("goal couldn't be validated" ≠ "goal failed") is genuinely useful and avoids the burn-budget trap.

</details>

<details>
<summary><b>🟢 Validator-error-as-failed-attempt masks programming bugs (already in my prior review on the issue side; reinforcing)</b></summary>

`plugin.ts:269-273` swallows validator throws as `{ passed: false, feedback: 'Validator error: ...' }`. If a user's validator has a `TypeError` (e.g. `response.foo.bar` where `foo` is undefined), every attempt throws identically, the loop burns `maxAttempts` retrying against a broken validator, and the user sees a `maxAttempts` stopReason rather than "your validator is broken".

Suggestion: at minimum, `logger.warn` on validator throw. Better: distinguish thrown validator errors from `{ passed: false }` outcomes and terminate with `stopReason: 'validatorError'` after N consecutive throws (e.g. `N=2`). Our actor-critic handles this via `try/except` around critic creation that logs and accepts on critic failure — symmetric pattern works here.

</details>

<details>
<summary><b>🟢 What this PR gets right that we got wrong — credit where due</b></summary>

Inverting the polarity: **the cross-invocation state lifecycle in this plugin is better than ours.**

The `BeforeInvocationEvent` reset of `_run` on every fresh invoke (and the `cross-invocation lastResult lifecycle` test block) is exactly the right pattern. Our `ActorCriticPlugin` initialises `_retry_count = 0` and `_start_time = None` in `__init__` and never resets — so re-using the plugin across multiple `agent()` calls in the same process carries retry counters over. We're filing a follow-up against ourselves to copy this pattern.

If you have a moment, the `_initialised` flag with explicit error on second-agent attach is also a nicer guardrail than what we have. Considering stealing that too.

</details>

---

### Summary

| Concern | Severity | Empirical basis |
|---|---|---|
| Empty/errored response silently dropped | 🔴 | 12/27 runs hit it in our equivalent over 2.5 days |
| NL-judge text-only context | 🟡 | Verdict quality measurably improved when we added tool-call summary |
| Binary verdict over-retries on unverifiable goals | 🟡 | We added BYPASS specifically to fix this |
| Validator-error-as-fail masks bugs | 🟢 | Conceptual; we hit similar pattern |

Only 🔴 is something I'd consider a merge blocker. The 🟡s are design conversations that are easier to have now than after customers depend on the current shape. None of this changes the verdict that the core machinery is sound.

Happy to elaborate or share specific code from our equivalent if useful — drop a comment or ping me on the issue thread.

strands-ts/src/vended-plugins/goal/plugin.ts
```diff
@@ -0,0 +1,413 @@
...
+}
+
+/**
+ * Validator must return `true`, `false`, or a `ValidationOutcome`. Booleans
+ * are shorthand: `true` → pass, `false` → fail with no feedback. Use the
+ * object form when you have actionable feedback for the next attempt.
+ */
+export type Validator =
+  | string
+  | ((response: Message) => boolean | ValidationOutcome | Promise<boolean | ValidationOutcome>)
```
mkmeral: should we also pass the agent instance or messages? we use the agent.messages in our own judge, right?

strands-ts/src/vended-plugins/goal/plugin.ts
```diff
@@ -0,0 +1,413 @@
...
+   * @defaultValue true
+   */
+  preserveContext?: boolean
+  /**
+   * Builds the user message fed to the agent before each retry. Receives the
+   * trimmed validator feedback (or `undefined` if the validator gave none).
+   * Override to localize the default English, retune the framing, or embed
+   * feedback in a domain-specific structure.
+   */
+  resumePromptTemplate?: (feedback: string | undefined) => string
```
mkmeral: do we want only string? can/should we allow other types? or can we do it later?

strands-ts/src/vended-plugins/goal/plugin.ts
```diff
@@ -0,0 +1,413 @@
...
+    if (this._initialised) {
+      throw new Error(`${this.name}: GoalLoop instances cannot be shared across agents; construct one per agent`)
+    }
+    // Two GoalLoops on the same agent both register an AfterInvocationEvent
+    // hook and both write `event.resume` — last writer wins, so one's feedback
+    // would be silently dropped. Compose constraints in a single validator
+    // function instead.
+    if (agentsWithGoalLoop.has(agent)) {
+      throw new Error(
+        `${this.name}: another GoalLoop is already attached to this agent; only one GoalLoop is supported per agent`
```
mkmeral, line 291: why one? 

also this limitation now makes me think if goal loop should be intervention 😅  bc the main value of interventions is solving this problem

strands-ts/src/vended-plugins/goal/plugin.ts
```diff
@@ -0,0 +1,413 @@
...
+export class GoalLoop implements Plugin {
+  readonly name: string
+
+  private readonly _validate: Validator
+  private readonly _evaluatorModel?: Model
+  private readonly _maxAttempts: number
+  private readonly _timeout: number
+  private readonly _preserveContext: boolean
+  private readonly _resumePromptTemplate: (feedback: string | undefined) => string
+  private _run: RunState | undefined = undefined
```
mkmeral: should we have a single run state? What happens if I pass my plugin into multiple agents?

strands-ts/src/vended-plugins/goal/plugin.ts
notowen333: Good point. We can just add `LocalAgent` for as much context as the implementer wants

strands-ts/src/vended-plugins/goal/plugin.ts
notowen333: From a previous race condition comment, we explicitly disabled one plugin being attached to many agents but I think we can roll back that assumption

strands-ts/src/vended-plugins/goal/plugin.ts
notowen333: See below comment. And I think we already aligned that based on state and mutative properties Plugin is a better fit

strands-ts/src/vended-plugins/goal/plugin.ts
notowen333: Sure we can widen to `| ContentBlock[]`

strands-ts/src/vended-plugins/goal/judge.ts
```diff
@@ -0,0 +1,71 @@
...
+import { z } from 'zod'
+import type { Message } from '../../types/messages.js'
+
+/**
+ * Structured outcome the judge agent fills via the `strands_structured_output`
+ * tool. Pass this to a custom judge `Agent` via `structuredOutputSchema` to
+ * mirror the shape `GoalLoop` expects from `validate`.
+ */
+export const JUDGE_OUTCOME_SCHEMA = z.object({
+  passed: z.boolean().describe('True iff the response fully satisfies the criteria.'),
```
mkmeral: nit: `iff` 

did you type this with your hands??  lol

strands-ts/src/vended-plugins/goal/plugin.ts
```diff
@@ -0,0 +1,415 @@
...
+    }
+    const goalDescription = validator
+    // The NL judge intentionally ignores the `response` argument — its prompt
+    // includes the full host transcript (via `buildJudgePrompt`) so the judge
+    // can evaluate against context, not just the last assistant turn.
+    return async () => {
+      const judge = new Agent({
+        model: this._evaluatorModel ?? hostAgent.model,
+        printer: false,
+        systemPrompt: JUDGE_SYSTEM_PROMPT,
```
mkmeral: I think system prompt should be customizable too. Maybe check how we do these things for summarizing conversation manageR?

strands-ts/src/vended-plugins/goal/judge.ts
```diff
@@ -0,0 +1,71 @@
...
+ * Tool calls and results are summarised inline so the judge can grade goals
+ * that depend on tool behaviour (e.g. "did the agent run the tests and act on
+ * the failures?"). Without this, a tool-using agent's transcript would look
+ * empty to the judge whenever the model's text output was sparse.
+ *
+ * @param description - Natural-language goal the judge evaluates against.
+ * @param transcript - Working agent's conversation messages.
+ * @returns Composed prompt string ready to feed to a judge `Agent.invoke`.
+ */
+export function buildJudgePrompt(description: string, transcript: readonly Message[]): string {
```
mkmeral, line 73: nit: maybe say it's for building judge input prompt? I thought at first we were adding this to system prompt

strands-ts/src/vended-plugins/goal/judge.ts
notowen333: It's short hand for `if and only if` I'll expand it

strands-ts/src/vended-plugins/goal/plugin.ts
notowen333: Makes sense will add an interface for this

strands-ts/src/vended-plugins/goal/plugin.ts
```diff
@@ -0,0 +1,453 @@
...
+export interface GoalLoopOptions {
+  /**
+   * Natural-language goal. An internal judge Agent grades each attempt against
+   * it and returns feedback on failure. Mutually exclusive with `validator`.
+   */
+  goal?: string
+  /** Programmatic validator predicate. Mutually exclusive with `goal`. */
+  validator?: Validator
+  /** Tuning for the auto-built judge used when `goal` is set. */
+  judge?: JudgeConfig
```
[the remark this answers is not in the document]
zastrowm: `judgeSystemPrompt` can make sense; why not make it just `model` though?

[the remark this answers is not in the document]
zastrowm: I presume we already decided that passing an agent directly in makes no sense?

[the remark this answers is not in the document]
notowen333: Yes we decided not to pass an Agent because we construct a fresh agent per turn

mkmeral (approved):

PR #1084, opened 2026-05-18
feat(steering): add vended-interventions/steering module

notowen333:
## Description

Adds a steering module under `vended-interventions/steering`. Steering provides
just-in-time contextual guidance — handlers gate or redirect agent actions
based on accumulated context, instead of front-loading all instructions in the
system prompt.

The module is built on the **interventions primitive**, not plugins.
`SteeringHandler` extends `InterventionHandler` and is attached via
`interventions:` on the agent. `evaluateToolCall` and `evaluateModelOutput` map
onto intervention actions (`Proceed | Guide | Confirm` for tool calls,
`Proceed | Guide` for model output).

Context comes from `SteeringContextProvider` instances. The included
`ToolLedgerProvider` records tool history (args, status, timing) by
self-subscribing to `BeforeToolCallEvent`/`AfterToolCallEvent`. Handler-level
configurable `name` lets multiple steering handlers coexist on one agent
without colliding in `InterventionRegistry`.

To support stateful sub-systems like context providers, this PR adds a new
`LifecycleObserver` interface in `src/types`. Implementors expose
`observeAgent(agent)` — given the agent, they self-subscribe to any
`HookableEvent` via `agent.addHook`. `InterventionRegistry` invokes
`observeAgent` on any handler that also implements the interface.
`SteeringHandler` implements `LifecycleObserver` and forwards to its providers;
`SteeringContextProvider` extends `LifecycleObserver`, so providers like
`ToolLedgerProvider` self-register their hooks. The contract is deliberately
narrower than `Plugin.initAgent`: implementors get the `LocalAgent` surface
scoped to `addHook`-style observation, not the broader plugin power to
register tools or mutate config.

`LLMSteeringHandler` constructs a fresh inner `Agent` per evaluation rather
than reusing a single inner agent — this keeps the handler stateless across
calls and avoids re-entrancy on the inner agent's invocation lock.

### Example

```typescript
import { Agent } from '@strands-agents/sdk'
import { LLMSteeringHandler } from '@strands-agents/sdk/vended-interventions/steering'
import { BedrockModel } from '@strands-agents/sdk/models/bedrock'

const handler = new LLMSteeringHandler({
  model: new BedrockModel(),
  systemPrompt: 'Block tool calls that retry the same failed action more than twice.',
})

const agent = new Agent({
  model: new BedrockModel(),
  tools: [searchWeb, sendEmail],
  interventions: [handler],
})

await agent.invoke('Find the CEO of Acme and email them.')
```

Custom handlers subclass `SteeringHandler` directly:

```typescript
class RedirectEmailHandler extends SteeringHandler {
  override readonly name = 'redirect-email'

  override async evaluateToolCall(event) {
    if (event.toolUse.name === 'send_email') {
      return guide('Use send_notification instead.')
    }
    return proceed()
  }
}
```

### Divergence from Python

- TS reuses the intervention action types (`Proceed | Guide | Confirm`) instead
  of defining steering-specific action classes. Mapping happens in the
  registry, not the handler.
- TS bundles `ToolLedgerProvider` as a single class that self-subscribes to
  both tool-call hooks; Python decomposes into `LedgerBeforeToolCall` /
  `LedgerAfterToolCall` / `LedgerProvider`.
- TS default prompt is minimal and uses a `CachePointBlock`; Python uses a
  longer Agent-SOP-formatted template.

## Documentation PR

will follow up

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

Unit tests cover handler routing, context propagation, name configurability,
ledger lifecycle (pending → success/error, maxEntries cap), confirm
approve/deny, and the new `LifecycleObserver` contract (handler implementing
`observeAgent` receives the agent at construction and can self-register hooks).
Integ tests exercise the full agent loop with Bedrock (`describe.skipIf(bedrock.skip)`).

## Checklist

----

### DIFF
--- a/strands-ts/package.json
+++ b/strands-ts/package.json
@@ -83,6 +83,10 @@
     "./vended-interventions/hitl": {
       "types": "./dist/src/vended-interventions/hitl/index.d.ts",
       "default": "./dist/src/vended-interventions/hitl/index.js"
+    },
+    "./vended-interventions/steering": {
+      "types": "./dist/src/vended-interventions/steering/index.d.ts",
+      "default": "./dist/src/vended-interventions/steering/index.js"
     }
   },
   "scripts": {
@@ -127,13 +131,12 @@
     "@ai-sdk/openai": "^3.0.41",
     "@ai-sdk/provider": "^3.0.0",
     "@anthropic-ai/sdk": "^0.92.0",
-    "@aws/bedrock-token-generator": "^1.1.0",
     "@aws-sdk/client-bedrock": "^3.943.0",
     "@aws-sdk/client-s3": "^3.943.0",
     "@aws-sdk/client-secrets-manager": "^3.943.0",
     "@aws-sdk/client-sts": "^3.996.0",
     "@aws-sdk/credential-providers": "^3.943.0",
-    "@smithy/types": "^4.0.0",
+    "@aws/bedrock-token-generator": "^1.1.0",
     "@eslint/js": "^9.39.4",
     "@google/genai": "^1.40.0",
     "@opentelemetry/api": "^1.9.0",
@@ -143,6 +146,7 @@
     "@opentelemetry/sdk-metrics": "^2.6.1",
     "@opentelemetry/sdk-trace-base": "^2.6.1",
     "@opentelemetry/sdk-trace-node": "^2.6.1",
+    "@smithy/types": "^4.0.0",
     "@types/express": "^5.0.6",
     "@types/node": "^25.6.0",
     "@types/uuid": "^11.0.0",
@@ -174,6 +178,7 @@
   "dependencies": {
     "@aws-sdk/client-bedrock-runtime": "^3.1037.0",
     "@types/json-schema": "^7.0.15",
+    "dedent": "^1.7.2",
     "uuid": "^14.0.0",
     "yaml": "^2.8.3"
   },
@@ -181,9 +186,8 @@
     "@a2a-js/sdk": "^0.3.10",
     "@ai-sdk/provider": "^3.0.0",
     "@anthropic-ai/sdk": "^0.92.0",
-    "@aws/bedrock-token-generator": "^1.1.0",
     "@aws-sdk/client-s3": "^3.943.0",
-    "@smithy/types": "^4.0.0",
+    "@aws/bedrock-token-generator": "^1.1.0",
     "@google/genai": "^1.40.0",
     "@modelcontextprotocol/sdk": "^1.25.2",
     "@opentelemetry/api": "^1.9.0",
@@ -193,6 +197,7 @@
     "@opentelemetry/sdk-metrics": "^2.6.1",
     "@opentelemetry/sdk-trace-base": "^2.6.1",
     "@opentelemetry/sdk-trace-node": "^2.6.1",
+    "@smithy/types": "^4.0.0",
     "express": "^5.1.0",
     "openai": "^6.7.0",
     "zod": "^4.1.12"
--- a/strands-ts/package.json
+++ b/strands-ts/package.json
@@ -178,7 +178,6 @@
   "dependencies": {
     "@aws-sdk/client-bedrock-runtime": "^3.1037.0",
     "@types/json-schema": "^7.0.15",
-    "dedent": "^1.7.2",
     "uuid": "^14.0.0",
     "yaml": "^2.8.3"
   },
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -349,7 +349,7 @@ export class Agent implements LocalAgent, InvokableAgent {
     // Initialize hooks registry
     this._hooksRegistry = new HookRegistryImplementation()
 
-    this._interventionRegistry = new InterventionRegistry(config?.interventions ?? [], this._hooksRegistry)
+    this._interventionRegistry = new InterventionRegistry(config?.interventions ?? [], this)
 
     // `undefined` (omitted) → install the default; `null`/`[]` → explicit opt-out.
     const retryStrategies: RetryStrategy[] =
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -38,7 +38,7 @@ import { AgentPrinter, getDefaultAppender, type Printer } from './printer.js'
 import type { Plugin } from '../plugins/plugin.js'
 import type { InterventionHandler } from '../interventions/handler.js'
 import { InterventionRegistry } from '../interventions/registry.js'
-import { isLifecycleObserver } from '../types/lifecycle-observer.js'
+import type { LifecycleObserver } from '../types/lifecycle-observer.js'
 import { PluginRegistry } from '../plugins/registry.js'
 import { SlidingWindowConversationManager } from '../conversation-manager/sliding-window-conversation-manager.js'
 import { NullConversationManager } from '../conversation-manager/null-conversation-manager.js'
@@ -459,8 +459,9 @@ export class Agent implements LocalAgent, InvokableAgent {
     await this._pluginRegistry.initialize(this)
 
     for (const handler of this._interventionRegistry.handlers) {
-      if (isLifecycleObserver(handler)) {
-        await handler.observeAgent(this)
+      const observer = handler as Partial<LifecycleObserver>
+      if (typeof observer.observeAgent === 'function') {
+        await observer.observeAgent(this)
       }
     }
 
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -455,8 +455,6 @@ export class Agent implements LocalAgent, InvokableAgent {
       })
     )
 
-    await this._interventionRegistry.initialize(this)
-
     await this._pluginRegistry.initialize(this)
 
     await this._hooksRegistry.invokeCallbacks(new InitializedEvent({ agent: this }))
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -349,7 +349,7 @@ export class Agent implements LocalAgent, InvokableAgent {
     // Initialize hooks registry
     this._hooksRegistry = new HookRegistryImplementation()
 
-    this._interventionRegistry = new InterventionRegistry(config?.interventions ?? [], this)
+    this._interventionRegistry = new InterventionRegistry(config?.interventions ?? [], this._hooksRegistry)
 
     // `undefined` (omitted) → install the default; `null`/`[]` → explicit opt-out.
     const retryStrategies: RetryStrategy[] =
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -38,6 +38,7 @@ import { AgentPrinter, getDefaultAppender, type Printer } from './printer.js'
 import type { Plugin } from '../plugins/plugin.js'
 import type { InterventionHandler } from '../interventions/handler.js'
 import { InterventionRegistry } from '../interventions/registry.js'
+import { isLifecycleObserver } from '../types/lifecycle-observer.js'
 import { PluginRegistry } from '../plugins/registry.js'
 import { SlidingWindowConversationManager } from '../conversation-manager/sliding-window-conversation-manager.js'
 import { NullConversationManager } from '../conversation-manager/null-conversation-manager.js'
@@ -456,6 +457,12 @@ export class Agent implements LocalAgent, InvokableAgent {
     )
 
     await this._pluginRegistry.initialize(this)
+
+    for (const handler of this._interventionRegistry.handlers) {
+      if (isLifecycleObserver(handler)) {
+        await handler.observeAgent(this)
+      }
+    }
 
     await this._hooksRegistry.invokeCallbacks(new InitializedEvent({ agent: this }))
 
--- a/strands-ts/src/agent/agent.ts
+++ b/strands-ts/src/agent/agent.ts
@@ -455,6 +455,8 @@ export class Agent implements LocalAgent, InvokableAgent {
       })
     )
 
+    await this._interventionRegistry.initialize(this)
+
     await this._pluginRegistry.initialize(this)
 
     await this._hooksRegistry.invokeCallbacks(new InitializedEvent({ agent: this }))
--- a/strands-ts/src/interventions/handler.ts
+++ b/strands-ts/src/interventions/handler.ts
@@ -7,7 +7,7 @@ import type {
 } from '../hooks/events.js'
 import type { Proceed, Deny, Guide, Confirm, Transform } from './actions.js'
 
-type Awaitable<T> = T | Promise<T>
+export type Awaitable<T> = T | Promise<T>
 
 /**
  * What to do when a handler throws during evaluation.
--- a/strands-ts/src/interventions/handler.ts
+++ b/strands-ts/src/interventions/handler.ts
@@ -5,7 +5,6 @@ import type {
   BeforeModelCallEvent,
   AfterModelCallEvent,
 } from '../hooks/events.js'
-import type { LocalAgent } from '../types/agent.js'
 import type { Proceed, Deny, Guide, Confirm, Transform } from './actions.js'
 
 type Awaitable<T> = T | Promise<T>
@@ -45,14 +44,6 @@ export abstract class InterventionHandler {
 
   /** What to do when this handler throws. Defaults to 'throw'. */
   readonly onError: OnError = 'throw'
-
-  /**
-   * Optional one-time setup called when the agent initializes. Use this to
-   * register additional hooks or stateful sub-systems the handler depends on
-   * (for example, context providers that observe events between intervention
-   * calls). Default is a no-op.
-   */
-  initAgent(_agent: LocalAgent): void | Promise<void> {}
 
   beforeInvocation(_event: BeforeInvocationEvent): Awaitable<Proceed | Deny | Guide | Transform> {
     return { type: 'proceed' }
--- a/strands-ts/src/interventions/handler.ts
+++ b/strands-ts/src/interventions/handler.ts
@@ -5,6 +5,7 @@ import type {
   BeforeModelCallEvent,
   AfterModelCallEvent,
 } from '../hooks/events.js'
+import type { LocalAgent } from '../types/agent.js'
 import type { Proceed, Deny, Guide, Confirm, Transform } from './actions.js'
 
 type Awaitable<T> = T | Promise<T>
@@ -44,6 +45,14 @@ export abstract class InterventionHandler {
 
   /** What to do when this handler throws. Defaults to 'throw'. */
   readonly onError: OnError = 'throw'
+
+  /**
+   * Optional one-time setup called when the agent initializes. Use this to
+   * register additional hooks or stateful sub-systems the handler depends on
+   * (for example, context providers that observe events between intervention
+   * calls). Default is a no-op.
+   */
+  initAgent(_agent: LocalAgent): void | Promise<void> {}
 
   beforeInvocation(_event: BeforeInvocationEvent): Awaitable<Proceed | Deny | Guide | Transform> {
     return { type: 'proceed' }
--- a/strands-ts/src/interventions/handler.ts
+++ b/strands-ts/src/interventions/handler.ts
@@ -5,6 +5,7 @@ import type {
   BeforeModelCallEvent,
   AfterModelCallEvent,
 } from '../hooks/events.js'
+import type { LocalAgent } from '../types/agent.js'
 import type { Proceed, Deny, Guide, Confirm, Transform } from './actions.js'
 
 type Awaitable<T> = T | Promise<T>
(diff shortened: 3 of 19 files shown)


strands-ts/src/vended-interventions/steering/providers/tool-ledger.ts
```diff
@@ -0,0 +1,137 @@
...
+      throw new Error(
+        `ToolLedgerProvider is already attached to a different Agent. Construct a new provider per Agent instead of sharing.`
+      )
+    }
+    this._attachedAgent = agent
+
+    // Rehydrate from appState if a previous session was restored
+    const saved = agent.appState.get(this.name)
+    if (Array.isArray(saved)) {
+      this._toolCalls.push(...(saved as unknown as LedgerToolCall[]))
```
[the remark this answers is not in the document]
notowen333: Keeping the cast — appState here is round-tripped state we wrote ourselves, not external input. We don't validate other appState reads either, and treating this one specially would be inconsistent. If a corrupted entry slips through, downstream code (status coercion in AfterToolCall) is forgiving.

strands-ts/package.json
```diff
@@ -170,16 +174,16 @@
   "dependencies": {
     "@aws-sdk/client-bedrock-runtime": "^3.1037.0",
     "@types/json-schema": "^7.0.15",
+    "dedent": "^1.7.2",
```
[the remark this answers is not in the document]
notowen333: 28KB is acceptable and dedent is a generally useful prompt-building lib we'll likely reuse.

strands-ts/src/vended-interventions/steering/providers/context-provider.ts
```diff
@@ -0,0 +1,53 @@
...
+ *     })
+ *   }
+ *
+ *   get context(): SteeringContextData {
+ *     return { type: 'costTracker', totalTokens: this._totalTokens }
+ *   }
+ * }
+ * ```
+ */
+export interface SteeringContextProvider extends Pick<Plugin, 'initAgent' | 'name'> {
```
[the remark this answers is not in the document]
notowen333: Keeping the Pick — it's idiomatic reuse of two stable Plugin field signatures (name, initAgent), and Plugin.initAgent's shape is load-bearing across the SDK so churn there is unlikely without a coordinated migration.

strands-ts/src/vended-interventions/steering/handlers/llm.ts
```diff
@@ -0,0 +1,181 @@
...
+      case 'interrupt':
+        return { type: 'confirm', prompt: decision.reason } satisfies Confirm
+    }
+  }
+
+  // Constructs a fresh inner agent per call so the handler has no shared
+  // mutable state between invocations — this keeps it safe to attach to
+  // multiple parent agents (whose tool calls may evaluate concurrently).
+  private async _invoke(prompt: string | ContentBlock[]): Promise<SteeringDecision> {
+    const inner = new Agent({
```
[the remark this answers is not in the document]
notowen333: Inner Agent has no plugins/interventions/tools, so initialize() is cheap relative to the LLM round-trip that dominates the call. Pooling is a fine subclass override if anyone hits the cost.

strands-ts/src/interventions/handler.ts
```diff
@@ -45,6 +46,14 @@ export abstract class InterventionHandler {
   /** What to do when this handler throws. Defaults to 'throw'. */
   readonly onError: OnError = 'throw'
 
+  /**
+   * Optional one-time setup called when the agent initializes. Use this to
+   * register additional hooks or stateful sub-systems the handler depends on
+   * (for example, context providers that observe events between intervention
+   * calls). Default is a no-op.
+   */
+  initAgent(_agent: LocalAgent): void | Promise<void> {}
```
lizradway: I don't think we should be adding this to the Interventions API surface unless we _really_ have to. @mkmeral can speak more to this, but we do not want to overlap agent state with plugins as this does.

Instead of adding `initAgent` to the `InterventionHandler` base class, `SteeringHandler` can derive its context providers directly from its own intervention hook method overrides (`beforeToolCall`, `afterToolCall`). We can treat the hook providers as passive observers of the context accumulation.


This keeps the change local to the steering module without modifications to the base class, registry, or `agent.ts`. If a provider later needs a new event, you just add another override on `SteeringHandler`.

strands-ts/package.json
lizradway: do we really need to add this as a concrete runtime dependency?

strands-ts/src/vended-interventions/steering/providers/context-provider.ts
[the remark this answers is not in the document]
lizradway: We should not be linking this to plugins. Ideally we get rid of AgentInit and then we can implement the intervention name field

strands-ts/src/interventions/registry.ts
```diff
@@ -42,6 +43,16 @@ export class InterventionRegistry {
     this._registerHooks(hookRegistry)
   }
 
+  /**
+   * Run each handler's optional `initAgent` hook. Called once by the agent
+   * during {@link Agent.initialize}.
+   */
+  async initialize(agent: LocalAgent): Promise<void> {
```
lizradway: if we remove `initAgent`, we can remove this and any changes to `agent.ts`.

strands-ts/src/vended-interventions/steering/handlers/handler.ts
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * Subclass {@link SteeringHandler} and override {@link steerBeforeTool} and/or
+ * {@link steerAfterModel}. Both methods have default implementations that return Proceed.
+ *
+ * @example
+ * ```typescript
+ * class MySteeringHandler extends SteeringHandler {
+ *   override async steerBeforeTool(agent, toolUse) {
+ *     if (toolUse.name === 'dangerous_tool') {
+ *       return { type: 'guide', feedback: 'This tool requires extra caution.' }
```
lizradway: we should be using the intervention actions functions: https://github.com/strands-agents/sdk-typescript/blob/61db42875eaa44f9c91790b030f0813e62af5010/strands-ts/src/interventions/actions.ts#L158-L208

strands-ts/src/vended-interventions/steering/handlers/handler.ts
```diff
@@ -0,0 +1,115 @@
...
+import { InterventionHandler } from '../../../interventions/handler.js'
+import type { Confirm, Guide, Proceed } from '../../../interventions/actions.js'
+import { logger } from '../../../logging/logger.js'
+import type { LocalAgent } from '../../../types/agent.js'
+import type { Message, StopReason } from '../../../types/messages.js'
+import type { ToolUse } from '../../../tools/types.js'
+import type { SteeringContextData, SteeringContextProvider } from '../providers/context-provider.js'
+
+/** Steering decisions valid before tool execution. */
+export type ToolSteeringAction = Proceed | Guide | Confirm
```
lizradway: not necessary imo.


in interventions we pushed towards keeping the union explicit inline to keep return types visible. also avoids a barrel export of a bunch of randomly assorted defined actions for each intervention handler when there are only 5 in total

strands-ts/src/vended-interventions/steering/handlers/handler.ts
```diff
@@ -0,0 +1,115 @@
...
+ * {@link InterventionHandler} — pass it via `interventions:` on the agent.
+ */
+export abstract class SteeringHandler extends InterventionHandler {
+  override readonly name: string
+
+  private readonly _contextProviders: SteeringContextProvider[]
+
+  constructor(config?: SteeringHandlerConfig) {
+    super()
+    this.name = config?.name ?? 'strands:steering'
```
lizradway: what if we have multiple steering handlers?

strands-ts/src/vended-interventions/steering/handlers/handler.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+    return this.steerAfterModel(event.agent, event.stopData.message, event.stopData.stopReason)
+  }
+
+  /**
+   * Provide contextual guidance before tool execution. Override to customize.
+   * Default implementation proceeds.
+   */
+  async steerBeforeTool(_agent: LocalAgent, _toolUse: ToolUse): Promise<ToolSteeringAction> {
+    return { type: 'proceed' }
```
lizradway: +1 to all of these for using the [prebuilt action functions](https://github.com/strands-agents/sdk-typescript/blob/61db42875eaa44f9c91790b030f0813e62af5010/strands-ts/src/interventions/actions.ts#L158-L208)

strands-ts/src/vended-interventions/steering/handlers/llm.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  override async steerBeforeTool(_agent: LocalAgent, toolUse: ToolUse): Promise<ToolSteeringAction> {
+    const context = this.getSteeringContext()
+    const prompt = this._promptBuilder(context, toolUse)
+    const decision = await this._invoke(prompt)
+
+    switch (decision.type) {
+      case 'proceed':
+        return { type: 'proceed', reason: decision.reason } satisfies Proceed
+      case 'guide':
```
lizradway, line 219: use the existing action types, data, structures, and functions :D

https://github.com/strands-agents/sdk-typescript/blob/61db42875eaa44f9c91790b030f0813e62af5010/strands-ts/src/interventions/actions.ts

strands-ts/src/vended-interventions/steering/handlers/handler.ts
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * Collect context from all registered providers. Subclasses (and tests)
+   * may call this to inspect the accumulated provider snapshots.
+   */
+  getSteeringContext(): SteeringContextData[] {
+    return this._contextProviders.map((provider) => provider.context)
+  }
+
+  override async beforeToolCall(event: BeforeToolCallEvent): Promise<ToolSteeringAction> {
+    return this.steerBeforeTool(event.agent, event.toolUse)
```
lizradway: i dont think this (or `steerAfterModel`) is necessary, i would just override `beforeToolCall`.

feels like unnecessary indirection for a prettier function name that could just be addressed via comment.

strands-ts/src/vended-interventions/steering/handlers/llm.ts
```diff
@@ -0,0 +1,182 @@
...
+// ---------------------------------------------------------------------------
+// LLM steering handler
+// ---------------------------------------------------------------------------
+
+/**
+ * Configuration for the LLMSteeringHandler.
+ */
+export interface LLMSteeringHandlerConfig {
+  /** Model for steering evaluation. */
+  model: Model
```
lizradway: a little surprised at no fallback?

strands-ts/src/vended-interventions/steering/handlers/llm.ts
```diff
@@ -0,0 +1,182 @@
...
+   * Identifier for this handler instance. Defaults to `'strands:steering'`.
+   * Override when attaching multiple steering handlers to the same agent.
+   */
+  name?: string
+}
+
+/** Schema returned by the steering LLM. */
+const STEERING_DECISION = z.object({
+  type: z
+    .enum(['proceed', 'guide', 'interrupt'])
```
lizradway: `interrupt` is `confirm`

strands-ts/src/vended-interventions/steering/handlers/llm.ts
```diff
@@ -0,0 +1,182 @@
...
+  reason: z.string().describe('Clear explanation of the steering decision and any guidance provided'),
+})
+
+type SteeringDecision = z.infer<typeof STEERING_DECISION>
+
+/**
+ * Steering handler that uses an LLM to provide contextual guidance.
+ *
+ * Uses natural language prompts to evaluate tool calls and produce an
+ * intervention action. `interrupt` decisions become a {@link Confirm} so the
```
lizradway: huh? why are we doing this? confused

strands-ts/src/vended-interventions/steering/providers/tool-ledger.ts
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Configuration for {@link ToolLedgerProvider}.
+ */
+export interface ToolLedgerProviderConfig {
+  /** Maximum number of tool calls to retain. Older entries are dropped. Defaults to 100. */
+  maxEntries?: number
+  /**
+   * Identifier used as the {@link Agent.appState} key for persisted ledger
+   * data. Defaults to `'strands:steering:toolLedger'`. Override when
+   * attaching multiple ledger providers to the same agent so each persists
```
lizradway: i dont like that a user has to see a comment to know this. is there any way to avoid this? (same with base class steering)

strands-ts/src/vended-interventions/steering/providers/tool-ledger.ts
```diff
@@ -0,0 +1,137 @@
...
+  /** Unique tool use identifier. */
+  id: string
+  /** Tool name. */
+  name: string
+  /** Tool execution result. */
+  result?: JSONValue
+  /** When the tool call was initiated. */
+  startTime: string
+  /** Current execution state: pending, success, or error. */
+  status: 'pending' | 'success' | 'error'
```
lizradway: wonder if its worth reusing [`ToolResultStatus`](https://github.com/strands-agents/sdk-typescript/blob/61db42875eaa44f9c91790b030f0813e62af5010/strands-ts/src/tools/types.ts#L7)

`status: 'pending' | ToolResultStatus`

strands-ts/src/vended-interventions/steering/providers/tool-ledger.ts
```diff
@@ -0,0 +1,137 @@
...
+      throw new Error(
+        `ToolLedgerProvider is already attached to a different Agent. Construct a new provider per Agent instead of sharing.`
+      )
+    }
+    this._attachedAgent = agent
+
+    // Rehydrate from appState if a previous session was restored
+    const saved = agent.appState.get(this.name)
+    if (Array.isArray(saved)) {
+      this._toolCalls.push(...(saved as unknown as LedgerToolCall[]))
```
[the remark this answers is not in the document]
lizradway: in removing init agent resaturation logic can move into the intervention hook overrides + no more concern of reattaching.

strands-ts/src/vended-interventions/steering/providers/tool-ledger.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    agent.addHook(BeforeToolCallEvent, (event) => {
```
lizradway: why aren't these in the existing intervention hook overrides?

strands-ts/src/interventions/handler.ts
notowen333: Noted on the desire to avoid adding `initAgent`

What we get from the current design is simplicity/flexibility in the `SteeringHandler`. `SteeringHandler` doesn't keep track of which events are providing context.

> If a provider later needs a new event, you just add another override on SteeringHandler.

Right now `SteeringContextProvider` doesn't determine which events can be defined. So, we'd be limiting the design to only pre-prescribed extension points.

strands-ts/src/vended-interventions/steering/handlers/handler.ts
notowen333: I think these are organized and help with readability in the steering context:

```
/** Steering decisions valid before tool execution. */
export type ToolSteeringAction = Proceed | Guide | Confirm

/** Steering decisions valid after a model response. */
export type ModelSteeringAction = Proceed | Guide
```

but can remove the types

strands-ts/src/vended-interventions/steering/handlers/handler.ts
notowen333: You can support multiple handlers by passing a `name` in the config.

As a side note, there are other constructs that don't expose name which don't have a path to multiple instances

strands-ts/src/vended-interventions/steering/handlers/handler.ts
notowen333: sure I can inline the implementation

strands-ts/src/vended-interventions/steering/handlers/llm.ts
notowen333: Good point I can default it to `Bedrock` like we do elsewhere

strands-ts/src/interventions/handler.ts
lizradway: yeah i totally get the ease, but given that this can be done without `initAgent`,  I do not think we should be linking agent subclasses to agent objects and modifying the interventions api surface unless we have no other choice.


We can just add new lifecycle events to interventions if we want to add new context events, we're always free to add more/as many lifecycle events to interventions as we want.

strands-ts/src/vended-interventions/steering/handlers/handler.ts
lizradway: gotcha

strands-ts/src/vended-interventions/steering/handlers/llm.ts
notowen333: Can do it as long as we are OK with the python divergence

strands-ts/src/vended-interventions/steering/providers/tool-ledger.ts
notowen333: As noted above, this is an issue throughout the SDK. Maybe we can consider adding an additional property or salt to these names?

But I think for now the pattern in this PR is standard.

strands-ts/src/vended-interventions/steering/providers/tool-ledger.ts
notowen333: This will be contingent on the discussion on how if/how Providers will work

strands-ts/src/interventions/handler.ts
notowen333: > We can just add new lifecycle events to interventions if we want to add new context events, we're always free to add more/as many lifecycle events to interventions as we want.

The tension here is that the `ContextProvider` plugin design includes /all/ events. So we'd need to add everything to cover the same scope and then we're re-inventing plugins.

For something like the `ToolLedgerProvider` it's not an issue because the scope is well defined, so transitioning to the proposed approach would be simple.

strands-ts/src/vended-interventions/steering/providers/tool-ledger.ts
lizradway: yeah if its standard thats fine

strands-ts/src/interventions/handler.ts
lizradway: for this existing pr, the passive approach has all events to work out of the box.

 if a future provider implementation needs a new life cycle event that can/should be gated, we'd just add that as a new lifecycle method on the intervention primitive itself so that any future intervention handlers can benefit from its protection. 

the scope of a steering context provider _should_ be the intervention lifecycle events.

notowen333: I synced offline with @pgrayy who authored the context provider and @mkmeral this morning. Everyone aligned with your proposed removal of the plugins.

While parting with complete configurability of the ContextProvider as a Plugin, I'll do an audit of events to see if there's anything else we want to add to interventions hooks registry.

strands-ts/src/interventions/registry.ts
```diff
@@ -39,32 +40,37 @@ export class InterventionRegistry {
       seen.add(h.name)
     }
     this._handlers = handlers
-    this._registerHooks(hookRegistry)
+    this._registerHooks(agent)
+    for (const handler of handlers) {
```
lizradway: should this registration be in `agent.ts` instead?

strands-ts/src/vended-interventions/steering/handlers/handler.ts
```diff
@@ -0,0 +1,111 @@
...
+    for (const provider of this._contextProviders) {
+      await provider.observeAgent(agent)
+    }
+  }
+
+  // ---------------------------------------------------------------------------
+  // Subclass extension points — narrow steering contract.
+  // ---------------------------------------------------------------------------
+
+  /** Evaluate a pending tool call. Default: proceed. */
```
lizradway: I really think we should remove these and just have the subclasses override `beforeToolCall`, etc

strands-ts/src/types/lifecycle-observer.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export interface LifecycleObserver {
+  /** Stable identifier for this observer. Used for logging and duplicate detection. */
+  readonly name: string
+
+  /**
+   * Called once when the observer is registered with an agent. Implementations
+   * typically subscribe to one or more events via `agent.addHook`.
+   */
+  observeAgent(agent: LocalAgent): void | Promise<void>
```
zastrowm: In anticipation of adding `observeMultiAgent` (or whatever) should this be an optional method?

strands-ts/src/agent/agent.ts
zastrowm, line 1:
```
const handler = new LLMSteeringHandler({
  model: new BedrockModel(),
  systemPrompt: 'Block tool calls that retry the same failed action more than twice.',
})
```

Should `model` be optional and allow using the `model` of the agent being used?

strands-ts/src/interventions/index.ts
```diff
@@ -2,4 +2,4 @@ export type { InterventionAction, LifecycleEvent, Proceed, Deny, Guide, Confirm,
 import { proceed, deny, guide, confirm, transform } from './actions.js'
 export const InterventionActions = { proceed, deny, guide, confirm, transform }
 export { InterventionHandler } from './handler.js'
-export type { OnError } from './handler.js'
+export type { OnError, Awaitable } from './handler.js'
```
lizradway: can we remove this from the barrel export but keep it exported internally?

lizradway (approved):

PR #1086, opened 2026-05-19
feat: support deriving tools from existing tools via inputSchema

pgrayy:
## Description

Users building agents with tools like `httpRequest` often need to add restrictions (URL blocklists, auth headers, rate limiting) without rebuilding the tool from scratch. Today, reusing another tool's input schema requires manually copying `toolSpec.inputSchema` (losing type information) or re-declaring the Zod schema.

This PR adds a new overload to `tool()` that accepts an existing tool as the `inputSchema` field. TypeScript infers the callback's input type from the source tool's generic parameters, giving users fully typed wrappers with zero casts.

```typescript
import { tool } from '@strands-agents/sdk'
import { httpRequest } from '@strands-agents/sdk/vended-tools/http-request'

const safeHttp = tool({
  name: 'safe_http_request',
  description: 'HTTP requests with domain restrictions',
  inputSchema: httpRequest,
  callback: async (input, context) => {
    if (new URL(input.url).hostname === 'blocked.com') {
      throw new Error('Domain blocked')
    }
    return httpRequest.invoke(input, context)
  },
})
```

When the source tool is a ZodTool, the derived tool preserves Zod runtime validation. When the source is a FunctionTool or McpTool, the JSON schema is inherited and a FunctionTool is created.

This also exposes `inputSchema` as a public readonly property on `ZodTool` (previously private as `_inputSchema`), enabling the factory to access the Zod schema for runtime validation preservation.

## Alternatives Considered

We explored adding a public `inputSchema` property to every tool type (`ZodTool`, `FunctionTool`, `McpTool`) and surfacing it via an intersection on the `tool()` return type (`InvokableTool<...> & { readonly inputSchema: TSchema }`). This would let users do `tool({ inputSchema: existingTool.inputSchema, ... })` directly.

The approach worked but had downsides:

- Adding `inputSchema` to the `Tool` abstract class is a breaking change (all subclasses must implement it).
- Without adding it to `Tool`, the property is invisible when a tool is passed through a function typed as `InvokableTool`. The intersection only exists at the factory call site and gets erased when assigned to the interface type.
- For `FunctionTool` and `McpTool`, the exposed schema is `JSONSchema` (opaque to TypeScript), so passing it to `tool()` still results in `unknown` input. Only ZodTool benefits from typed schema exposure.

The chosen approach (passing the tool itself as `inputSchema`) avoids these issues. Types are inferred from `InvokableTool<TInput, TReturn>`'s generic parameters rather than from a schema property, so it works uniformly regardless of the underlying tool type or how the tool reference is passed around.

## Related Issues

N/A

## Documentation PR

Will be created if this proposal is accepted.

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`
- [x] Added unit tests covering ZodTool and FunctionTool source paths
- [x] Added type-level tests verifying input and return type inference
- [x] All 2729 existing tests continue to pass

## Checklist

----

### DIFF
--- a/strands-ts/src/tools/tool-factory.ts
+++ b/strands-ts/src/tools/tool-factory.ts
@@ -1,4 +1,5 @@
-import type { InvokableTool } from './tool.js'
+import type { InvokableTool, ToolContext } from './tool.js'
+import { Tool } from './tool.js'
 import { FunctionTool } from './function-tool.js'
 import type { FunctionToolConfig } from './function-tool.js'
 import type { JSONValue } from '../types/json.js'
@@ -13,6 +14,29 @@ import { ZodTool, type ZodToolConfig } from './zod-tool.js'
  */
 function isZodType(value: unknown): value is z.ZodType {
   return value instanceof z.ZodType
+}
+
+/**
+ * Configuration for creating a tool that reuses another tool's input schema.
+ *
+ * @typeParam TInput - Input type inferred from the source tool
+ * @typeParam TReturn - Return type of the callback function
+ */
+interface DerivedToolConfig<TInput, TReturn extends JSONValue> {
+  /** The unique name of the tool */
+  name: string
+
+  /** Human-readable description of the tool's purpose. Defaults to the source tool's description. */
+  description?: string
+
+  /** An existing tool whose input schema and types will be reused */
+  inputSchema: InvokableTool<TInput, unknown>
+
+  /** Function that implements the tool logic with typed input from the source tool */
+  callback: (
+    input: TInput,
+    context?: ToolContext
+  ) => AsyncGenerator<JSONValue, TReturn, never> | Promise<TReturn> | TReturn
 }
 
 /**
@@ -36,11 +60,26 @@ export function tool<TInput extends z.ZodType, TReturn extends JSONValue = JSONV
 export function tool(config: FunctionToolConfig): InvokableTool<unknown, JSONValue>
 
 /**
- * Creates an InvokableTool from either a Zod schema or JSON schema configuration.
+ * Creates an InvokableTool that reuses another tool's input schema.
+ * The callback receives the same typed input as the source tool.
+ *
+ * @typeParam TInput - Input type inferred from the source tool
+ * @typeParam TReturn - Return type of the callback function
+ * @param config - Tool configuration with a source tool as inputSchema
+ * @returns An InvokableTool with input typed from the source tool
+ */
+export function tool<TInput, TReturn extends JSONValue = JSONValue>(
+  config: DerivedToolConfig<TInput, TReturn>
+): InvokableTool<TInput, TReturn>
+
+/**
+ * Creates an InvokableTool from a Zod schema, JSON schema, or existing tool.
  *
  * When a Zod schema is provided as `inputSchema`, input is validated at runtime and
- * the callback receives typed input. When a JSON schema (or no schema) is provided,
- * the callback receives `unknown` input with no runtime validation.
+ * the callback receives typed input. When an existing tool is provided as `inputSchema`,
+ * the callback receives input typed from that tool's generic parameters. When a JSON
+ * schema (or no schema) is provided, the callback receives `unknown` input with no
+ * runtime validation.
  *
  * @example
  * ```typescript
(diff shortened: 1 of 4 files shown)


strands-ts/src/tools/tool-factory.ts
```diff
@@ -15,6 +16,29 @@ function isZodType(value: unknown): value is z.ZodType {
...
+
+  /** An existing tool whose input schema and types will be reused */
+  inputSchema: InvokableTool<TInput, unknown>
+
+  /** Function that implements the tool logic with typed input from the source tool */
+  callback: (
+    input: TInput,
+    context?: ToolContext
+  ) => AsyncGenerator<JSONValue, TReturn, never> | Promise<TReturn> | TReturn
+}
```
pgrayy, line 40: Each tool type has a config defined like this ([example](https://github.com/strands-agents/sdk-typescript/blob/main/strands-ts/src/tools/zod-tool.ts#L20)).

strands-ts/src/tools/tool-factory.ts
```diff
@@ -15,6 +16,29 @@ function isZodType(value: unknown): value is z.ZodType {
...
 
+/**
+ * Configuration for creating a tool that reuses another tool's input schema.
+ *
+ * @typeParam TInput - Input type inferred from the source tool
+ * @typeParam TReturn - Return type of the callback function
+ */
+interface DerivedToolConfig<TInput, TReturn extends JSONValue> {
+  /** The unique name of the tool */
+  name: string
```
zastrowm, line 27: Any reason not to inherit `name` as well?

strands-ts/src/tools/tool-factory.ts
```diff
@@ -36,11 +60,26 @@ export function tool<TInput extends z.ZodType, TReturn extends JSONValue = JSONV
...
- * Creates an InvokableTool from either a Zod schema or JSON schema configuration.
+ * Creates an InvokableTool that reuses another tool's input schema.
+ * The callback receives the same typed input as the source tool.
+ *
+ * @typeParam TInput - Input type inferred from the source tool
+ * @typeParam TReturn - Return type of the callback function
+ * @param config - Tool configuration with a source tool as inputSchema
+ * @returns An InvokableTool with input typed from the source tool
+ */
+export function tool<TInput, TReturn extends JSONValue = JSONValue>(
```
zastrowm, line 71: Did we consider anything like a `wrapTool` for clarity?

zastrowm (dismissed):

strands-ts/src/tools/tool-factory.ts
zastrowm: Oh I guess we can't have a helper method on tools because they're interfaces could we?  otherwise this devx would be cool and fairly cool:

```
httpRequest.wrap({
  ....
})
```

strands-ts/src/tools/tool-factory.ts
pgrayy: I considered it but figured it would be good to force it at least for now since a derived tool is something different.

strands-ts/src/tools/tool-factory.ts
pgrayy: Just didn't want to give the users a separate export. I thought it might be simpler to overload. But definitely open to suggestions here.

strands-ts/src/tools/zod-tool.ts
```diff
@@ -58,9 +58,8 @@ export class ZodTool<TInput extends z.ZodType | undefined, TReturn extends JSONV
 
   /**
    * Zod schema for input validation.
-   * Note: undefined is normalized to z.void() in constructor, so this is always defined.
    */
-  private readonly _inputSchema: z.ZodType
+  readonly inputSchema: TInput extends z.ZodType ? TInput : z.ZodVoid
```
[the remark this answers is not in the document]
pgrayy: Users might want to access it for custom type inference, testing, or building their own composition utilities. Also, we already expose Zod through the public contract. This is the ZodTool after all.

strands-ts/src/tools/tool-factory.ts
zastrowm: Given that it's agent-visible my gut says that most would want to keep the name.  Like I don't expect both tools being passed to an agent being a common use-case.

Not a blocker FWIW.

strands-ts/src/tools/tool-factory.ts
zastrowm: Of the options that fit our criteria, 

```ts
tool.wrap(httpTool, { ... }
```

Might be the most clear - it guides you towards thinking of it as "inheriting" or "wrapping".

But either way, I don't expect most folks to self-discover this, so I think it's going to be a docs thing; but `inputSpec: tool` seems a *little* weird IMHO

PR #1089, opened 2026-05-19
feat(bedrock): add TTL support for prompt caching

kpx-dev:
## Summary

- Adds optional `ttl` field to `CachePointBlock` so user-supplied cache points can opt into Bedrock's extended TTL (`'5m'` / `'1h'`). The Bedrock formatter preserves it end-to-end.
- Adds optional `ttlTool` and `ttlMessages` fields to `CacheConfig`, applied independently to the SDK's auto-injected cache points (after `toolConfig.tools` and on the last user message). When unset, behavior is unchanged.

Per-target TTL control is needed because Bedrock requires checkpoint TTLs to be **non-increasing** across `toolConfig` → system → messages. A single shared `ttl` would force users into a uniform value and could not satisfy this rule when callers want different durations per checkpoint.

Mirrors sdk-python feature parity:
- strands-agents/sdk-python#1660 — user-supplied `cachePoint.ttl` preservation
- strands-agents/sdk-python#2232 — TTL propagation onto auto-injected cache points (split fields, equivalent to `cache_tools_ttl` + `CacheConfig.ttl` on the Python side)

API reference: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CachePointBlock.html

Closes #1088

## Test plan

- [x] `npm run type-check` clean
- [x] `npm run lint` clean
- [x] `npm run format:check` clean
- [x] `npm run test` — full unit suite passes (2729 tests), including new tests:
  - `CachePointBlock` ttl in constructor / `toJSON` / `fromJSON`
  - Bedrock formatter preserves `ttl` on user-supplied cache points (in messages and system prompt)
  - Bedrock auto-inject propagates `cacheConfig.ttlTool` and `cacheConfig.ttlMessages` independently
  - Backward compat: omitting both fields produces the same output as before

## Example usage

```ts
import { BedrockModel, CachePointBlock, TextBlock } from '@strands-agents/sdk'

// 1) User-supplied cache point with TTL (preserved end-to-end):
new BedrockModel({ modelId: 'us.anthropic.claude-haiku-4-5-20251001-v1:0' })
  .stream(messages, {
    systemPrompt: [
      new TextBlock(largeContext),
      new CachePointBlock({ cacheType: 'default', ttl: '1h' }),
    ],
  })

// 2) Auto-inject with independent TTLs per checkpoint
//    (must respect Bedrock's non-increasing TTL rule across tools -> system -> messages):
new BedrockModel({
  modelId: 'us.anthropic.claude-haiku-4-5-20251001-v1:0',
  cacheConfig: { strategy: 'auto', ttlTool: '1h', ttlMessages: '5m' },
})

// 3) Auto-inject with a single TTL on just the messages checkpoint:
new BedrockModel({
  modelId: 'us.anthropic.claude-haiku-4-5-20251001-v1:0',
  cacheConfig: { strategy: 'auto', ttlMessages: '1h' },
})
```

### DIFF
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -10,6 +10,7 @@
 import {
   BedrockRuntimeClient,
   type BedrockRuntimeClientConfig,
+  type CachePointBlock as BedrockCachePointBlock,
   type ContentBlock as BedrockContentBlock,
   type ContentBlockDeltaEvent as BedrockContentBlockDeltaEvent,
   type ContentBlockStartEvent as BedrockContentBlockStartEvent,
@@ -124,6 +125,18 @@ const DEFAULT_REDACT_INPUT_MESSAGE = '[User input redacted.]'
  * Default message for redacted output.
  */
 const DEFAULT_REDACT_OUTPUT_MESSAGE = '[Assistant output redacted.]'
+
+/**
+ * TTL durations accepted by Bedrock for prompt-cache checkpoints.
+ *
+ * Bedrock currently accepts `'5m'` (default) and `'1h'`. Note that Bedrock requires
+ * checkpoint TTLs to be **non-increasing** across `toolConfig` → system → messages —
+ * setting a longer TTL on a later checkpoint than an earlier one will be rejected by
+ * the service.
+ *
+ * @see https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CachePointBlock.html
+ */
+export type BedrockCacheTTL = '5m' | '1h'
 
 /**
  * Redaction configuration for Bedrock guardrails.
@@ -679,8 +692,12 @@ export class BedrockModel extends Model<BedrockModelConfig> {
       )
 
       if (this._shouldEnableCaching()) {
-        const ttl = this._config.cacheConfig?.ttlTool
-        tools.push({ cachePoint: { type: 'default', ...(ttl !== undefined && { ttl }) } })
+        const cachePoint: BedrockCachePointBlock = { type: 'default' }
+        const ttl = this._config.cacheConfig?.toolsTTL
+        if (ttl !== undefined) {
+          cachePoint.ttl = ttl as BedrockCacheTTL
+        }
+        tools.push({ cachePoint })
       }
 
       const toolConfig: ToolConfiguration = {
@@ -836,8 +853,12 @@ export class BedrockModel extends Model<BedrockModelConfig> {
     if (lastUserIdx !== null) {
       const lastMsg = messages[lastUserIdx]
       if (lastMsg && lastMsg.content) {
-        const ttl = this._config.cacheConfig?.ttlMessages
-        lastMsg.content.push({ cachePoint: { type: 'default', ...(ttl !== undefined && { ttl }) } })
+        const cachePoint: BedrockCachePointBlock = { type: 'default' }
+        const ttl = this._config.cacheConfig?.messagesTTL
+        if (ttl !== undefined) {
+          cachePoint.ttl = ttl as BedrockCacheTTL
+        }
+        lastMsg.content.push({ cachePoint })
         logger.debug(`msg_idx=<${lastUserIdx}> | added cache point to last user message`)
       }
     }
@@ -1042,8 +1063,13 @@ export class BedrockModel extends Model<BedrockModelConfig> {
         }
       }
 
-      case 'cachePointBlock':
-        return { cachePoint: { type: block.cacheType, ...(block.ttl !== undefined && { ttl: block.ttl }) } }
+      case 'cachePointBlock': {
+        const cachePoint: BedrockCachePointBlock = { type: block.cacheType }
+        if (block.ttl !== undefined) {
+          cachePoint.ttl = block.ttl as BedrockCacheTTL
+        }
+        return { cachePoint }
+      }
 
       case 'imageBlock':
         return {
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -679,7 +679,7 @@ export class BedrockModel extends Model<BedrockModelConfig> {
       )
 
       if (this._shouldEnableCaching()) {
-        const ttl = this._config.cacheConfig?.ttl
+        const ttl = this._config.cacheConfig?.ttlTool
         tools.push({ cachePoint: { type: 'default', ...(ttl && { ttl }) } })
       }
 
@@ -836,7 +836,7 @@ export class BedrockModel extends Model<BedrockModelConfig> {
     if (lastUserIdx !== null) {
       const lastMsg = messages[lastUserIdx]
       if (lastMsg && lastMsg.content) {
-        const ttl = this._config.cacheConfig?.ttl
+        const ttl = this._config.cacheConfig?.ttlMessages
         lastMsg.content.push({ cachePoint: { type: 'default', ...(ttl && { ttl }) } })
         logger.debug(`msg_idx=<${lastUserIdx}> | added cache point to last user message`)
       }
--- a/strands-ts/src/models/bedrock.ts
+++ b/strands-ts/src/models/bedrock.ts
@@ -136,7 +136,34 @@ const DEFAULT_REDACT_OUTPUT_MESSAGE = '[Assistant output redacted.]'
  *
  * @see https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CachePointBlock.html
  */
-export type BedrockCacheTTL = '5m' | '1h'
+const BEDROCK_CACHE_TTLS = ['5m', '1h'] as const
+export type BedrockCacheTTL = (typeof BEDROCK_CACHE_TTLS)[number]
+
+/**
+ * Bedrock-specific prompt-caching configuration. Narrows the TTL fields on the common
+ * {@link CacheConfig} to the values accepted by Bedrock.
+ */
+export interface BedrockCacheConfig extends CacheConfig {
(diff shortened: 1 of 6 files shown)


strands-ts/src/models/bedrock.ts
```diff
@@ -835,7 +836,8 @@ export class BedrockModel extends Model<BedrockModelConfig> {
     if (lastUserIdx !== null) {
       const lastMsg = messages[lastUserIdx]
       if (lastMsg && lastMsg.content) {
-        lastMsg.content.push({ cachePoint: { type: 'default' } })
+        const ttl = this._config.cacheConfig?.ttlMessages
+        lastMsg.content.push({ cachePoint: { type: 'default', ...(ttl !== undefined && { ttl }) } })
```
notowen333: for clarity can we here and elsewhere go for:

````
if (ttl !== undefined) {
  cachePoint.ttl = ttl
}
```

strands-ts/src/models/model.ts
```diff
@@ -74,6 +75,18 @@ export interface CacheConfig {
    * - "anthropic": Force enable Anthropic-style caching (useful for application inference profiles)
    */
   strategy: 'auto' | 'anthropic'
+
+  /**
+   * Optional TTL applied to the auto-injected cache point appended after `toolConfig.tools`.
+   * When omitted, the provider's default TTL is used.
+   */
+  ttlTool?: CacheTTL
```
notowen333: prefer `toolsTTL` for closer parity with python and readability

strands-ts/src/models/model.ts
```diff
@@ -74,6 +75,18 @@ export interface CacheConfig {
...
+
+  /**
+   * Optional TTL applied to the auto-injected cache point appended to the last user message.
+   * When omitted, the provider's default TTL is used.
+   */
+  ttlMessages?: CacheTTL
```
notowen333: similarly prefer `messagesTTL`

strands-ts/src/types/messages.ts
```diff
@@ -530,6 +530,15 @@ export class ReasoningBlock
...
 }
 
+/**
+ * TTL duration for a cache point.
+ *
+ * Bedrock currently accepts `'5m'` (default) and `'1h'`.
+ *
+ * @see https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CachePointBlock.html
+ */
+export type CacheTTL = '5m' | '1h'
```
notowen333: We can't make a type that's used in the common `model.ts` which is specifically tied to Bedrock. This should be renamed `BedrockCacheTTL` and be defined in `bedrock.ts`. 

We'll similarly need a path forward for how to define `CacheTTL` in a common representation.

notowen333 (changes requested):

strands-ts/src/models/model.ts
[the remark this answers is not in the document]
notowen333: The bedrock non-increasing check would not live on the common `CacheConfig`

That would go in `bedrock.ts`

strands-ts/src/models/bedrock.ts
kpx-dev: Done — switched all three sites to the imperative form in 05e6236.

strands-ts/src/models/model.ts
kpx-dev: Renamed to `toolsTTL` in 05e6236.

strands-ts/src/models/model.ts
kpx-dev: Renamed to `messagesTTL` in 05e6236.

strands-ts/src/types/messages.ts
kpx-dev: Done in 05e6236 — `CacheTTL` is gone from common code; `BedrockCacheTTL` (`'5m' | '1h'`) now lives in `bedrock.ts`. Common `CachePointBlock.ttl` and `CacheConfig.*TTL` are typed as `string` with provider-specific value spaces noted in TSDoc, and the Bedrock formatter casts at the AWS-SDK boundary.

strands-ts/src/models/model.ts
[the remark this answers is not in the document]
kpx-dev: Moved — non-increasing TTL constraint is now documented in the `BedrockCacheTTL` TSDoc in `bedrock.ts`, not on common `CacheConfig`.

strands-ts/src/models/bedrock.ts
```diff
@@ -679,7 +692,12 @@ export class BedrockModel extends Model<BedrockModelConfig> {
       )
 
       if (this._shouldEnableCaching()) {
-        tools.push({ cachePoint: { type: 'default' } })
+        const cachePoint: BedrockCachePointBlock = { type: 'default' }
+        const ttl = this._config.cacheConfig?.toolsTTL
+        if (ttl !== undefined) {
+          cachePoint.ttl = ttl as BedrockCacheTTL
```
notowen333: this cast doesn't sit well with me... Can we try this pattern:

```
  // model.ts
  export interface CacheConfig {
    strategy: 'auto' | 'anthropic'
    toolsTTL?: string
    messagesTTL?: string
  }

  // bedrock.ts
  const BEDROCK_CACHE_TTLS = ['5m', '1h'] as const
  export type BedrockCacheTTL = (typeof BEDROCK_CACHE_TTLS)[number]

  export interface BedrockCacheConfig extends CacheConfig {
    toolsTTL?: BedrockCacheTTL
    messagesTTL?: BedrockCacheTTL
  }

  function validateBedrockCacheTTL(v: string, field: string): void {
    if (!(BEDROCK_CACHE_TTLS as readonly string[]).includes(v)) {
      throw new Error(<resourceful message>)
    }
  }

  // in BedrockModelConfig:
  cacheConfig?: BedrockCacheConfig
```

notowen333 (changes requested):

strands-ts/src/models/bedrock.ts
kpx-dev: Done in 1f7e2b3 — applied the pattern: `BedrockCacheTTL` from a const array, `BedrockCacheConfig extends CacheConfig` narrows the TTL fields, and `validateBedrockCacheTTL` (assertion function) handles the user-supplied path. No more casts.

strands-ts/src/models/model.ts
```diff
@@ -74,6 +74,20 @@ export interface CacheConfig {
    * - "anthropic": Force enable Anthropic-style caching (useful for application inference profiles)
    */
   strategy: 'auto' | 'anthropic'
+
+  /**
+   * Optional TTL applied to the auto-injected cache point appended after `toolConfig.tools`.
+   * When omitted, the provider's default TTL is used. The accepted value space is
+   * provider-specific.
+   */
+  toolsTTL?: string
```
notowen333: Sorry for the churn here. Discussed with the team and we settled on the below shape. The key gains are around containing the Bedrock logic to bedrock and relying on server side validations so we don't have a stale enum. Since we don't have immediate messagesTTL and toolsTTL for other providers, we shouldn't introduce it.

```
 Changes

  1. model.ts
  
  Remove toolsTTL and messagesTTL from CacheConfig. Restore it to:
  export interface CacheConfig {
    strategy: 'auto' | 'anthropic'
  }
  
  2. bedrock.ts

  Add the Bedrock-specific type and config interface:
  export type BedrockCacheTTL = '5m' | '1h' | (string & {})

  export interface BedrockCacheConfig extends CacheConfig {
    toolsTTL?: BedrockCacheTTL
    messagesTTL?: BedrockCacheTTL
  }
  
  In BedrockModelConfig, change cacheConfig?: CacheConfig to cacheConfig?: BedrockCacheConfig.

  3. Drop the three as BedrockCacheTTL casts

  At bedrock.ts:697, bedrock.ts:858, bedrock.ts:1067 — assignments narrow naturally from BedrockCacheConfig and
  CachePointBlockData.ttl.

  The AWS SDK's CachePointBlock.ttl is strictly '5m' | '1h', while our BedrockCacheTTL includes (string & {}). If TS complains at the
  AWS-boundary assignment, use one cast at that boundary: cachePoint.ttl = ttl as CacheTTL (importing CacheTTL from
  @aws-sdk/client-bedrock-runtime).

  4. No validation

  No validateBedrockCacheTTL, no assertion functions. Bedrock's API rejects bad TTLs server-side with a ValidationException.

  ---
  Gains
  
  - Common CacheConfig stays neutral. Bedrock-specific TTL semantics no longer leak into the shared type that Anthropic and Vercel
  will eventually consume.
  - Bedrock users get autocomplete on '5m' | '1h' at the call site, while still being able to pass any string (forward-compat when AWS
   adds new TTL values — no SDK update needed).
  - No runtime validation to maintain. Bedrock validates server-side and returns a clear ValidationException. No drift between our
  list and AWS's.
  - Existing precedent. BedrockCacheConfig extends CacheConfig matches OpenAIChatConfig extends OpenAIBaseConfig and the broader
  pattern across this codebase.
  - (string & {}) precedent in this codebase too — same idiom as StopReason. Familiar shape for future readers.
  - Net cast count: 0–1. Down from 3
```

notowen333 (changes requested):

strands-ts/src/models/model.ts
kpx-dev: Done in b197200 — `CacheConfig` is back to neutral, TTL fields live on `BedrockCacheConfig` only, `BedrockCacheTTL` uses `'5m' | '1h' | (string & {})` for forward-compat, validator is gone, and there's one boundary cast against the AWS SDK `CacheTTL` at each of the three sites.

notowen333 (approved):

gautamsirdeshmukh (approved):

opieter-aws (approved):

PR #1011, opened 2026-05-06
feat: Sandbox parent PR (base/shell/remote/docker/tools/plugins)

gautamsirdeshmukh:
## NOTE - SPLITTING INTO MULTIPLE PRS

- Sandbox/ShellSandbox base interface --> strands-agents/sdk-typescript#1090
- SSH/Docker sandbox interfaces
- Vended tools/plugins
- Agent integration

---

## Description

Introduces `Sandbox`, a core primitive for the Strands Agent Harness SDK.

`Sandbox` decouples tool *command execution, code execution and filesystem access* from the agent's runtime. Tool logic still runs in the agent's host Node.js process (argument parsing, output formatting, error handling). However, the underlying operations they perform (spawning a shell, running code, file read/write, etc) are routed through the configured `Sandbox` backend.

When a user explicitly provides a `Sandbox` (Docker container, remote host via SSH, etc), the agent automatically gains four vended tools (`exec`, `codeInterpreter`, `fileEditor`) that route their operations through the `Sandbox`. This enables the same agent to execute locally during development or in an isolated/remote environment with a single config change.

When no `Sandbox` is provided, the agent behaves exactly as it does today (operating on the host).

## Overview of Changes

Core Interface (src/sandbox/)

- `Sandbox` abstract class – 7 abstract methods (`executeStreaming`, `executeCodeStreaming`, `readFile`, `writeFile`, `removeFile`, `listFiles`, `statFile`) + convenience wrappers (`execute`, `executeCode`, `readText`, `writeText`)
  - `ShellSandbox` – abstract base for shell-accessible backends (subclasses implement `executeStreaming`)
    - `RemoteSandbox` – SSH transport with `accept-new` host key default
    - `DockerSandbox` – container isolation with start/stop/pause/resume (snapshot)

Vended Tools (src/vended-tools/)

- `exec` — shell command execution with cwd and timeout
- `codeInterpreter` — execute code in any language via stdin piping
- `fileEditor` — view, create, str_replace, insert, undo

All four are auto-registered when a `Sandbox` is explicitly configured on an agent. All four work in both sandbox mode (operations routed through `sandbox.{method}()`) and default mode (native host execution).

Plugin Integrations

- `SandboxStorage` — new storage backend for `ContextOffloader` that writes offloaded content to the sandbox filesystem
- `AgentSkills` – added method `Skill.fromSandbox()` for loading skills from the sandbox filesystem

Agent Integration

[...]

PR #1090, opened 2026-05-21
feat: base Sandbox interface

gautamsirdeshmukh:
## Description

Introduces the `Sandbox` interface, the core abstraction that decouples tool logic from where code runs.

This is the base interface only. Concrete implementations (Docker, SSH), vended tools/plugins and the Agent integration ship in follow-up PRs (incrementally broken off from the parent mega-PR #1011).

## What's Included

Core Interface (src/sandbox/base.ts)

- Sandbox abstract class
  - 6 abstract methods:
    - `executeStreaming(command, options?)` — stream stdout/stderr from a shell command
    - `executeCodeStreaming(code, language, options?)` — stream output from code execution via interpreter
    - `readFile(path)` — read file as Uint8Array
    - `writeFile(path, content)` — write Uint8Array, creates parent directories
    - `removeFile(path)` — delete a file
    - `listFiles(path)` — list directory contents as FileInfo[]
  - 4 convenience methods on the base class:
    - `execute(command, options?)` — non-streaming wrapper over executeStreaming
    - `executeCode(code, language, options?)` — non-streaming wrapper over executeCodeStreaming
    - `readText(path)` — UTF-8 decode wrapper over readFile
    - `writeText(path, content)` — UTF-8 encode wrapper over writeFile

Shell-Based Defaults (src/sandbox/posix-shell.ts)

- `PosixShellSandbox` abstract class (subclasses implement only executeStreaming)
  - Code execution via base64-encoded heredoc piped to interpreter
  - File read/write via base64 encoding over shell
  - Directory listing via ls -1ap with test -d guard

## Related Issues

- #1011 (parent mega-PR containing interfaces, implementation, integration)
- https://github.com/strands-agents/sdk-python/pull/2198 (original Python PR)
- https://github.com/strands-agents/docs/pull/681 (original Python design)

## Documentation PR

Not yet

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/strands-ts/eslint.config.js
+++ b/strands-ts/eslint.config.js
@@ -55,6 +55,9 @@ function sdkRules(options) {
         process: 'readonly',
         setTimeout: 'readonly',
         clearTimeout: 'readonly',
+        atob: 'readonly',
+        btoa: 'readonly',
+        crypto: 'readonly',
       },
     },
     plugins: {
--- a/strands-ts/src/index.ts
+++ b/strands-ts/src/index.ts
@@ -303,6 +303,11 @@ export { AgentTrace } from './telemetry/tracer.js'
 // Local Metrics
 export { AgentMetrics } from './telemetry/meter.js'
 
+// Sandbox
+export { Sandbox, type ExecuteOptions } from './sandbox/base.js'
+export { PosixShellSandbox } from './sandbox/posix-shell.js'
+export type { StreamType, StreamChunk, FileInfo, OutputFile, ExecutionResult } from './sandbox/types.js'
+
 // Multi-agent orchestration
 export { Graph } from './multiagent/index.js'
 export { Swarm } from './multiagent/index.js'
--- a/strands-ts/src/sandbox/__tests__/posix-shell.test.node.ts
+++ b/strands-ts/src/sandbox/__tests__/posix-shell.test.node.ts
@@ -1,0 +1,292 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import fs from 'fs'
+import { TestSandbox } from '../../__fixtures__/test-sandbox.node.js'
+import { streamProcess } from '../stream-process.js'
+import type { ExecutionResult, StreamChunk } from '../types.js'
+
+const TEST_DIR = '/tmp/strands-test-shell-sandbox'
+
+describe.skipIf(process.platform === 'win32')('PosixShellSandbox', () => {
+  let sandbox: TestSandbox
+
+  beforeEach(() => {
+    fs.rmSync(TEST_DIR, { recursive: true, force: true })
+    fs.mkdirSync(TEST_DIR, { recursive: true })
+    sandbox = new TestSandbox(TEST_DIR)
+  })
+
+  afterEach(() => {
+    fs.rmSync(TEST_DIR, { recursive: true, force: true })
+  })
+
+  describe('execute (via shell commands)', () => {
+    it('runs a command', async () => {
+      const result = await sandbox.execute('echo hello')
+      expect(result.exitCode).toBe(0)
+      expect(result.stdout).toBe('hello\n')
+    })
+
+    it('runs in workingDir', async () => {
+      const result = await sandbox.execute('pwd')
+      expect(result.stdout.trim()).toContain('strands-test-shell-sandbox')
+    })
+
+    it('respects cwd option', async () => {
+      const result = await sandbox.execute('pwd', { cwd: '/tmp' })
+      expect(result.stdout.trim()).toMatch(/\/tmp$/)
+    })
+  })
+
+  describe('executeCode (via shell quoting)', () => {
+    it('runs python code through shell', async () => {
+      const result = await sandbox.executeCode('print(2 + 2)', 'python3')
+      expect(result.exitCode).toBe(0)
+      expect(result.stdout).toBe('4\n')
+    })
+
+    it('handles code with special characters', async () => {
+      const result = await sandbox.executeCode('print(\'hello "world"\')', 'python3')
+      expect(result.stdout).toBe('hello "world"\n')
+    })
+
+    it('handles code with single quotes', async () => {
+      const result = await sandbox.executeCode('print("it\'s working")', 'python3')
+      expect(result.stdout).toBe("it's working\n")
+    })
+  })
+
+  describe('language validation', () => {
+    it('rejects path traversal', async () => {
+      await expect(sandbox.executeCode('x', '../../../bin/sh')).rejects.toThrow('invalid characters')
+    })
+
+    it('rejects shell metacharacters', async () => {
+      await expect(sandbox.executeCode('x', 'python;rm -rf /')).rejects.toThrow('invalid characters')
+    })
+
+    it('rejects spaces', async () => {
+      await expect(sandbox.executeCode('x', 'python -c')).rejects.toThrow('invalid characters')
+    })
+
+    it('allows valid interpreters', async () => {
+      const result = await sandbox.executeCode('print("safe")', 'python3')
+      expect(result.exitCode).toBe(0)
+    })
+
+    it('allows dots and hyphens', async () => {
+      const result = await sandbox.executeCode('x', 'fake-lang.99')
+      expect(result.exitCode).toBe(127)
+    })
+  })
+
+  describe('read/write (via base64 encoding over shell)', () => {
+    it('text file roundtrip', async () => {
+      await sandbox.writeText('test.txt', 'hello shell')
+      const text = await sandbox.readText('test.txt')
+      expect(text).toBe('hello shell')
+    })
+
+    it('binary file roundtrip', async () => {
+      const bytes = new Uint8Array([0, 1, 2, 127, 128, 254, 255])
+      await sandbox.writeFile('binary.bin', bytes)
+      const read = await sandbox.readFile('binary.bin')
+      expect(Array.from(read)).toStrictEqual(Array.from(bytes))
+    })
+
+    it('all 256 byte values roundtrip', async () => {
+      const bytes = new Uint8Array(256)
+      for (let i = 0; i < 256; i++) bytes[i] = i
+      await sandbox.writeFile('all-bytes.bin', bytes)
+      const read = await sandbox.readFile('all-bytes.bin')
+      expect(Array.from(read)).toStrictEqual(Array.from(bytes))
+    })
+
+    it('creates parent directories', async () => {
+      await sandbox.writeText('deep/nested/file.txt', 'deep')
+      const text = await sandbox.readText('deep/nested/file.txt')
+      expect(text).toBe('deep')
+    })
+
+    it('handles unicode content', async () => {
+      const content = '日本語 🚀 émojis'
+      await sandbox.writeText('unicode.txt', content)
+      const text = await sandbox.readText('unicode.txt')
+      expect(text).toBe(content)
+    })
+
+    it('handles shell metacharacters in content', async () => {
+      const content = '$(rm -rf /) `whoami` && || $HOME'
+      await sandbox.writeText('meta.txt', content)
+      const text = await sandbox.readText('meta.txt')
+      expect(text).toBe(content)
+    })
+
+    it('throws on nonexistent file', async () => {
+      await expect(sandbox.readFile('nope.txt')).rejects.toThrow()
+    })
+  })
+
+  describe('remove', () => {
+    it('removes a file', async () => {
+      await sandbox.writeText('delete-me.txt', 'bye')
+      await sandbox.removeFile('delete-me.txt')
+      await expect(sandbox.readFile('delete-me.txt')).rejects.toThrow()
+    })
+
+    it('throws on nonexistent file', async () => {
+      await expect(sandbox.removeFile('nope.txt')).rejects.toThrow()
+    })
+  })
+
+  describe('list (via ls -1ap parsing)', () => {
+    it('lists directory contents', async () => {
+      await sandbox.writeText('a.txt', 'a')
+      await sandbox.writeText('b.txt', 'b')
+      const files = await sandbox.listFiles('.')
+      const names = files.map((f) => f.name)
+      expect(names).toContain('a.txt')
+      expect(names).toContain('b.txt')
+    })
+
+    it('identifies directories', async () => {
+      await sandbox.execute('mkdir -p subdir')
+      const files = await sandbox.listFiles('.')
+      const subdir = files.find((f) => f.name === 'subdir')
+      expect(subdir?.isDir).toBe(true)
+    })
+
+    it('excludes . and .. entries', async () => {
+      await sandbox.writeText('file.txt', '')
+      const files = await sandbox.listFiles('.')
+      const names = files.map((f) => f.name)
+      expect(names).not.toContain('.')
+      expect(names).not.toContain('..')
+    })
+
+    it('throws on nonexistent directory', async () => {
+      await expect(sandbox.listFiles('/tmp/nonexistent-dir-xyz')).rejects.toThrow()
+    })
+
+    it('throws when path is a file, not a directory', async () => {
+      await sandbox.writeText('not-a-dir.txt', 'hello')
+      await expect(sandbox.listFiles('not-a-dir.txt')).rejects.toThrow()
+    })
+  })
+
+  describe('shellQuote', () => {
+    it('handles paths with spaces', async () => {
+      await sandbox.execute('mkdir -p "with spaces"')
+      await sandbox.writeText('with spaces/file.txt', 'spaced')
+      const text = await sandbox.readText('with spaces/file.txt')
+      expect(text).toBe('spaced')
+    })
+
+    it('handles paths with single quotes', async () => {
+      await sandbox.execute('mkdir -p "it\'s"')
+      await sandbox.writeText("it's/file.txt", 'quoted')
+      const text = await sandbox.readText("it's/file.txt")
+      expect(text).toBe('quoted')
+    })
+  })
+
+  describe('timeout', () => {
+    it('kills process on timeout', async () => {
+      const start = Date.now()
+      await expect(sandbox.execute('sleep 60', { timeout: 0.2 })).rejects.toThrow('timed out')
+      const elapsed = Date.now() - start
+      expect(elapsed).toBeLessThan(2000)
+    })
+
+    it('does not timeout fast commands', async () => {
+      const result = await sandbox.execute('echo fast', { timeout: 5 })
+      expect(result.exitCode).toBe(0)
+      expect(result.stdout).toBe('fast\n')
+    })
+  })
+
+  describe('abort signal', () => {
+    it('kills process when signal is aborted', async () => {
+      const controller = new AbortController()
+      const promise = sandbox.execute('sleep 60', { signal: controller.signal })
+      setTimeout(() => controller.abort(), 100)
+      await expect(promise).rejects.toThrow('aborted')
+    })
+
+    it('rejects immediately if signal is already aborted', async () => {
+      const controller = new AbortController()
+      controller.abort()
+      await expect(sandbox.execute('sleep 60', { signal: controller.signal })).rejects.toThrow('aborted')
+    })
+  })
+
+  describe('concurrent execution', () => {
+    it('handles multiple concurrent commands', async () => {
+      const results = await Promise.all([
+        sandbox.execute('echo one'),
+        sandbox.execute('echo two'),
+        sandbox.execute('echo three'),
+      ])
+      expect(results.map((r) => r.stdout.trim()).sort()).toStrictEqual(['one', 'three', 'two'])
+    })
+
+    it('handles concurrent file writes to different files', async () => {
+      await Promise.all([
+        sandbox.writeText('a.txt', 'aaa'),
+        sandbox.writeText('b.txt', 'bbb'),
+        sandbox.writeText('c.txt', 'ccc'),
+      ])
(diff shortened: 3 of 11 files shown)


strands-ts/src/sandbox/__tests__/shell.test.node.ts
```diff
@@ -0,0 +1,287 @@
...
+  async *executeStreaming(
+    command: string,
+    options?: ExecuteOptions
+  ): AsyncGenerator<StreamChunk | ExecutionResult, void, undefined> {
+    const cwd = options?.cwd ?? this.workingDir
+    const fullCommand = `cd ${shellQuote(cwd)} && ${command}`
+    const proc = spawn('sh', ['-c', fullCommand])
+    yield* streamProcess(proc, { timeout: options?.timeout, signal: options?.signal })
+  }
+}
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Refactoring to use TestSandbox in both places

strands-ts/src/sandbox/index.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Sandbox abstraction providing isolated environments for agent tools —
+ * command execution, code interpretation, and filesystem access.
+ */
+
+export { Sandbox, type ExecuteOptions } from './base.js'
+export { ShellSandbox } from './shell.js'
```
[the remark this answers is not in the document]
gautamsirdeshmukh: `streamProcess` is intentionally internal, all implementations that need it (Docker, SSH) will live within this package and import it directly. External consumers extending `ShellSandbox` only need to implement `executeStreaming`, they don't need `streamProcess` themselves

strands-ts/src/sandbox/stream-process.ts
```diff
@@ -0,0 +1,168 @@
...
+ * Stream stdout/stderr from a child process, then yield the final result.
+ *
+ * Bridges Node.js event emitters to an async generator. Chunks are
+ * yielded incrementally as the process produces output. The final
+ * yield is an ExecutionResult with the exit code and complete output.
+ *
+ * All listeners are attached synchronously before any await to prevent
+ * missed events from fast-completing processes.
+ */
+export async function* streamProcess(
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Adding

strands-ts/src/sandbox/stream-process.ts
```diff
@@ -0,0 +1,168 @@
...
+  SIGINT: 2,
+  SIGQUIT: 3,
+  SIGABRT: 6,
+  SIGKILL: 9,
+  SIGSEGV: 11,
+  SIGPIPE: 13,
+  SIGTERM: 15,
+}
+
+export interface StreamProcessOptions {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Adding

strands-ts/src/index.ts
```diff
@@ -296,6 +296,10 @@ export { AgentTrace } from './telemetry/tracer.js'
 // Local Metrics
 export { AgentMetrics } from './telemetry/meter.js'
 
+// Sandbox
+export { Sandbox, type ExecuteOptions } from './sandbox/base.js'
+export type { StreamType, StreamChunk, FileInfo, OutputFile, ExecutionResult } from './sandbox/types.js'
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Updating to export from barrel

strands-ts/src/sandbox/shell.ts
```diff
@@ -0,0 +1,96 @@
+/**
+ * Shell sandbox with default implementations for file and code operations.
+ *
+ * Subclasses only need to implement {@link ShellSandbox.executeStreaming} —
+ * all other operations are implemented by running shell commands through it.
+ * Use this for remote environments where only shell access is available
+ * (Docker containers, SSH connections, cloud runtimes).
+ */
+
+import { randomBytes } from 'crypto'
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Replacing `randomBytes` with `randomUUID`

strands-ts/src/utils/shell-quote.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Shell-escape a string for safe inclusion in a shell command.
+ *
+ * Wraps the value in single quotes and escapes any embedded single quotes
+ * using the '\'' pattern. Single quotes disable all shell expansion
+ * (variables, backticks, globbing), making this safe against injection.
+ */
+export function shellQuote(value: string): string {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Adding

strands-ts/src/sandbox/constants.ts
```diff
@@ -0,0 +1 @@
+export const LANGUAGE_PATTERN = /^[a-zA-Z0-9._-]+$/
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Adding

strands-ts/src/sandbox/__tests__/posix-shell.test.node.ts
```diff
@@ -0,0 +1,258 @@
...
+    sandbox = new TestSandbox(TEST_DIR)
+  })
+
+  afterEach(() => {
+    execSync(`rm -rf ${TEST_DIR}`)
+  })
+
+  describe('execute (via shell commands)', () => {
+    it('runs a command', async () => {
+      const result = await sandbox.execute('echo hello')
```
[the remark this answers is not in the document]
gautamsirdeshmukh: These tests verify specific behaviors (timeout kills process, cwd is respected, binary roundtrips correctly, etc.), not the full result shape. Full object assertions would make every test brittle to future additions to ExecutionResult and obscure what each test is actually checking. 

This also matches the existing convention across the codebase where toBe/toContain is the dominant pattern for behavioral tests.

strands-ts/src/sandbox/base.ts
```diff
@@ -0,0 +1,166 @@
...
+   * Execute a shell command, streaming output.
+   *
+   * Yields {@link StreamChunk} objects for stdout and stderr as output
+   * arrives. The final yield is an {@link ExecutionResult} with the
+   * exit code and complete output.
+   *
+   * @param command - The shell command to execute.
+   * @param options - Execution options (timeout, cwd).
+   */
+  abstract executeStreaming(command: string, options?: ExecuteOptions): AsyncIterable<StreamChunk | ExecutionResult>
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Added

strands-ts/eslint.config.js
```diff
@@ -55,6 +55,11 @@ function sdkRules(options) {
         process: 'readonly',
         setTimeout: 'readonly',
         clearTimeout: 'readonly',
+        setInterval: 'readonly',
+        clearInterval: 'readonly',
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Removed

strands-ts/src/sandbox/stream-process.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+  }
+
+  const terminate = (reason: Error): void => {
+    if (terminating || done) return
+    terminating = true
+    error = reason
+    proc.kill('SIGTERM')
+    wake()
+    setTimeout(() => {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Addressed

agent-of-gsird (approved):
## Independent Thorough Review — PR #1090 (Base Sandbox Interface)

**Assessment**: ✅ Approve

@gautamsirdeshmukh — I cloned the branch and read every new file in full: `base.ts`, `shell.ts`, `types.ts`, `constants.ts`, `stream-process.ts`, `shell-quote.ts`, the test fixture, and the full test suite. This is an independent read — here's my evaluation.

---

### ✅ What is genuinely well done

**`base.ts` — Clean, minimal abstract class**
- 6 abstract operations + 4 convenience wrappers is the right split. Streaming as the primitive with `execute()`/`executeCode()` as consumers is a clean pattern.
- TSDoc is complete with proper `@param`, `@returns`, `@throws` annotations.
- No lifecycle methods (start/stop/pause) — addresses @mkmeral's concern from #1011. The sandbox is assumed running when passed to tools.
- `ExecuteOptions` with `timeout`, `cwd`, and `signal` covers the essential knobs without overloading.

**`shell.ts` — ShellSandbox template method pattern**
- "Implement one method (`executeStreaming`), get everything else for free" — excellent DX for implementors.
- Dynamic sentinels (`crypto.randomUUID()`) prevent the injection attack @mkmeral flagged.
- Base64 heredoc approach for `writeFile` avoids `ARG_MAX` limits on large files.
- `listFiles` uses `.replace(/\r$/, '')` (not `.trim()`) — preserves filenames with leading/trailing whitespace.
- `LANGUAGE_PATTERN` regex blocks path traversal, spaces, and shell metacharacters.

**`stream-process.ts` — Process management is correct**
- All event listeners attached synchronously before the first `await` — no race on fast-completing processes.
- Signal → exit-code mapping (`128 + SIGNAL_CODES[signal]`) is correct.
- SIGTERM → SIGKILL escalation after 1s with `terminating` flag prevents double-kill.
- AbortSignal support with proper cleanup (`removeEventListener` in `finally`).
- `enoentMessage` for friendly error when the binary isn't found.
- `setTimeout(resolve, 50)` wake-up ensures `generator.return()` always reaches `finally` (the process-leak fix from #1011).

**`shell-quote.ts` — Correct and minimal**
- Single-quote wrapping with `'\''` escape is the canonical POSIX shell quoting approach. Injection-safe.

**Test suite (`shell.test.node.ts`)**
- 30+ tests covering: basic execution, cwd, code execution, language validation (injection, traversal, metacharacters), binary roundtrip (all 256 bytes), unicode, shell metacharacters in content, remove, list (directories, dot-entries, non-directory path), concurrent execution, streaming, timeout, AbortSignal. Comprehensive.

---

### 🟡 Minor observations (non-blocking)

**1. No `statFile` in this PR**

The parent PR (#1011) had `statFile(path)` on the base class. This PR omits it — I assume it will come in a follow-up (implementations PR) where Docker/Remote need it. Worth noting that `fileEditor` in the tools PR will need it for size-guarding and binary detection. Just flagging for sequencing awareness.

**2. Export surface for implementors**

`Sandbox`, `ShellSandbox`, `ExecuteOptions`, and all types are exported from the main `index.ts`. `shellQuote`, `streamProcess`, `LANGUAGE_PATTERN` are **not** exported — they're internal utilities. This is the right decision for most users.

However, if someone builds a custom `ShellSandbox` subclass that needs `streamProcess` (e.g., for a cloud SSH backend), they'd need to import from the deep path. Consider whether a `./sandbox` sub-export with `streamProcess` and `shellQuote` is needed for the implementations PR.

**3. `OutputFile[]` in `ExecutionResult`**

Always `[]` for shell-based sandboxes. This is forward-looking API design for future Jupyter/API-backed sandboxes. Intentional and documented. ✅

**4. `killTimer` in `stream-process.ts` is not `unref()`'d**

The `killTimer = setTimeout(...)` that escalates to SIGKILL keeps the event loop alive for 1 second even if the process has exited. Since `clearTimeout(killTimer)` is in `finally`, this is handled correctly for the normal flow. But if an error occurs between `terminate()` and the `finally` block (unlikely but possible), the timer would keep the process alive. Adding `.unref()` would be defensive:
```typescript
killTimer = setTimeout(() => { if (!done) proc.kill('SIGKILL') }, 1000)
killTimer.unref()
```
Non-blocking — the `finally` cleanup handles the normal case correctly.

---

### 📋 Process

| # | Item | Recommendation |
|---|---|---|
| 1 | `needs-api-review` label | Recommend applying — new abstract class customers will extend |
| 2 | Documentation PR | Link a tracking issue before merge |

---

### Summary

This is a clean, well-scoped extraction from #1011. The interface is minimal, streaming-first, and addresses all the architectural concerns raised by @mkmeral (no lifecycle methods, no tool coupling, no agent integration in this PR). The test suite is comprehensive and exercises real shell execution with security-relevant edge cases.

All issues I raised across 17+ review passes on #1011 that are relevant to this scope have been correctly addressed in this extraction.

**No technical objections. Ready to merge.**

---

strands-ts/src/sandbox/shell.ts
```diff
@@ -0,0 +1,95 @@
...
+ * Subclasses only need to implement {@link executeStreaming}. The remaining
+ * operations — `executeCodeStreaming`, `readFile`, `writeFile`, `removeFile`,
+ * and `listFiles` — are implemented via shell commands piped through
+ * `executeStreaming`.
+ *
+ * Subclasses may override any method with a native implementation for
+ * better performance or to handle edge cases (e.g., binary-safe file
+ * transfer via Docker stdin pipes, or native API calls for cloud backends).
+ */
+export abstract class ShellSandbox extends Sandbox {
```
mkmeral: not 100% sure, but wanted to open up discussion. should we call it posix shell sandbox? or is shell okay? 

I'm leaning latter, but if we are being explicit about it, it might make sense to call something like `PosixShellSandbox`? 🤔

strands-ts/src/sandbox/base.ts
```diff
@@ -0,0 +1,173 @@
+/**
+ * Base sandbox interface for agent tools.
```
mkmeral: another devx question. should the sandbox implement tool provider in any way? 🤔 

i was thinking about adding default tools when sandbox is provided, but we cannot probably be too generic about it, if we want those tools to be good (i.e. some sandboxes have limitations, different os, tools, etc). for example code interpereter tool is a good generic tool that wouldn't work if env doesn't have python (and llm would default to that, if it's not prompted properly)

cc @gautamsirdeshmukh @zastrowm

mkmeral: I do feel like there is value in being able to optimize these tools per environment. Not sure what's the best shape to achieve it yet 

ps: we probably want both as a toolprovider and as a high level agent.sandbox, because sandbox is also a cross-cutting concern (it is also used by plugins e.g. context offloader, skills)

strands-ts/src/sandbox/__tests__/shell.test.node.ts
```diff
@@ -0,0 +1,258 @@
...
+import { execSync } from 'child_process'
+import { TestSandbox } from '../../__fixtures__/test-sandbox.node.js'
+
+const TEST_DIR = '/tmp/strands-test-shell-sandbox'
+
+describe.skipIf(process.platform === 'win32')('ShellSandbox', () => {
+  let sandbox: TestSandbox
+
+  beforeEach(() => {
+    execSync(`rm -rf ${TEST_DIR} && mkdir -p ${TEST_DIR}`)
```
zastrowm: Can we use node apis for this? It's odd to shell out to do file deletion and creation

strands-ts/src/sandbox/base.ts
zastrowm: > I do feel like there is value in being able to optimize these tools per environment

Does this jive with having tools use the sandbox?  What sort of optimizations do we imagine?  

I wonder if there's a general feature here which is "optimizing tools for different use cases"?

zastrowm: > we probably want both as a toolprovider and as a high level agent.sandbox

Yeah, I can imagine this, but I think we do this as a follow-up since it will be opt-in; I don't imagine we want a sandbox contributing tools by default (that's now what you were saying, is it?)

strands-ts/src/sandbox/shell.ts
zastrowm: If the expected use case is deriving implementations from it, I'd go with the longer/verbose name.

The devx is more about sandbox builders so I like making the expectations clear that the expectation is `PoxixShell`

strands-ts/src/sandbox/shell.ts
```diff
@@ -0,0 +1,95 @@
...
+  async *executeCodeStreaming(
+    code: string,
+    language: string,
+    options?: ExecuteOptions
+  ): AsyncGenerator<StreamChunk | ExecutionResult, void, undefined> {
+    if (!LANGUAGE_PATTERN.test(language)) {
+      throw new Error(`language parameter contains unsafe characters: ${language}`)
```
zastrowm: It's not really about safety, it's about invalid-ness IMHO:

```suggestion
      throw new Error(`language parameter contains invalid characters: ${language}`)
```

strands-ts/src/sandbox/posix-shell.ts
```diff
@@ -0,0 +1,95 @@
...
+  async *executeCodeStreaming(
+    code: string,
+    language: string,
+    options?: ExecuteOptions
+  ): AsyncGenerator<StreamChunk | ExecutionResult, void, undefined> {
+    if (!LANGUAGE_PATTERN.test(language)) {
+      throw new Error(`language parameter contains unsafe characters: ${language}`)
+    }
+    const encoded = btoa(Array.from(new TextEncoder().encode(code), (b) => String.fromCharCode(b)).join(''))
+    const eof = `STRANDS_EOF_${crypto.randomUUID().slice(0, 16)}`
```
zastrowm, line 39: Is `crypto` supported in the browser?

strands-ts/src/sandbox/types.ts
```diff
@@ -0,0 +1,61 @@
...
+ * A file produced as output by code execution.
+ *
+ * Used to carry binary artifacts (images, charts, PDFs, compiled files)
+ * from sandbox execution back to the agent. Shell-based sandboxes
+ * typically return an empty array. Jupyter-backed or API-backed
+ * sandboxes can populate this with generated artifacts.
+ */
+export interface OutputFile {
+  readonly name: string
+  readonly content: Uint8Array
```
zastrowm, line 48: Do we end up having limits on what file output can be returned?   E.g. do we need to document some max size?

strands-ts/src/sandbox/types.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Result of command or code execution in a sandbox.
+ */
+export interface ExecutionResult {
+  readonly type: 'executionResult'
+  readonly exitCode: number
+  readonly stdout: string
+  readonly stderr: string
+  readonly outputFiles: OutputFile[]
```
zastrowm, line 60: When is this used?

gautamsirdeshmukh: This field is for sandbox implementations whose `executeStreaming`/`executeCodeStreaming` methods produce file artifacts (plots, PDFs, etc). It allows those artifacts to be returned inline so tools can pass them back to the model as content blocks without needing to make another trip using `readFile`.

strands-ts/src/sandbox/types.ts
gautamsirdeshmukh: No limits, `content` is bounded by available memory like any other byte array, we aren't introducing a new risk.   

Both `executeStreaming` and `executeCodeStreaming` methods on the base `PosixShellSandbox` always return [] for outputFiles. If a subclass's implementation of those methods produces file artifacts, it can control any size limits there.

strands-ts/src/sandbox/posix-shell.ts
gautamsirdeshmukh: Yep - and furthermore, we don't use any Node-specific dependencies in the posix shell sandbox

strands-ts/src/sandbox/__tests__/shell.test.node.ts
gautamsirdeshmukh: Good call, updating

strands-ts/src/sandbox/base.ts
gautamsirdeshmukh: Agreed, not in scope for this PR. Will address the tool provider pattern and `agent.sandbox` access in either the vended tools/plugins or Agent integration PR. 

(On that topic, I think file operation tools should be included by default, as well as shell tool for shell sandbox. This is pretty standard across the industry. On the other hand, code interpreter should NOT be included automatically, at least to begin with.)

strands-ts/src/sandbox/shell.ts
gautamsirdeshmukh: I'm aligned with `PosixShellSandbox`

strands-ts/src/sandbox/__tests__/posix-shell.test.node.ts
```diff
@@ -0,0 +1,259 @@
...
+      await sandbox.removeFile('delete-me.txt')
+      await expect(sandbox.readFile('delete-me.txt')).rejects.toThrow()
+    })
+
+    it('throws on nonexistent file', async () => {
+      await expect(sandbox.removeFile('nope.txt')).rejects.toThrow()
+    })
+  })
+
+  describe('list (via ls -1aF parsing)', () => {
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Done

strands-ts/src/index.ts
```diff
@@ -303,6 +303,11 @@ export { AgentTrace } from './telemetry/tracer.js'
 // Local Metrics
 export { AgentMetrics } from './telemetry/meter.js'
 
+// Sandbox
+export { Sandbox, type ExecuteOptions } from './sandbox/base.js'
+export { PosixShellSandbox } from './sandbox/posix-shell.js'
+export type { StreamType, StreamChunk, FileInfo, OutputFile, ExecutionResult } from './sandbox/types.js'
```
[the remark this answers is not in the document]
gautamsirdeshmukh: Out of scope for this PR. Will address this if needed in an upcoming change.

mkmeral (dismissed):

mkmeral (approved):

PR #1091, opened 2026-05-21
feat(strands-py-wasm): wire Agent runtime through wasmtime-py

chaynabors:
Python agent runtime working end to end.

- Runtime adapter at `strands-py-wasm/src/strands/_runtime.py`
- `_generated.py` from host bindgen I'm upstreaming
- Swap `stream<T>` for pull-based resources because wasmtime doesn't yet have good support for streams.
- `BedrockModel` resolves AWS creds via botocore on the host since the wasm bundle uses the AWS SDK browser build
- `@aws/bedrock-token-generator` pinned to a forked release tarball
- Deleted the WASI buffer-reuse patch since it appears unneeded. Likely another bug. May later revert.
- Fixed some quirks in the TS SDK that meant some functions were unresolved in the component.


strands-ts/package.json
```diff
@@ -123,7 +123,7 @@
     "@ai-sdk/openai": "^3.0.41",
     "@ai-sdk/provider": "^3.0.0",
     "@anthropic-ai/sdk": "^0.92.0",
-    "@aws/bedrock-token-generator": "^1.1.0",
+    "@aws/bedrock-token-generator": "https://github.com/pgrayy/wasm-deps/releases/download/v44.0.3/aws-bedrock-token-generator-1.1.0.tgz",
```
zastrowm: This forces this dependency for production TS users, no?  If we need to do this can we do it on a WASM specific branch?

strands-py-wasm/src/strands/__init__.py
```diff
@@ -657,125 +619,105 @@ def __init__(
...
-    callbacks.
-    """
+    """Declare an MCP client the host should open and route tools from."""
 
     def __init__(
         self,
         *,
         client_id: str,
-        transport: _t.McpTransport,
+        transport: Any,
```
zastrowm, line 741: Why did this turn into any out of curiosity?

strands-py-wasm/src/strands/__init__.py
```diff
@@ -581,43 +546,40 @@ def _json_default(obj: Any) -> Any:
...
 
 
 class Graph(_t.GraphConfig):
     def __init__(
         self,
         *,
         id: str,
-        nodes: list[_t.NodeConfig],
-        edges: list[_t.EdgeConfig] | None = None,
+        nodes: list[Any],
```
zastrowm, line 602: Oh, Any seems to be used throughout now?

strands-py-wasm/src/strands/__init__.py
```diff
@@ -208,108 +211,94 @@ def __init__(
...
+        _t.OpenaiConfig(model_id=model_id, api_key=api_key, additional_config=_extras_to_json(extras))
+    )
 
 
-class GoogleModel(_t.ModelConfig_Gemini):
-    def __init__(self, model_id: str | None = None, *, api_key: str | None = None, **extras: Any) -> None:
-        super().__init__(
-            value=_t.GeminiConfig(model_id=model_id, api_key=api_key, additional_config=_extras_to_json(extras))
-        )
+def GoogleModel(model_id: str | None = None, *, api_key: str | None = None, **extras: Any) -> Any:
```
zastrowm, line 276: Is the shift from classes to functions expected?

strandly/scripts/generate_types.py
zastrowm, line 1: Can you fill out the PR description with the context/background for this?

strands-ts/package.json
chaynabors: I noticed this too and overrode in strands-wasm. Will appear in next revision

strands-py-wasm/src/strands/__init__.py
chaynabors: Bindgen is work in progress. Not what I actually want to surface as the API. Variants need special casing.

strands-py-wasm/src/strands/__init__.py
chaynabors: I'm in the business of plumbing right now. Patrick is going to pick up API shape and tailor toward something we actually want to expose.

pgrayy (approved):

PR #1098, opened 2026-05-22
fix(wasm): update deps and externalize node-only modules for WASM build

pgrayy:
## Motivation

PR #1070 (`feat(mcp): load MCP servers from JSON`) introduced `mcp-config.ts` which dynamically imports `node:fs/promises`, `node:os`, and `node:path`. It also dynamically imports `@modelcontextprotocol/sdk/client/stdio.js` (which statically imports `node:process` and `node:stream`). When esbuild bundles with `platform: "browser"` and a single `outfile`, it eagerly resolves dynamic imports and inlines their contents, hoisting those Node-only static imports to the top level of the bundle. componentize-js (StarlingMonkey) then fails at module initialization because it cannot resolve `node:process`.

This also updates the `pgrayy-wasmtime` and `@aws/bedrock-token-generator` dependency versions to their latest published releases.

## Public API Changes

No public API changes.

## Type of Change

Bug fix

## Testing

- Verified `npm run build -w strands-wasm` produces `strands-agent.wasm` successfully

### DIFF
--- a/strands-wasm/build.js
+++ b/strands-wasm/build.js
@@ -35,7 +35,15 @@ await build({
   platform: 'browser',
   target: 'es2022',
   define: { 'import.meta.vitest': 'undefined' },
-  external: ['strands:*', 'fs', 'path'],
+  external: [
+    '@modelcontextprotocol/sdk/client/sse.js',
+    '@modelcontextprotocol/sdk/client/stdio.js',
+    'child_process',
+    'fs',
+    'node:*',
+    'path',
+    'strands:*',
+  ],
   outfile: 'dist/bundle.js',
   logLevel: 'info',
 });
--- a/package-lock.json
+++ b/package-lock.json
@@ -1345,7 +1345,7 @@
     },
     "node_modules/@aws/bedrock-token-generator": {
       "version": "1.1.0",
-      "resolved": "https://github.com/pgrayy/wasm-deps/releases/download/v44.0.3/aws-bedrock-token-generator-1.1.0.tgz",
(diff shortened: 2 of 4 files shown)


strands-wasm/build.js
```diff
@@ -35,7 +35,15 @@ await build({
   platform: 'browser',
   target: 'es2022',
   define: { 'import.meta.vitest': 'undefined' },
-  external: ['strands:*', 'fs', 'path'],
+  external: [
+    '@modelcontextprotocol/sdk/client/sse.js',
+    '@modelcontextprotocol/sdk/client/stdio.js',
```
zastrowm, line 40: Should all MCP be externalized?

strands-wasm/build.js
```diff
@@ -35,7 +35,15 @@ await build({
...
+    'child_process',
+    'fs',
+    'node:*',
+    'path',
+    'strands:*',
```
zastrowm, line 45: What does this do?

zastrowm (approved):

strands-wasm/build.js
pgrayy: Much of MCP is runtime agnostic. It's similar to how OTEL does it.

strands-wasm/build.js
pgrayy: This tells the bundler that if it comes across one of these imports, just leave them alone. Don't try to resolve them. They will be resolved at runtime. However, in our case, some of these are dynamic imports and so we just let runtime errors propagate.

pgrayy: If you are asking though why `strands:*` specifically is in this list, I am not sure, I was going to follow up on that.

zastrowm: Yeah, strands specifically was my concern

PR #1106, opened 2026-05-26
feat: add per-invocation limits (turns / outputTokens / totalTokens)

notowen333:
## Description

Adds an optional `limits` object to `Agent.invoke()` / `Agent.stream()` via `InvokeOptions`, with three per-invocation caps: `turns`, `outputTokens`, and `totalTokens`.

Today an agent loop runs until the model stops requesting tools, a hook halts it, or it's cancelled — there is no caller-side guardrail on cost or runaway behavior. Callers who want to bound a run have to wire a hook themselves or wrap the call in a timeout, which races against the model rather than terminating cleanly at a turn boundary.

When a cap is reached, the agent loop terminates gracefully at the next turn boundary and returns an `AgentResult` with a new `stopReason` — no throw, mirroring how `'cancelled'` already works. `lastMessage` and `metrics` remain accessible on the result.

```typescript
// Cap loop iterations
const result = await agent.invoke('Plan a trip', { limits: { turns: 5 } })
if (result.stopReason === 'limitTurns') { /* ... */ }

// Cap cumulative model-generated tokens
const result = await agent.invoke('Summarize', { limits: { outputTokens: 10_000 } })
if (result.stopReason === 'limitOutputTokens') { /* ... */ }

// Cap cumulative input + output tokens (compounds across turns; approximates billed spend)
const result = await agent.invoke('Research X', { limits: { totalTokens: 100_000 } })
if (result.stopReason === 'limitTotalTokens') { /* ... */ }

// Combine
const result = await agent.invoke('Plan a trip', {
  limits: { turns: 10, outputTokens: 50_000, totalTokens: 200_000 },
})
```

`StopReason` gains three string variants: `'limitTurns'`, `'limitOutputTokens'`, `'limitTotalTokens'`. The existing `'maxTokens'` (and `MaxTokensError`) is unchanged — that one signals the model provider's per-call output cap and still throws. The new caller-set caps are graceful.

### Why a `limits` object instead of top-level options

Grouping under `limits` keeps the top-level `InvokeOptions` surface stable as future caps (wall-clock, cost, etc.) are added, and reads naturally as a cohesive concept. The stopReasons are namespaced with a `limit` prefix so granular telemetry / log buckets are preserved — a single `limitExceeded` reason would force callers to derive which cap tripped from `result.metrics` + `limits`, losing the dashboard-level cardinality.

### Semantics

- **Per-invocation, not lifetime**: counters reset on each `invoke()` / `stream()` call against the same agent.
- **Priority on simultaneous trip**: `turns` → `totalTokens` → `outputTokens`.
- **Reinvokable state**: tools requested by the previous turn always run to completion before a cap fires, so `agent.messages` stays in a state the agent can be reinvoked from.
- **Soft cap**: a single oversized model response can overshoot the budget by one turn — the cap is checked at turn boundaries, not within an individual model call.
- **Validation**: each cap, when set, must be a positive finite number; invalid values throw `TypeError` at the start of the invocation.

Backward compatible; omitting `limits` preserves prior behavior.

## Related Issues

N/A

## Documentation PR

N/A

## Type of Change

New feature

## Testing

- [x] I ran `npm run check`

## Checklist

----

### DIFF
--- a/strands-ts/src/types/agent.ts
+++ b/strands-ts/src/types/agent.ts
@@ -118,6 +118,73 @@ export interface InvokeOptions {
    * ```
    */
   cancelSignal?: AbortSignal
+
+  /**
+   * Maximum number of agent loop iterations (turns) for this invocation.
+   * A turn is one model call plus any tool execution that follows.
+   *
+   * Per-invocation: counted from this `invoke()` / `stream()` call, not
+   * cumulative across reuses of the same agent. Checked at the top of each
+   * loop iteration against
+   * `metrics.latestAgentInvocation.cycles.length`. Tools requested by the
+   * previous turn always run to completion before the cap fires, so
+   * `agent.messages` remains in a reinvokable state.
+   *
+   * Setting `maxTurns: 0` returns immediately with
+   * `stopReason: 'maxTurnsExceeded'` and a synthesized empty assistant
+   * message — no model call is made and no user input is appended.
+   *
+   * Omit (or set to `undefined`) for no limit.
+   */
+  maxTurns?: number
+
+  /**
+   * Maximum cumulative model-generated tokens for this invocation, summed
+   * across every model call in the agent loop
+   * (`metrics.latestAgentInvocation.usage.outputTokens`).
+   *
+   * Per-invocation: counted from this call, not cumulative across reuses of
+   * the same agent. Distinct from per-call provider-level `maxTokens`
+   * settings (e.g. `GoogleModelConfig.params.maxOutputTokens`), which bound
+   * a single model call's output. This option bounds the loop's cumulative
+   * output across however many calls it makes.
+   *
+   * Checked at the top of each loop iteration. When the count is at or above
+   * `maxOutputTokens`, the agent returns with
+   * `stopReason: 'maxOutputTokensExceeded'`.
+   *
+   * Soft cap: a single oversized model response can overshoot the budget.
+   * The agent stops at the first turn boundary on or after the budget is
+   * reached; it does not bound any individual model call.
+   *
+   * Omit for no limit.
+   */
+  maxOutputTokens?: number
+
+  /**
+   * Maximum cumulative input + output tokens for this invocation
+   * (`metrics.latestAgentInvocation.usage.totalTokens`). Each model call's
+   * input includes prior turns, so this counter compounds across the run —
+   * it approximates the total token spend you would be billed for.
+   *
+   * Per-invocation: counted from this call, not cumulative across reuses of
+   * the same agent.
+   *
+   * Checked at the top of each loop iteration. When the count is at or above
+   * `maxTotalTokens`, the agent returns with
+   * `stopReason: 'maxTotalTokensExceeded'`.
+   *
+   * Soft cap: a single oversized model response can overshoot the budget.
+   * The agent stops at the first turn boundary on or after the budget is
+   * reached; it does not bound any individual model call.
+   *
+   * If both `maxOutputTokens` and `maxTotalTokens` would trip on the same
+   * iteration, `maxTotalTokensExceeded` wins. If `maxTurns` would also trip,
+   * `maxTurnsExceeded` wins over both.
+   *
+   * Omit for no limit.
+   */
+  maxTotalTokens?: number
 }
 
 /**
--- a/strands-ts/src/types/agent.ts
+++ b/strands-ts/src/types/agent.ts
@@ -120,68 +120,58 @@ export interface InvokeOptions {
   cancelSignal?: AbortSignal
 
   /**
-   * Maximum number of agent loop iterations (turns) for this invocation.
-   * A turn is one model call plus any tool execution that follows.
+   * Per-invocation budget caps. Each cap, when set, bounds the agent loop
+   * for this `invoke()` / `stream()` call only — counters are not cumulative
+   * across reuses of the same agent.
    *
-   * Per-invocation: counted from this `invoke()` / `stream()` call, not
-   * cumulative across reuses of the same agent. Checked at the top of each
-   * loop iteration against
-   * `metrics.latestAgentInvocation.cycles.length`. Tools requested by the
-   * previous turn always run to completion before the cap fires, so
+   * Caps are checked at the top of each loop iteration. Tools requested by
+   * the previous turn always run to completion before a cap fires, so
    * `agent.messages` remains in a reinvokable state.
    *
-   * Must be a positive finite number. Omit (or set to `undefined`) for no
-   * limit.
+   * Each cap, when set, must be a positive finite number. Omit any field
+   * (or `limits` itself) for no limit on that dimension.
+   *
+   * Priority on simultaneous trip (highest first): `turns`, `totalTokens`,
+   * `outputTokens`. The corresponding `stopReason` is `'limitTurns'`,
+   * `'limitTotalTokens'`, or `'limitOutputTokens'`.
    */
-  maxTurns?: number
+  limits?: {
+    /**
+     * Maximum number of agent loop iterations (turns). A turn is one model
+     * call plus any tool execution that follows. Counted against
+     * `metrics.latestAgentInvocation.cycles.length`.
+     */
+    turns?: number
 
-  /**
-   * Maximum cumulative model-generated tokens for this invocation, summed
-   * across every model call in the agent loop
-   * (`metrics.latestAgentInvocation.usage.outputTokens`).
-   *
-   * Per-invocation: counted from this call, not cumulative across reuses of
-   * the same agent. Distinct from per-call provider-level `maxTokens`
-   * settings (e.g. `GoogleModelConfig.params.maxOutputTokens`), which bound
-   * a single model call's output. This option bounds the loop's cumulative
-   * output across however many calls it makes.
-   *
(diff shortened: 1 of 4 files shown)


strands-ts/src/types/agent.ts
```diff
@@ -118,6 +118,59 @@ export interface InvokeOptions {
...
+   *
+   * Checked at the top of each loop iteration: when the meter's `cycleCount`
+   * is at or above `maxTurns`, the agent returns an `AgentResult` with
+   * `stopReason: 'maxTurnsExceeded'`. Tools requested by the previous turn
+   * always run to completion before the cap fires, so `agent.messages`
+   * remains in a reinvokable state.
+   *
+   * Omit (or set to `undefined`) for no limit.
+   */
+  maxTurns?: number
```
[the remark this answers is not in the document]
notowen333: Intentionally not validating these. Bad inputs (negative / NaN / fractional) either no-op or trip immediately, both of which are obvious from the call site, and adding a per-field validator for three optional numbers is more code than the failure mode warrants. Leaving the bounds enforcement to the caller.

agent-of-mkmeral (commented):
Hey @notowen333 — pulled the branch, ran the suite locally (all 107 tests pass including the 12 new `budget caps` tests, no type errors). The design is clean and the per-invocation scoping via `latestAgentInvocation` is the right call. A few targeted observations beyond the bot reviews:

### 🟡 Real issue — observability gap in the budget-stop path

The budget-stop branch (`agent.ts` ~L965-979) does:

```ts
if (budgetStopReason) {
  const lastMessage = this.messages.at(-1) ?? new Message({ role: 'assistant', content: [] })
  return new AgentResult({ ... })   // ← bare return, never assigns to `result`
}
```

Unlike the `endTurn` (L1038), `cancelled` (L1069), and tool-cancelled (L1127) paths, this one doesn't assign to the local `result`. That matters for the `finally` block at L1201-1206:

```ts
this._tracer.endAgentSpan(agentSpan, {
  ...(caughtError && { error: caughtError }),
  ...(result?.lastMessage && { response: result.lastMessage }),       // ← missing
  accumulatedUsage: this._meter.metrics.accumulatedUsage,
  ...(result?.stopReason && { stopReason: result.stopReason }),       // ← missing
})
```

So budget-stopped invocations end up with an agent span that has **no `response` and no `stopReason`** in telemetry — every other termination path records both. Easy fix: assign to `result` before returning, matching the existing patterns. (The bot called this out as fragility; the concrete impact is missing OTEL/Langfuse attributes on every budget-stopped run, which is the exact case operators will want to see in dashboards.)

### 🟡 `maxTurns: 0` — `agent.messages` ≠ `result.lastMessage`

When `maxTurns: 0` trips on iteration 1, the synthesized empty assistant message is returned as `result.lastMessage` but **not appended** to `agent.messages`. The `cancelled` path at L1166 explicitly appends its synthetic message. So:

```ts
result.lastMessage           // empty assistant message
agent.messages.at(-1)        // last user message
```

These should be consistent or the asymmetry should be commented. I'd lean toward appending (matches `cancelled`) since the test already documents `lastMessage` as an empty assistant message — appending it would also make `agent.messages` reinvokable per the TSDoc claim that "`agent.messages` remains in a reinvokable state." Right now, with `maxTurns: 0`, that claim doesn't hold.

### 🟢 On the input-validation thread

Your decision to skip validation is defensible — the failure modes are obvious from the call site for negative/fractional values. One asymmetry worth noting though: `maxTurns: -1` produces a *silent* `maxTurnsExceeded` with an empty assistant message (since `0 >= -1`), while `maxTurns: NaN` silently no-ops. Different inputs that both "look broken" produce different surprising behaviors. Not a blocker, but if you want to lean into the "obvious path is happy path" tenet, a one-liner check at the top of `_stream` is cheap insurance. Up to you.

### 🟢 Naming & API review

`needs-api-review` is correctly applied. The TSDoc explicitly contrasts `maxOutputTokens` (loop-cumulative) against `GoogleModelConfig.params.maxOutputTokens` (per-call), which I think is sufficient — but worth flagging in the API review meeting since the names are now overloaded.

### 🟢 Test coverage

Very solid. I particularly like the cross-invocation reuse test (catches the lifetime-vs-per-invocation regression) and the `>=` boundary test. Two small nits already raised by the bot review: stronger assertions on `r2.lastMessage` text in the reuse test, and the streaming test could assert `cycleCount: 1` in metrics. Both nice-to-have, neither blocking.

### Summary

- **Fix before merge:** assign `result` in the budget-stop branch (1 line) — observability regression.
- **Worth addressing:** `maxTurns: 0` message-append inconsistency (TSDoc claim doesn't hold today).
- **Author's call:** input validation, naming, test assertion strength.

Nice feature, integrates well with existing patterns. 🚀

agent-of-mkmeral (approved):
Pulled `a047689`, all 110 unit tests pass locally with no type errors. ✅

### Fixes verified

1. **Observability fix** — `result = new AgentResult(...); return result` matches the existing `endTurn` / `cancelled` patterns. The `finally` block now sees `result` and `endAgentSpan` populates `response` + `stopReason` for budget-stopped invocations. ✅

2. **Reject-zero approach** — nicely sidesteps the `maxTurns: 0` swallow-the-prompt edge case. With `maxTurns >= 1` enforced, at least one cycle completes before `_checkMaxBudgets` can trip, so `this.messages.at(-1)!` is safe by construction (no need for the synthesized empty-assistant fallback). The non-null assertion is well-grounded.

3. **`_validateMaxBudgets`** — clean implementation. `Number.isFinite(value) && value > 0` correctly catches `0`, negatives, `NaN`, `+Infinity`, and `-Infinity` in one expression. Allowing fractional values is the right call — useful for token budgets derived from arithmetic (e.g. `total * 0.8`), and harmless since `>=` comparison still works.

4. **Cross-invocation test** — the new `latestAgentInvocation.cycles.length === 1` assertion per invocation is exactly what proves the per-invocation scoping. This would catch any future regression to lifetime accumulators.

5. **`it.each` for invalid inputs** — covers negative / zero / NaN / Infinity cleanly.

### Naming

Agreed on deferring `maxOutputTokens` to API review per the label. The TSDoc contrast against `GoogleModelConfig.params.maxOutputTokens` is sufficient documentation in the meantime.

LGTM 🚀

strands-ts/src/types/agent.ts
```diff
@@ -118,6 +118,70 @@ export interface InvokeOptions {
...
+   * cumulative across reuses of the same agent. Checked at the top of each
+   * loop iteration against
+   * `metrics.latestAgentInvocation.cycles.length`. Tools requested by the
+   * previous turn always run to completion before the cap fires, so
+   * `agent.messages` remains in a reinvokable state.
+   *
+   * Must be a positive finite number. Omit (or set to `undefined`) for no
+   * limit.
+   */
+  maxTurns?: number
```
zastrowm: Can we generalize this to a `limits` object and the stopReason to `limitExceeded`?

notowen333: discussed offline; so users can see the string reason we're going to prefix all the stop reason with `limit`

zastrowm (approved):

agent-of-mkmeral (approved):
Pulled `36fe874`, ran the full agent suite locally — **110/110 pass, no type errors**. ✅

### `limits` refactor verified

The `@zastrowm` API suggestion landed cleanly:

- **Shape**: Top-level `maxTurns` / `maxOutputTokens` / `maxTotalTokens` → single `limits: { turns?, outputTokens?, totalTokens? }` object. `InvokeOptions` surface stays narrow; future caps (wall-clock, cost, etc.) join under `limits` without further top-level growth.
- **Stop reasons**: `maxTurnsExceeded` / `maxOutputTokensExceeded` / `maxTotalTokensExceeded` → `limitTurns` / `limitOutputTokens` / `limitTotalTokens`. Symmetric with the input shape, granular enough for telemetry filtering. Prefix-grouping is a nice touch.
- **TSDoc**: Hoisted the per-invocation / soft-cap / `agent.messages` reinvokability invariants to the parent `limits` JSDoc; removed duplication across the three fields. The `GoogleModelConfig.params.maxOutputTokens` distinction is preserved on the `outputTokens` field.

### Safety properties preserved end-to-end

| Property | Status |
|---|---|
| Observability fix — `result` assigned before return in limits-stop branch | ✅ preserved |
| `_validateLimits` rejects 0/negative/NaN/±Infinity with `TypeError` | ✅ preserved (now at `options?.limits?.{turns,outputTokens,totalTokens}`) |
| `this.messages.at(-1)!` non-null assertion safe | ✅ still grounded — `limits.turns >= 1` enforced → at least one cycle completes before `_checkLimits` can trip |
| Per-invocation scoping via `latestAgentInvocation` | ✅ preserved |
| Priority order on simultaneous trip: turns → totalTokens → outputTokens | ✅ preserved |

### Test updates

Test file mechanically updated to match (`{ maxTurns: 1 }` → `{ limits: { turns: 1 } }`; `'maxTurnsExceeded'` → `'limitTurns'`; etc.). Coverage unchanged — 12 budget-caps tests + 3 invalid-input tests still in `it.each` form. The cross-invocation reuse test still asserts `latestAgentInvocation.cycles.length === 1` per call. All boundary semantics, priority, natural-end-wins, streaming, and reuse cases preserved.

Good call by `@zastrowm`, clean execution by `@notowen333`. LGTM 🚀

